PR #6, opened 2023-09-07
DNS | Fix zone updates failing

hcsa73:
API returns error on update if `primaries` field is set, even if it's set to its previous value
Fixes:
- Set `primaries` field as `RequiresReplace`
- Do not set `primaries` in `toUpdatePayload`


joaopalet (approved):

stackit/services/dns/zone/resource_test.go
```diff
@@ -306,22 +305,18 @@ func TestToPayloadUpdate(t *testing.T) {
...
-				ExpireTime:    types.Int64Value(5),
-				DefaultTTL:    types.Int64Value(4534534),
-				NegativeCache: types.Int64Value(-4534534),
-				IsReverseZone: types.BoolValue(true),
+				RetryTime:         types.Int64Value(3),
+				RefreshTime:       types.Int64Value(4),
+				ExpireTime:        types.Int64Value(5),
+				DefaultTTL:        types.Int64Value(4534534),
+				NegativeCache:     types.Int64Value(-4534534),
+				IsReverseZone:     types.BoolValue(true),
```
joaopalet, line 313: Shoud we instead leave the field here in the input? As that's what is actually happening on the resource right? Even though we then still expect the output to be `nil`

hcsa73: Yes, makes sense. Fixed it

joaopalet (approved):

PR #15, opened 2023-09-11
Add migration import examples for the SKE and LogMe services

joaopalet:
- Also fixes the description of the DNS record set internal ID

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -78,5 +78,44 @@ resource "stackit_ske_cluster" "cluster-example" {
   ]
   project_id = "project_id"
 }
-
 ```
+
+### Example (LogMe service)
+
+Import configuration:
+```terraform
+import {
+  id = "project_id,instance_id"
+  to = stackit_logme_instance.example-instance
+}
+
+import {
+  id = "project_id,instance_id,credentials_id"
+  to = stackit_logme_credentials.example-credentials
+}
+```
+
+Generated configuration:
+```terraform
+# __generated__ by Terraform
+# Please review these resources and move them into your main configuration files.
+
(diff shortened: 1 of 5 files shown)


MIGRATION.md
```diff
@@ -10,3 +10,54 @@ Before you begin the migration process, please ensure that you have done the nec
 For existing resources created with the old provider, you'll need to import them into your new configuration. Terraform provides a feature for importing existing resources and auto-generating new Terraform configuration files. To generate configuration code for the imported resources, refer to the official [Terraform documentation](https://developer.hashicorp.com/terraform/language/import/generating-configuration) for step-by-step guidance.
 
 Once the configuration is generated, compare the generated file with your existing configuration. Be aware that field names may have changed so you should adapt the configuration accordingly. However, not all attributes from the generated configuration are needed for managing the infrastructure, meaning this set of fields can be reduced to the relevant ones from your previous configuration. Check the Terraform plan for the imported resource to identify any differences.
+
+### Example (DNS service)
```
vicentepinto98, line 14: Can you also add an example for the Kubernetes service?

The process is the same but I think it would be more helpful since users would definitely need to migrate their cluster.

(Note: DNS is not part of the community provider)

vicentepinto98 (changes requested):

MIGRATION.md
joaopalet: Ah right, I'll replace it with a SKE and LogMe examples then

joaopalet: And there we could also highlighting the current issue with the import of DSAs (which will be fixed next sprint)

vicentepinto98 (approved):

PR #20, opened 2023-09-19
Extend DSA `Read`s to compute plan name and version from plan ID

joaopalet:


stackit/services/mariadb/instance/resource.go
```diff
@@ -622,3 +627,25 @@ func (r *instanceResource) loadPlanId(ctx context.Context, diags *diag.Diagnosti
 	}
 	diags.AddError("Invalid plan_name", fmt.Sprintf("Couldn't find plan_name '%s' for version %s, available names are:%s", planName, version, availablePlanNames))
 }
+
+func (r *instanceResource) loadPlanNameAndVersion(ctx context.Context, diags *diag.Diagnostics, model *Model) {
```
joaopalet, line 631: Instead of duplicating this for the data sources, we could also pass in the client as a function parameter (as it's the same client for both resource and data source)

joaopalet: Or even fetch the offerings outside this function, and make it a local function that just searches for the respective plan name and version

vicentepinto98: I would go with the first option

stackit/services/mariadb/instance/resource.go
```diff
@@ -292,6 +292,10 @@ func (r *instanceResource) Read(ctx context.Context, req resource.ReadRequest, r
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error mapping fields", err.Error())
 		return
 	}
+
+	// Compute and store values not present in the API response
+	r.loadPlanNameAndVersion(ctx, &diags, &state)
```
vicentepinto98, line 297: We should check for errors after `loadPlanNameAndVersion`.
The question is that if the `Read` should fail when loading the name and version fails. I would say it shouldnt, but maybe a warning would be good. In that case, I would make line 635 below a warning instead of error

stackit/services/mariadb/instance/datasource.go
```diff
@@ -174,8 +177,33 @@ func (r *instanceDataSource) Read(ctx context.Context, req datasource.ReadReques
...
+	// Compute and store values not present in the API response
+	r.loadPlanNameAndVersion(ctx, &diags, &state)
+
 	// Set refreshed state
 	diags = resp.State.Set(ctx, &state)
 	resp.Diagnostics.Append(diags...)
 	tflog.Info(ctx, "MariaDB instance read")
 }
+
+func (r *instanceDataSource) loadPlanNameAndVersion(ctx context.Context, diags *diag.Diagnostics, model *Model) {
```
vicentepinto98, line 190: see comment below, but I think here we should not duplicate the function and use the one defined in `resource.go`, like we do for the `mapFields` functions

stackit/services/mariadb/instance/resource.go
joaopalet: Agree, added a check and early return in case the call to `GetOfferings` fails but kept the warning if for some reason the `plan_name` and `version` can't be found for the returned `plan_id`. This way it doesn't block the normal functioning of the TFP (as it's only relevant for the import) and there the user can then recover by getting the fields from the Portal e.g.

vicentepinto98 (approved):

PR #35, opened 2023-09-27
Add error details to logs

hcsa73:


stackit/internal/core/core.go
```diff
@@ -51,12 +51,12 @@ func DiagsToError(diags diag.Diagnostics) error {
...
 func LogAndAddError(ctx context.Context, diags *diag.Diagnostics, summary, detail string) {
-	tflog.Error(ctx, summary)
+	tflog.Error(ctx, fmt.Sprintf("%s | %s", summary, detail))
 	diags.AddError(summary, detail)
 }
 
 // LogAndAddWarning Logs the warning and adds it to the diags
 func LogAndAddWarning(ctx context.Context, diags *diag.Diagnostics, summary, detail string) {
-	tflog.Warn(ctx, summary)
+	tflog.Warn(ctx, fmt.Sprintf("%s | %s", summary, detail))
```
vicentepinto98, line 60: Maybe ":" instead of "|" ?

vicentepinto98 (approved):

stackit/internal/core/core.go
hcsa73: Minor detail, it's just to better separate summary from details. Also they're usually both capitalized

PR #45, opened 2023-09-28
Object Storage: implement bucket

hcsa73:


stackit/internal/services/objectstorage/bucket/datasource.go
```diff
@@ -0,0 +1,150 @@
...
+	r.client = apiClient
+	tflog.Info(ctx, "ObjectStorage bucket client configured")
+}
+
+// Schema defines the schema for the resource.
+func (r *bucketDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	descriptions := map[string]string{
+		"main":                     "ObjectStorage credentials data source schema.",
+		"id":                       "Terraform's internal resource identifier. It is structured as \"`project_id`,`bucket_name`\".",
+		"bucket_name":              "The bucket name. Has to be DNS conform.",
```
vicentepinto98, line 79:
```suggestion
		"bucket_name":              "The bucket name. It must be DNS conform.",
```

stackit/internal/services/objectstorage/bucket/datasource.go
```diff
@@ -0,0 +1,150 @@
...
+func NewBucketDataSource() datasource.DataSource {
+	return &bucketDataSource{}
+}
+
+// bucketDataSource is the data source implementation.
+type bucketDataSource struct {
+	client *objectstorage.APIClient
+}
+
+// Metadata returns the resource type name.
```
vicentepinto98, line 33:
```suggestion
// Metadata returns the datasource type name.
```

stackit/internal/services/objectstorage/bucket/resource.go
```diff
@@ -0,0 +1,295 @@
...
+	ctx = tflog.SetField(ctx, "bucket_name", bucketName)
+
+	// Create new recordset
+	_, err := r.client.CreateBucket(ctx, projectId, bucketName).Execute()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating bucket", fmt.Sprintf("Calling API: %v", err))
+		return
+	}
+
+	wr, err := objectstorage.CreateBucketWaitHandler(ctx, r.client, projectId, bucketName).SetTimeout(1 * time.Minute).WaitWithContext(ctx)
```
vicentepinto98, line 164: Why just 1 minute? Even if it is really fast, why should we have such a small timeout?

stackit/internal/services/objectstorage/bucket/resource.go
```diff
@@ -0,0 +1,295 @@
...
+	bucketName := model.BucketName.ValueString()
+	ctx = tflog.SetField(ctx, "project_id", projectId)
+	ctx = tflog.SetField(ctx, "bucket_name", bucketName)
+
+	// Delete existing bucket
+	_, err := r.client.DeleteBucket(ctx, projectId, bucketName).Execute()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error deleting bucket", fmt.Sprintf("Calling API: %v", err))
+	}
+	_, err = objectstorage.DeleteBucketWaitHandler(ctx, r.client, projectId, bucketName).SetTimeout(1 * time.Minute).WaitWithContext(ctx)
```
vicentepinto98, line 248: same as above for the deletion timeout

stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -0,0 +1,99 @@
...
+					bucket_name    = "%s"
+				}
+				`,
+		testutil.ObjectStorageProviderConfig(),
+		bucketResource["project_id"],
+		bucketResource["bucket_name"],
+	)
+}
+
+func TestAccObjectStorageResource(t *testing.T) {
```
vicentepinto98, line 34: It is missing the TestCheckDestroy() function to confirm the resource has been deleted. Check the other acc tests

stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -0,0 +1,99 @@
...
+			{
+				ResourceName: "stackit_objectstorage_bucket.bucket",
+				ImportStateIdFunc: func(s *terraform.State) (string, error) {
+					r, ok := s.RootModule().Resources["stackit_objectstorage_bucket.bucket"]
+					if !ok {
+						return "", fmt.Errorf("couldn't find resource stackit_objectstorage_bucket.bucket")
+					}
+					bucketName, ok := r.Primary.Attributes["bucket_name"]
+					if !ok {
+						return "", fmt.Errorf("couldn't find attribute instance_id")
```
vicentepinto98, line 88:
```suggestion
						return "", fmt.Errorf("couldn't find attribute bucket_name")
```

vicentepinto98 (changes requested):

stackit/internal/services/objectstorage/bucket/resource.go
hcsa73: From my tests, creation was always under 5s and deletion was under 15s. Set the timeout as 1 minute because it seems like, if it would take that long, something is definitely wrong.

stackit/internal/services/objectstorage/bucket/resource.go
hcsa73: Check above

stackit/internal/services/objectstorage/bucket/resource.go
vicentepinto98: Ok lets start with this and then check again if there are a lot of timeouts

hcsa73: It's also used by other resources as well

stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+	})
+}
+
+func testAccCheckObjectStorageDestroy(s *terraform.State) error {
+	ctx := context.Background()
+	var client *objectstorage.APIClient
+	var err error
+	if testutil.ObjectStorageCustomEndpoint == "" {
+		client, err = objectstorage.NewAPIClient()
```
vicentepinto98, line 113: region config is missing here. If you copied this from other acc tests, it might be missing there too

stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -0,0 +1,156 @@
...
+		if bucket.Name == nil {
+			continue
+		}
+		bucketName := *bucket.Name
+		if utils.Contains(bucketsToDestroy, *bucket.Name) {
+			_, err := client.DeleteBucketExecute(ctx, testutil.ProjectId, bucketName)
+			if err != nil {
+				return fmt.Errorf("destroying bucket %s during CheckDestroy: %w", bucketName, err)
+			}
+			_, err = objectstorage.DeleteBucketWaitHandler(ctx, client, testutil.ProjectId, bucketName).WaitWithContext(ctx)
```
vicentepinto98, line 149: you can add the 1 minute timeout here also

vicentepinto98 (changes requested):

stackit/internal/services/objectstorage/objectstorage_acc_test.go
hcsa73: Ideally this should access the resource's API client, it shouldn't have to initialize a new one. But I couldn't find a way to do it. Settled for the quick fix of setting the region as "eu01", which is aligned with what `testutil.ObjectStorageProviderConfig()` (the provider config used for the acceptance tests) does

vicentepinto98 (approved):

PR #82, opened 2023-10-13
Object storage misc fixes

hcsa73:
- Credentials group logic wasn't fetching "credentials_group_id"
- Credential's "credentials_group_id" was set as "computed" in the data source schema
- Credential's "expiration_timestamp" was set as "required" in the data source schema
- Credential's "expiration_timestamp" is now RFC3339 without fractional seconds (API sometimes returns .000 in fractional seconds even if they're not 0) 
- Wrong checks in acc tests, also skipped import check on credential keys (can't be fetched fetched on import)
- Set retry limit for credential's unit test that uses mocked HTTP server (to be faster)

### DIFF
--- a/stackit/internal/services/objectstorage/credential/resource.go
+++ b/stackit/internal/services/objectstorage/credential/resource.go
@@ -98,7 +98,7 @@ func (r *credentialResource) Schema(_ context.Context, _ resource.SchemaRequest,
 		"credential_id":        "The credential ID.",
 		"credentials_group_id": "The credential group ID.",
 		"project_id":           "STACKIT Project ID to which the credential group is associated.",
-		"expiration_timestamp": "Expiration timestamp, in RFC339 format (seconds only). If null, the credential never expires.",
+		"expiration_timestamp": "Expiration timestamp, in RFC339 format without fractional seconds. Example: \"2025-01-01T00:00:00Z\". If not set, the credential never expires.",
 	}
 
 	resp.Schema = schema.Schema{
--- a/stackit/internal/services/objectstorage/credential/resource.go
+++ b/stackit/internal/services/objectstorage/credential/resource.go
@@ -124,7 +124,7 @@ func (r *credentialResource) Schema(_ context.Context, _ resource.SchemaRequest,
 			},
 			"credentials_group_id": schema.StringAttribute{
 				Description: descriptions["credentials_group_id"],
-				Computed:    true,
+				Required:    true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
 				},
--- a/stackit/internal/services/objectstorage/credential/resource.go
+++ b/stackit/internal/services/objectstorage/credential/resource.go
@@ -358,6 +358,18 @@ func mapFields(credentialResp *objectstorage.CreateAccessKeyResponse, model *Mod
 		return fmt.Errorf("credential id not present")
 	}
 
+	if credentialResp.Expires == nil {
+		model.ExpirationTimestamp = types.StringNull()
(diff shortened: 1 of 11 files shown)


stackit/internal/services/objectstorage/credential/resource_test.go
```diff
@@ -32,7 +31,8 @@ func (c *objectStorageClientMocked) CreateProjectExecute(_ context.Context, proj
 }
 
 func TestMapFields(t *testing.T) {
```
joaopalet, line 33: Does it make sense to add a test case for the "Harmonize the timestamp format"?

stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -168,14 +168,6 @@ func TestAccObjectStorageResource(t *testing.T) {
...
 						"data.stackit_objectstorage_credential.credential", "credential_id",
 					),
-					resource.TestCheckResourceAttrPair(
-						"stackit_objectstorage_credential.credential", "access_key",
-						"data.stackit_objectstorage_credential.credential", "access_key",
-					),
-					resource.TestCheckResourceAttrPair(
-						"stackit_objectstorage_credential.credential", "secret_access_key",
-						"data.stackit_objectstorage_credential.credential", "secret_access_key",
-					),
```
joaopalet, line 178: What was happening with these two?

stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -221,8 +213,9 @@ func TestAccObjectStorageResource(t *testing.T) {
 					}
 					return fmt.Sprintf("%s,%s,%s", testutil.ProjectId, credentialsGroupId, credentialId), nil
 				},
-				ImportState:       true,
-				ImportStateVerify: true,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"access_key", "secret_access_key"},
```
joaopalet, line 218: Here only `secret_access_key` is sensitive, right? Why can't we import `access_key`?

stackit/internal/services/objectstorage/credential/resource.go
```diff
@@ -99,6 +98,7 @@ func (r *credentialResource) Schema(_ context.Context, _ resource.SchemaRequest,
 		"credential_id":        "The credential ID.",
 		"credentials_group_id": "The credential group ID.",
 		"project_id":           "STACKIT Project ID to which the credential group is associated.",
+		"expiration_timestamp": "Expiration timestamp, in RFC339 format (seconds only). If null, the credential never expires.",
```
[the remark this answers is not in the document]
hcsa73: I mean without fractional seconds. Rewrote the field description

stackit/internal/services/objectstorage/objectstorage_acc_test.go
hcsa73: Was failing because the data source doesn't have these keys, because they're only given on creation

stackit/internal/services/objectstorage/objectstorage_acc_test.go
hcsa73: As far as I know, only `secret_access_key` is sensitive. But `access_key` isn't provided as well in the GET response, so can't be fetched for imported resources

stackit/internal/services/objectstorage/credential/resource_test.go
hcsa73: Yes it does

joaopalet (approved):

PR #84, opened 2023-10-13
Object storage improvements

hcsa73:
- Add examples
- Rename bucket's `bucket_name` --> `bucket`


hcsa73:
The example was already merged

joaopalet:
Ah I think the issue is that the example folder names need to reflect the name of the resource/data source.

Meaning it should be `examples/…/stackit_objectstorage_…/` instead of `examples/…/stackit_object_storage_…/`

hcsa73:
Ah, it's the folder names that are wrong

joaopalet (approved):

PR #88, opened 2023-10-16
Onboard MongoDB Flex user

joaopalet:
- Also adds examples and generates documentation


docs/data-sources/secretsmanager_instance.md
hcsa73, line 1: Don't commit this file in this PR, will be added in a secrets manager PR

docs/resources/secretsmanager_instance.md
hcsa73, line 1: Don't commit this file in this PR, will be added in a secrets manager PR

stackit/internal/services/mongodbflex/user/resource.go
```diff
@@ -0,0 +1,440 @@
...
+				},
+			},
+			"username": schema.StringAttribute{
+				Optional: true,
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"roles": schema.SetAttribute{
```
hcsa73, line 159: Maybe add available roles in the aqttribute's description, and what they do. Or are they subject to change in the API's side?

stackit/internal/services/mongodbflex/user/resource.go
```diff
@@ -0,0 +1,440 @@
...
+				ElementType: types.StringType,
+				Required:    true,
+				Validators: []validator.Set{
+					setvalidator.ValueStringsAre(
+						stringvalidator.OneOf("read", "readWrite"),
+					),
+				},
+			},
+			"database": schema.StringAttribute{
+				Required: true,
```
hcsa73, line 169: If no updates are possible, then this must have a `RequiresReplace` planning modifier

stackit/internal/services/mongodbflex/user/datasource.go
```diff
@@ -0,0 +1,232 @@
...
+	// Set refreshed state
+	diags = resp.State.Set(ctx, model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	tflog.Info(ctx, "PostgresFlex user read")
+}
+
+func mapDataSourceFields(userResp *mongodbflex.GetUserResponse, model *DataSourceModel) error {
```
hcsa73, line 187: Isn't this the same as `mapFields`? If not, then add a test for this function

stackit/internal/services/mongodbflex/user/resource_test.go
```diff
@@ -0,0 +1,374 @@
...
+				diff := cmp.Diff(state, &tt.expected)
+				if diff != "" {
+					t.Fatalf("Data does not match: %s", diff)
+				}
+			}
+		})
+	}
+}
+
+func TestMapCreate(t *testing.T) {
```
hcsa73, line 158: Should be `TestMapFields`

stackit/internal/services/mongodbflex/user/resource.go
joaopalet: I would stick to the field description, also since the available roles are already in the field validation

stackit/internal/services/mongodbflex/user/resource.go
joaopalet: Forgot to add it here, good catch

stackit/internal/services/mongodbflex/user/datasource.go
joaopalet: The model argument is different, here it is `DataSourceModel` (which does not contain the `password` field). Added test for it

stackit/internal/services/mongodbflex/user/resource.go
hcsa73: Does it say what values are valid if you use wrong values?

hcsa73 (approved):

stackit/internal/services/mongodbflex/user/resource.go
joaopalet: Yes

Issue #87, opened 2023-10-16
Missing required attribute region for stackit_postgresflex_instance


docbyte86:
Error:

│ Error: Error configuring API client
│ 
│   with module.db.stackit_postgresflex_instance.database,
│   on .terraform/modules/db/main.tf line 5, in resource "stackit_postgresflex_instance" "database":
│    5: resource "stackit_postgresflex_instance" "database" {
│ 
│ Configuring client: configuring region: no region was provided, available
│ regions are: [eu01]

The required attribute region is missing in the documentation of version 0.5.0
[stackit_postgresflex_instance (Resource)](https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/resources/postgresflex_instance)

If I set region as attribute like:

resource "stackit_postgresflex_instance" "database" {
  acl             = var.database_acl
  backup_schedule = var.database_backup_schedule
  flavor          = var.database_flavor
  name            = var.database_name
  project_id      = var.database_project_id
  replicas        = var.database_replicas
  storage         = var.database_storage
  version         = var.database_version
  region          = var.database_region
}

variable "database_region" {
  description = "The region where the instance runs in."
  type        = string
  default     = "eu01"
}

I'll get the following error:

│ Error: Unsupported argument
│ 
│   on .terraform/modules/db/main.tf line 14, in resource "stackit_postgresflex_instance" "database":
│   14:   region          = var.database_region
│ 
│ An argument named "region" is not expected here.

hcsa73:
Hello, thanks for your feedback!

The `region` argument is set in the provider configuration, as described [here](https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs):
```terraform
provider "stackit" {
    region = "eu01"
}
```

This argument is required to access some of the resources. However, neither the error message you found nor the PostgresFlex documentation page describe this clearly. We'll make changes to address this.

PR #90, opened 2023-10-17
Improve region details

hcsa73:
Follow-up from https://github.com/stackitcloud/terraform-provider-stackit/issues/87
- Add detail to error message that client configuration failing is due to provider misconfiguration
- Add detail in the schema description of resources/data sources that are region-specific
- Remove resourcemanager using `region`


stackit/internal/services/argus/credential/resource.go
```diff
@@ -86,6 +86,7 @@ func (r *credentialResource) Configure(ctx context.Context, req resource.Configu
 
 func (r *credentialResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
 	resp.Schema = schema.Schema{
+		Description: "Argus credential resource schema. Must have a `region` specified in the provider configuration.",
```
joaopalet, line 89: Maybe this could be a bit misleading as setting the custom endpoint could be enough

joaopalet: Also do you think this should be maintained here? Is it not subject to change for each service?

stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -104,7 +104,7 @@ func (r *projectResource) Configure(ctx context.Context, req resource.ConfigureR
 // Schema defines the schema for the resource.
 func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
 	descriptions := map[string]string{
-		"main":                "Resource Manager project resource schema.",
+		"main":                "Resource Manager project resource schema. Must have a `region` specified in the provider configuration.",
```
joaopalet, line 107: Isn't this one global?

stackit/internal/services/argus/credential/resource.go
hcsa73: Adding a custom endpoint does override the need for the region. I'd leave it as is, because using a custom endpoint won't be done in normal use.

It is subject to change, but that requires changing the API, then the SDK, then the `Configure` routine of the resource

stackit/internal/services/resourcemanager/project/resource.go
hcsa73: Yes it is, the `Configure` routine was reading `region` when it wasn't used. Removed

stackit/internal/services/mongodbflex/user/datasource.go
```diff
@@ -91,7 +91,7 @@ func (r *userDataSource) Configure(ctx context.Context, req datasource.Configure
 func (r *userDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
 	descriptions := map[string]string{
 		"main":        "PostgresFlex user data source schema.",
-		"id":          "Terraform's internal data source. ID. It is structured as \"`project_id`,`instance_id`,`user_id`\".",
```
joaopalet, line 94: You pasted in the wrong row, should be in the `main`

stackit/internal/services/mongodbflex/user/datasource.go
```diff
@@ -91,7 +91,7 @@ func (r *userDataSource) Configure(ctx context.Context, req datasource.Configure
...
+		"id":          "Terraform's internal data source. ID. It is structured as \"`project_id`,`instance_id`,`user_id`\". Must have a `region` specified in the provider configuration.",
```
joaopalet, line 94: You pasted in the wrong row, should be in the `main`

joaopalet (approved):

PR #93, opened 2023-10-17
Implement Secrets Manager ACL

hcsa73:


stackit/internal/services/secretsmanager/instance/resource_test.go
```diff
@@ -134,3 +176,296 @@ func TestToCreatePayload(t *testing.T) {
...
+			expectedACLsStates: map[string]bool{
+				"acl-1": true,
+				"acl-2": true,
+				"acl-3": true,
+			},
+			isValid: true,
+		},
+		{
+			description: "create_acl",
+			acls:        []string{"acl-1", "acl-2", "acl-3", "acl-4"},
```
joaopalet, line 227: Does Terraform already handle the case if the user adds two repeated elements to the set?

stackit/internal/services/secretsmanager/instance/datasource.go
```diff
@@ -108,6 +109,11 @@ func (r *instanceDataSource) Schema(_ context.Context, _ datasource.SchemaReques
 				Description: descriptions["name"],
 				Computed:    true,
 			},
+			"acls": schema.SetAttribute{
+				Description: descriptions["acls"],
```
joaopalet, line 113: The description is missing here in the data source file

stackit/internal/services/secretsmanager/instance/resource.go
```diff
@@ -304,3 +420,54 @@ func toCreatePayload(model *Model) (*secretsmanager.CreateInstancePayload, error
 		Name: model.Name.ValueStringPointer(),
 	}, nil
 }
+
+// updateACLs creates and deletes ACLs so that the instance's ACLs are the ones in the model
+func updateACLs(ctx context.Context, projectId, instanceId string, acls []string, client *secretsmanager.APIClient) error {
```
joaopalet, line 425: Clean solution 👍

stackit/internal/services/secretsmanager/secretsmanager_acc_test.go
```diff
@@ -44,12 +65,19 @@ func TestAccSecretsManager(t *testing.T) {
...
+					instanceResource["acl-1"],
+					instanceResource["acl-2"],
+				))),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					// Instance data
 					resource.TestCheckResourceAttr("stackit_secretsmanager_instance.instance", "project_id", instanceResource["project_id"]),
 					resource.TestCheckResourceAttrSet("stackit_secretsmanager_instance.instance", "instance_id"),
 					resource.TestCheckResourceAttr("stackit_secretsmanager_instance.instance", "name", instanceResource["name"]),
+					resource.TestCheckResourceAttr("stackit_secretsmanager_instance.instance", "acls.#", "2"),
+					resource.TestCheckResourceAttr("stackit_secretsmanager_instance.instance", "acls.0", instanceResource["acl-1"]),
```
joaopalet, line 79: Maybe rename `acl-1` to `acl-0` and `acl-2` to `acl-1` so they match here

stackit/internal/services/secretsmanager/secretsmanager_acc_test.go
```diff
@@ -89,6 +123,34 @@ func TestAccSecretsManager(t *testing.T) {
...
+			{
+				Config: resourceConfig(nil),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// Instance data
+					resource.TestCheckResourceAttr("stackit_secretsmanager_instance.instance", "project_id", instanceResource["project_id"]),
+					resource.TestCheckResourceAttrSet("stackit_secretsmanager_instance.instance", "instance_id"),
+					resource.TestCheckResourceAttr("stackit_secretsmanager_instance.instance", "name", instanceResource["name"]),
+					resource.TestCheckResourceAttr("stackit_secretsmanager_instance.instance", "acls.#", "0"),
+				),
+			},
```
joaopalet, line 153: The test is handled sequentially, right? Meaning the first update happens first and this second update happens after?

stackit/internal/services/secretsmanager/instance/resource_test.go
hcsa73: Good catch! Turns out no, `types.Set` doesn't preserve order but preserves repetitions apparently. Running `ACLs.Elements()` after you put in two repeated elements returns an array with two repeated elements.

Nothing bad should happen, because `updateACLs` should just skip over repeated elements and the API doesn't really care if there are two different ACLs (ie, with different IDs) that are the same.

Nonetheless, I've added test cases for this and guard clauses, and added a validator for checking uniqueness (required changing from `types.Set` to `types.List`)

stackit/internal/services/secretsmanager/secretsmanager_acc_test.go
hcsa73: Yes

joaopalet (approved):

PR #94, opened 2023-10-19
Implement Secrets Manager User, change ACL to Set

hcsa73:
- Implement User (Resource and data source)
- Instance.ACL is now a Set instead of a List. 
  - Was making acc tests fail due to order being important
  - Added acc test case for checking that duplicates are removed
- Reorder resource/data source list


stackit/internal/services/secretsmanager/user/resource.go
```diff
@@ -0,0 +1,410 @@
...
+			"user_id": schema.StringAttribute{
+				Description: descriptions["user_id"],
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
+				},
```
joaopalet, line 125: Do we want/need validators for a computed only field?

joaopalet (approved):

stackit/internal/services/secretsmanager/user/resource.go
hcsa73: Yes, this is telling TF that this value won't change on updated, so TF can show the current value on `terraform plan` instead of "(known after apply)"

PR #97, opened 2023-10-20
Migration guide - Add available resources table

hcsa73:


MIGRATION.md
joaopalet, line 1: - Is it possible to replace the circles with something like ❌ or ✅? The circles kind of remind me of something that is failing
- I would restrict the emojis to the `Import available` column. For the `Official provider` one maybe we can use `-`
- Also I would distinguish the ❌  from something like `-`. E.g. if the resource is not available on the official provider, we would have `-` in the `Import available` column, if it exists but the import is not available we would have ❌
- For the last column maybe we can a title "Notes" or something

hcsa73: Left the `Official provider` column empty whenever there isn't a corresponding resource

joaopalet (approved):

PR #107, opened 2023-10-26
Onboard Load Balancer (part 1: implement creation payload helpers)

joaopalet:
- Adds resource schema and model
- Adds `toCreatePayload` and tests

### DIFF
--- a/stackit/internal/services/loadbalancer/loadbalancer/resource.go
+++ b/stackit/internal/services/loadbalancer/loadbalancer/resource.go
@@ -512,7 +512,7 @@ func toOptionsPayload(ctx context.Context, model *Model) (*loadbalancer.LoadBala
 	if !(model.Options.IsNull() || model.Options.IsUnknown()) {
 		diags := model.Options.As(ctx, optionsModel, basetypes.ObjectAsOptions{})
 		if diags.HasError() {
-			return nil, fmt.Errorf("%v", core.DiagsToError(diags))
+			return nil, fmt.Errorf("%w", core.DiagsToError(diags))
 		}
 	}
 
@@ -521,7 +521,7 @@ func toOptionsPayload(ctx context.Context, model *Model) (*loadbalancer.LoadBala
 		var acl []string
 		diags := optionsModel.ACL.ElementsAs(ctx, &acl, false)
 		if diags.HasError() {
-			return nil, fmt.Errorf("converting acl: %v", core.DiagsToError(diags))
+			return nil, fmt.Errorf("converting acl: %w", core.DiagsToError(diags))
 		}
 		accessControl.AllowedSourceRanges = &acl
 	}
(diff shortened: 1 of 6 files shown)


stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+
+	r.client = apiClient
+	tflog.Info(ctx, "Load Balancer client configured")
+}
+
+// Schema defines the schema for the resource.
+func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	descriptions := map[string]string{
+		"main":                 "Load Balancer project resource schema.",
```
hcsa73, line 157: Is this a project?

stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -0,0 +1,551 @@
...
+				Computed:    true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"network_id": schema.StringAttribute{
+							Description: descriptions["network_id"],
+							Optional:    true,
+							Computed:    true,
+							Validators: []validator.String{
+								validate.UUID(),
+								validate.NoSeparator(),
```
hcsa73, line 267: You don't need the `NoSeparator`. That's only used for fields that are used to build the internal ID (in this case, `project_id` and `name`). It's so it's guaranteed that, when we split `id` by the separator during the `ImportState` routine, we aren't doing anything we shouldn't

stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -0,0 +1,551 @@
...
+							Computed:    true,
+						},
+					},
+				},
+			},
+			"name": schema.StringAttribute{
+				Description: descriptions["name"],
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.String{
```
hcsa73, line 250: Add `NoSeparator` validator (for reasons described in the `network_id` comment)

stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -0,0 +1,551 @@
...
+
+	return &networks
+}
+
+func toOptionsPayload(ctx context.Context, model *Model) (*loadbalancer.LoadBalancerOptions, error) {
+	var optionsModel = &Options{}
+	if !(model.Options.IsNull() || model.Options.IsUnknown()) {
+		diags := model.Options.As(ctx, optionsModel, basetypes.ObjectAsOptions{})
+		if diags.HasError() {
+			return nil, fmt.Errorf("%v", diags.Errors())
```
hcsa73, line 488: Use `core.DiagsToError` instead

stackit/internal/services/loadbalancer/loadbalancer/resource.go
joaopalet: Copy paste leftover

hcsa73 (approved):

PR #116, opened 2023-10-30
Onboard Load Balancer (part 3: add remaining CRUD login and datasource)

joaopalet:


stackit/provider.go
vicentepinto98, line 1: You also need to add the Loadbalancer datasource and resource to the provider in the `Resources` and `Datasources` methods

stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -209,41 +216,65 @@ func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+							},
 						},
 						"protocol": schema.StringAttribute{
 							Description: descriptions["protocol"],
 							Optional:    true,
 							Computed:    true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.RequiresReplace(),
+							},
+							Validators: []validator.String{
```
vicentepinto98, line 257: These oneOf string validators introduce maintenance effort for us, when the API supports new values. Are these defined as enum in the SDK or is this just documented in the OAS? 

Im not sure how we dealt with this in the other resources, but I think we usually dont have these validators

stackit/internal/services/loadbalancer/loadbalancer/datasource.go
```diff
@@ -0,0 +1,318 @@
...
+							Computed:    true,
+						},
+						"port": schema.Int64Attribute{
+							Description: descriptions["port"],
+							Computed:    true,
+						},
+						"protocol": schema.StringAttribute{
+							Description: descriptions["protocol"],
+							Computed:    true,
+							Validators: []validator.String{
```
vicentepinto98, line 147: In the datasource, I dont see why we should have these oneOf validators. If it is returned by the API, it should be a valid value. There is no need to have to maintain this list here

vicentepinto98 (changes requested):

stackit/provider.go
joaopalet: This was intentional, I am onboarding the service bit by bit and I will only add the resource and data source to the provider in the end once the acceptance test is complete and passing

joaopalet: So we don't have a possibly faulty resource /data source in the provider if we want to make a release in the meantime

stackit/internal/services/loadbalancer/loadbalancer/resource.go
joaopalet: True, but I did here since they are defined as an enum and the API error message doesn't indicate the possible values

joaopalet: > Im not sure how we dealt with this in the other resources, but I think we usually dont have these validators

Currently we have `oneOf`s for the DNS and MongoDB Flex also

stackit/internal/services/loadbalancer/loadbalancer/datasource.go
joaopalet: Agree, copy paste leftover

stackit/internal/services/loadbalancer/loadbalancer/resource.go
vicentepinto98: I remember this from the onboarding of the LoadBalancer to the SDK. The way their OAS is defined, the generator doesnt catch that this is an enum and generates these fields as regular string data type. Ideally, the OAS should be changed so that in the SDK these fields have an enum data type, where the validation would be done, instead of having this in the provider schema. 

For now we can keep it, for better user-experience, but we should ask the LoadBalancer team to adapt their OAS

vicentepinto98 (approved):

stackit/internal/services/loadbalancer/loadbalancer/resource.go
joaopalet: Created a [story](https://dev.azure.com/schwarzit/schwarzit.cloudportal/_backlogs/backlog/Developer%20Tools/Epics/?workitem=543989) to investigate this issue and remove these from the TFP once it is fixed

PR #117, opened 2023-10-31
Onboard Load Balancer (part 4: acc tests, examples and documentation)

joaopalet:


examples/resources/stackit_loadbalancer/resource.tf
```diff
@@ -0,0 +1,41 @@
...
+  project_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+  name       = "example-load-balancer"
+  target_pools = [
+    {
+      name        = "example-target-pool"
+      target_port = 80
+      targets = [
+        {
+          display_name = "example-target"
+          ip           = openstack_compute_instance_v2.example.network.0.fixed_ip_v4
```
vicentepinto98, line 11: maybe you should also add here the configuration of the openstack resources. I saw they are defined in the resource description but they should also be here for clarity

stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -195,6 +195,101 @@ func (r *loadBalancerResource) Schema(_ context.Context, _ resource.SchemaReques
 
 	resp.Schema = schema.Schema{
 		Description: descriptions["main"],
+		MarkdownDescription: `
+## Setting up supporting infrastructure` + "\n" + `
+
+### Configuring an OpenStack provider` + "\n" + `
+
+To automate the creation of load balancers, OpenStack can be used to setup the supporting infrastructure.
+To set up the OpenStack provider, you can create a token through the STACKIT Portal, in the your project's Infrastructure API page.
```
vicentepinto98, line 204:
```suggestion
To set up the OpenStack provider, you can create a token through the STACKIT Portal, in your project's Infrastructure API page.
```

stackit/internal/services/loadbalancer/loadbalancer_acc_test.go
```diff
@@ -0,0 +1,322 @@
...
+				role       = "%s"
+				}
+			]
+			options = {
+				private_network_only = %s
+			}
+			}
+		`,
+		supportingInfraResources(loadBalancerResource["name"], OpenStack{
+			userDomainName: "portal_mvp",
```
vicentepinto98, line 85: remove hardcoded values, read from utils

vicentepinto98 (changes requested):

examples/resources/stackit_loadbalancer/resource.tf
joaopalet: But then I would remove them from the resource description, right? Otherwise they would show twice in the documentation

stackit/internal/services/loadbalancer/loadbalancer_acc_test.go
joaopalet: Oops, I have deleted those credentials and will create new ones also

examples/resources/stackit_loadbalancer/resource.tf
vicentepinto98: Yeah I think its better like that. And in the documentation you can reference the example. That way the example can be used as a standalone file by the users to try it out

vicentepinto98 (approved):

PR #121, opened 2023-11-02
Replace terraform ValuePointer function with custom pointer conversion function

vicentepinto98:
- When converting a terraform model value to Go pointer, if the value is unknown set it to nil


hcsa73 (approved):

vicentepinto98:
Fixes #124

Issue #126, opened 2023-11-03
Value Conversion Error stackit_ske_cluster extensions object


Passe-cc:
Hello

I encountered the following error when using the provider:

>  Error: Value Conversion Error
│ 
│   with module.ske-cluster.stackit_ske_cluster.ske_cluster,
│   on ../main.tf line 24, in resource "stackit_ske_cluster" "ske_cluster":
│   24:   extensions                  = var.extensions
│ 
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.
│ 
│ Path: extensions
│ Target Type: *ske.Extensions
│ Suggested Type: basetypes.ObjectValue

### **Steps to Reproduce**
1. create a terraform module for stackit_ske_cluster

**Resource:**
``` yaml
resource "stackit_ske_cluster" "ske_cluster" {
  project_id         = var.project_id
  name               = var.name
  node_pools         = var.node_pools
  extensions         = var.extensions
  ...
```
**variable:**
``` yaml
variable "extensions" {
  type = object({
    acl = optional(object({
      allowed_cidrs = list(string)
      enabled       = bool
    }))
    argus = optional(object({
      enabled           = bool
      argus_instance_id = string
    }))
  })
  default = null
}
```
2. run terraform init & terraform plan

### **Additional Information**
- Provider version:  0.6.2
- Terraform version: 1.5.7

vicentepinto98:
Thank you for opening this issue.
This has been fixed in the v0.6.4 release.

Issue #122, opened 2023-11-02
Error while deploying SKE cluster with terraform due to unknown values


docbyte86:
Provider version 0.6.2

Error:

> │ Error: Value Conversion Error
> │ 
> │   with module.ske.stackit_ske_cluster.ske,
> │   on .terraform/modules/ske/main.tf line 12, in resource "stackit_ske_cluster" "ske":
> │   12:   extensions                  = var.ske_extensions
> │ 
> │ An unexpected error was encountered trying to build a value. This is always
> │ an error in the provider. Please report the following to the provider
> │ developer:
> │ 
> │ Received unknown value, however the target type cannot handle unknown
> │ values. Use the corresponding `types` package type or a custom type that
> │ handles unknown values.
> │ 
> │ Path: extensions
> │ Target Type: *ske.Extensions
> │ Suggested Type: basetypes.ObjectValue

**Child module:**

main.tf:

```
resource "stackit_ske_cluster" "ske" {
  project_id                  = var.ske_project_id
  name                        = var.ske_name
  kubernetes_version          = var.ske_version
  node_pools                  = var.ske_node_pool
  maintenance                 = var.ske_maintenance
  hibernations                = var.ske_hibernations
  extensions                  = var.ske_extensions
  allow_privileged_containers = var.ske_allow_privileged_containers
}
```

variables.tf

```
variable "ske_extensions" {
  description = "Specifies ACL and Argus Instance"
  type = object({
    acl = object({
      allowed_cidrs = list(string)
      enabled       = bool
    })
    argus = object({
      argus_instance_id = optional(string)
      enabled           = bool
    })
  })
}
```

**Root module:**

```
module "ske" {
  source         = "git::https://*****"
  ske_project_id = var.stackit_project_id
  ske_name       = local.ske_instance_name
  ske_version    = var.ske_version
  ske_node_pool  = var.ske_node_pool
  ske_extensions = {
    acl = {
      allowed_cidrs = ["0.0.0.0/0"]
      enabled       = true
    }
    argus = {
      argus_instance_id = "${module.argus.instance_id}"
      enabled           = true
    }
  }
}

```

I am not sure how to handle that error. IMHO I am not providing unknown values? I get this error during tf plan.

docbyte86:
I think this could be related to
`      argus_instance_id = "${module.argus.instance_id}"`
In the stage of tf plan the id is unknown, because it will be created in the same main.tf

```
module "argus" {
  source         = "****"
[...]

PR #127, opened 2023-11-03
SKE Cluster - Change Extensions field model to use types.Object

vicentepinto98:


vicentepinto98:
Fixes #126

vicentepinto98:
Fixes #122

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -771,28 +789,54 @@ func toHibernationsPayload(ctx context.Context, m *Cluster) (*ske.Hibernation, e
...
-	ex := &ske.Extension{}
-	if m.Extensions.Argus != nil {
-		ex.Argus = &ske.Argus{
-			Enabled:         conversion.BoolValueToPointer(m.Extensions.Argus.Enabled),
-			ArgusInstanceId: conversion.StringValueToPointer(m.Extensions.Argus.ArgusInstanceId),
-		}
+	ex := extensions{}
+	diags := m.Extensions.As(ctx, &ex, basetypes.ObjectAsOptions{})
+	if diags.HasError() {
+		return nil, fmt.Errorf("error in extensions object conversion %v", diags.Errors())
```
joaopalet, line 798:
```suggestion
		return nil, fmt.Errorf("converting extensions object: %v", diags.Errors())
```

The outtermost error message should already indicate that there is an error (see https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/WORKSPACE_DOCUMENT/-Mvdp0Q7_oO2mmJTLpNI/)

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -771,28 +789,54 @@ func toHibernationsPayload(ctx context.Context, m *Cluster) (*ske.Hibernation, e
...
 	}
-	if m.Extensions.ACL != nil {
+
+	var skeAcl *ske.ACL
+	if !(ex.ACL.IsNull() || ex.ACL.IsUnknown()) {
+		acl := acl{}
+		diags = ex.ACL.As(ctx, &acl, basetypes.ObjectAsOptions{})
+		if diags.HasError() {
+			return nil, fmt.Errorf("error in extensions.acl object conversion %v", diags.Errors())
```
joaopalet, line 806: Same here

joaopalet: And in the other occurences below

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1106,32 +1153,72 @@ func getMaintenanceTimes(ctx context.Context, cl *ske.ClusterResponse, m *Cluste
...
+				cidrs = append(cidrs, types.StringValue(v))
 			}
 		}
-		m.Extensions.ACL = &ACL{
-			Enabled:      types.BoolPointerValue(cl.Extensions.Acl.Enabled),
-			AllowedCIDRs: types.ListValueMust(types.StringType, cidr),
+
+		aclValues := map[string]attr.Value{
+			"enabled":       enabled,
+			"allowed_cidrs": types.ListValueMust(types.StringType, cidrs),
```
joaopalet, line 1179: I think we should not use the `...ValueMust` functions outside of tests, as they omit possible errors and panic instead. They also mention it in the description: `"This creation function is only recommended to create List values which will not potentially affect practitioners, such as testing, or exhaustively tested provider logic."`

In this case we can map it to a variable first and return an error in case it fails

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1106,32 +1153,72 @@ func getMaintenanceTimes(ctx context.Context, cl *ske.ClusterResponse, m *Cluste
...
+		}
```
joaopalet, line 1180: I think we can use `ListValueFrom` here to do this in one step (not exactly sure, but check for example `SetValueFrom` in the load balancer resource file)

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -803,14 +803,14 @@ func toExtensionsPayload(ctx context.Context, m *Cluster) (*ske.Extension, error
 		acl := acl{}
 		diags = ex.ACL.As(ctx, &acl, basetypes.ObjectAsOptions{})
 		if diags.HasError() {
-			return nil, fmt.Errorf("error in extensions.acl object conversion %v", diags.Errors())
+			return nil, fmt.Errorf("converting extensions.acl object %v", diags.Errors())
```
joaopalet, line 806: Should we also use the `:` to separate the messages?

joaopalet (dismissed):

joaopalet (approved):

PR #149, opened 2023-11-14
Add UseStateForUnknown to prevent unnecessary replacement

vicentepinto98:
- If a field is Computed and RequiresReplacement, an update to any field will cause a force replacement. To fix this, we add UseStateForUnknown


stackit/internal/services/dns/zone/resource.go
joaopalet, line 1: Why is it causing a replacement when updating any field? Is it because these fields change to an `unknown` value?

vicentepinto98: Yes, because the Computed field will be `known after apply`. And since Terraform doesn't know if it will change or not, it forces replacement

PR #161, opened 2023-12-12
Support project UUID identifier in resource manager project

vicentepinto98:
- Fix value conversion error when providing an UUID identifier as `parent_container_Id` on `resourcemanager_project`
- Add `project_id` field to the `resourcemanager_project` resource and data source
 - Improve documentation


stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -102,10 +104,11 @@ func (r *projectResource) Configure(ctx context.Context, req resource.ConfigureR
 // Schema defines the schema for the resource.
 func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
 	descriptions := map[string]string{
-		"main":                "Resource Manager project resource schema.",
+		"main":                "Resource Manager project resource schema. To use this resource, it is required that you set the service account email in the provider configuration.",
 		"id":                  "Terraform's internal resource ID. It is structured as \"`container_id`\".",
+		"project_id":          "Project UUID identifier. This is the ID that can be used in most of the other resources to identify the project.",
 		"container_id":        "Project container ID. Globally unique, user-friendly identifier.",
-		"parent_container_id": "Parent container ID",
+		"parent_container_id": "Parent resource identifier. Both container ID (user-friendly) and UUID are supported",
```
joaopalet, line 111: So, if I undertood correctly, in the provider we have:
- `project_id`: the UUID of the project (read-only)
- `container_id`: the user-friendly ID of the projecy (read-only)
- but `parent_container_id`: can be either a UUID or a user-friendly ID (required)

In the response of the API there's:
- `resp.ProjectId`: matches the TFP `project_id`
- `resp.ContainerId`: matches the TFP `container_id`
- `resp.Parent.Id`: the parent UUID (is mapped to `parent_container_id` if `parent_container_id` is a UUUID)
- `resp.Parent.ContainerId`: the parent user-friendly ID (is mapped to `parent_container_id` if `parent_container_id` is a user-fiendly ID)

joaopalet: I think it looks good on a functional perspective, but the naming conventions can be a bit confusing

It seems that the API calls `ContainerId` to the user-friendly IDs, and we follow that for the `container_id` resource field. 

What do you think of renaming `parent_container_id` to just `parent` (as it can either be a UUID or user-friendly ID)?

stackit/internal/services/resourcemanager/resourcemanager_acc_test.go
```diff
@@ -29,16 +30,22 @@ func resourceConfig(name string, label *string) string {
 		labelConfig = fmt.Sprintf("new_label = %q", *label)
 	}
 	return fmt.Sprintf(`
-				%s
+				%[1]s
 
-				resource "stackit_resourcemanager_project" "project" {
-					parent_container_id = "%s"
-					name = "%s"
+				resource "stackit_resourcemanager_project" "project_container_id" {
```
joaopalet, line 35:
```suggestion
				resource "stackit_resourcemanager_project" "parent_by_container_id" {
```

stackit/internal/services/resourcemanager/resourcemanager_acc_test.go
```diff
@@ -29,16 +30,22 @@ func resourceConfig(name string, label *string) string {
...
-						"billing_reference" = "%s"
-						%s
+						"billing_reference" = "%[4]s"
+						%[5]s
 					}
-					owner_email = "%s"
+					owner_email = "%[6]s"
+				}
+
+				resource "stackit_resourcemanager_project" "project_uuid" {
```
joaopalet, line 45:
```suggestion
				resource "stackit_resourcemanager_project" "parent_by_uuid" {
```

Makefile
```diff
@@ -24,6 +24,9 @@ generate-docs:
 	@echo "Generating documentation with tfplugindocs"
 	@$(SCRIPTS_BASE)/tfplugindocs.sh
 
+build:
+	@go build -o bin/terraform-provider-stackit
```
joaopalet, line 28: 👍

stackit/internal/services/resourcemanager/project/resource.go
vicentepinto98: I agree, but that would be a breaking change. 
For now, this PR just introduces the changes needed to fix the existing bug without breaking.
We will check how to proceed with supporting both these ID's. Ideally we should only support one, to avoid confusion

joaopalet (approved):

PR #198, opened 2024-01-09
Update core and key flow documentation

vicentepinto98:


README.md
```diff
@@ -68,10 +65,17 @@ The expected format of the service account key is a **json** with the following
...
+
+   - setting the fiels in the provider block: `service_account_key` or `service_account_key_path`
+   - setting the environment variable: `STACKIT_SERVICE_ACCOUNT_KEY_PATH`
+   - setting `STACKIT_SERVICE_ACCOUNT_KEY_PATH` in the credentials file (see above)
+
+4. **If you have provided your own RSA key-pair**, you can set it the same way (it will take precedence over the private key included in the service account key, if present):
+
+   - setting the fiels in the provider block: `private_key` or `private_key_path`
+   - setting the environment variable: `STACKIT_PRIVATE_KEY_PATH`
+   - setting `STACKIT_PRIVATE_KEY_PATH` in the credentials file (see above)
```
joaopalet, line 78: I feel like this shouldn't be a 4th point of the previous steps, it could either be:
- A different section (e.g. after `## Key Flow` having something like `### Generation a new SA Key` and `### Using your own RSA key pair`)
- Or (I think this one is easier to read) make it something like a note. E.g.:

```suggestion
> **If you have provided your own RSA key-pair**, you can set it the same way (it will take precedence over the private key included in the service account key, if present):

   - setting the fiels in the provider block: `private_key` or `private_key_path`
   - setting the environment variable: `STACKIT_PRIVATE_KEY_PATH`
   - setting `STACKIT_PRIVATE_KEY_PATH` in the credentials file (see above)
```

joaopalet: Also I am a bit confused by the phrasing, especially `**If you have provided your own RSA key-pair**, you can set it the same way`

Do you mean that if the user wants to use their own key-pair and provide the private key separately, they can follow a similar approach as to setting the service account key (provider config, env. variable or configuration file)?

vicentepinto98: Adapted it and also rephrased

README.md
```diff
@@ -68,10 +65,17 @@ The expected format of the service account key is a **json** with the following
...
+
+   - setting the fiels in the provider block: `service_account_key` or `service_account_key_path`
+   - setting the environment variable: `STACKIT_SERVICE_ACCOUNT_KEY_PATH`
+   - setting `STACKIT_SERVICE_ACCOUNT_KEY_PATH` in the credentials file (see above)
+
+> **If you have provided your own RSA key-pair when creating the service account key**, you also need to configure it (takes precedence over the private key included in the service account key, if present):
+>
+> - setting the fiels in the provider block: `private_key` or `private_key_path`
+> - setting the environment variable: `STACKIT_PRIVATE_KEY_PATH`
+> - setting `STACKIT_PRIVATE_KEY_PATH` in the credentials file (see above)
```
joaopalet, line 78:
```suggestion
> **If you have provided your own RSA key-pair when creating the service account key**, you also need to configure your private key (takes precedence over the private key included in the service account key, if present):
>
> - setting the field in the provider block: `private_key` or `private_key_path`
> - setting the environment variable: `STACKIT_PRIVATE_KEY_PATH`
> - setting `STACKIT_PRIVATE_KEY_PATH` in the credentials file (see above)

```

joaopalet: This doesn't seem to show what changed, but I suggested:
- `you also need to configure it` --> `you also need to configure your private key`
- `setting the fiels` --> `setting the field`

joaopalet (approved):

PR #206, opened 2024-01-12
Move checkAllowPriviledgedContainers to the Create and Update functions in the SKE resource

### LINKED ISSUE #201: cannot upgrade SKE version 1.24 to SKE version 1.25
# Issue Description
I already have SKE cluster with version 1.24, I'm trying to upgrade it to 1.25 but it fails with this error
```
This field is required for Kubernetes prior to 1.25
```
If I set the `kubernetes_version` directly to 1.25 without using a variable I get no errors.

# current behaviour
Supplying kubernetes version via a variable causes validation error

# expected behaviour
Supplying kubernetes version via a variable should not cause a validation error

# Steps to re-produce

- Create a new cluster with this code
```
resource "stackit_ske_cluster" "example" {
  project_id                  = "project ID
  name                        = "example"
  allow_privileged_containers = true
  kubernetes_version          = "1.24"
  node_pools = [
    {
      name               = "example"
      machine_type       = "m1.3"
      os_version         = var.os_version
      minimum            = "2"
      maximum            = "3"
      availability_zones = ["eu01-m"]
    }
  ]
}
```
- Update the code to use a variable for `kubernetes_version`
```
variable "k8s_version" {
  description = "The k8s version we use for SKE clusters"
  type        = string
  default     = "1.25"
}
resource "stackit_ske_cluster" "example" {
  project_id         = stackit_resourcemanager_project.api.id
  name               = "example1"
  kubernetes_version = var.k8s_version
  node_pools = [
    {
      name               = "example"
      machine_type       = "m1.3"
      os_version         = var.os_version
      minimum            = "2"
      maximum            = "3"
      availability_zones = ["eu01-m"]
    }
  ]
}
```
- Run `terraform plan` you will get this error
```
│ Error: 'Allow privilege containers' missing
│ 
│   with stackit_ske_cluster.example,
│   on ske.tf line 52, in resource "stackit_ske_cluster" "example":
│   52: resource "stackit_ske_cluster" "example" {
│ 
│ This field is required for Kubernetes prior to 1.25
```
- Set the version directly using `kubernetes_version = 1.25` and try to plan, no error now

Using a local value for kubernetes_version that read from the variable causes the same issue

The reason behind this could be that `ValidateConfig` is called before variables are rendered to their value, which means such checks should not be placed in `ValidateConfig`

vicentepinto98:
Fixes #201

### DIFF
--- a/stackit/internal/services/ske/cluster/resource.go
+++ b/stackit/internal/services/ske/cluster/resource.go
@@ -512,6 +512,9 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 	}
 }
 
+// needs to be executed inside the Create and Update methods
+// since ValidateConfig runs before variables are rendered to their value,
+// which causes errors like this: https://github.com/stackitcloud/terraform-provider-stackit/issues/201
 func checkAllowPrivilegedContainers(allowPrivilegeContainers types.Bool, kubernetesVersion types.String) diag.Diagnostics {
 	var diags diag.Diagnostics
 
--- a/stackit/internal/services/ske/cluster/resource.go
+++ b/stackit/internal/services/ske/cluster/resource.go
@@ -512,21 +512,6 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 	}
 }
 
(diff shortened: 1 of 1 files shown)


joaopalet (approved):
Looks like a bug with the Terraform `ValidateConfig` function, right? Maybe we could add a comment saying why it has to be in the `Create` and `Update` functions

joaopalet (approved):

PR #255, opened 2024-02-07
Add new resource and datasource for Load Balancer credentials

joaopalet:


stackit/internal/services/loadbalancer/credential/resource.go
```diff
@@ -0,0 +1,353 @@
...
+				},
+			},
+			"password": schema.StringAttribute{
+				Description: descriptions["password"],
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"credentials_ref": schema.StringAttribute{
```
hcsa73, line 146: Since this is doing what `[RESOURCE]_id` does usually,
- Nitpick: put it after the `id` field in the schema, so the identifiers are at the top
- Does it make sense to call it `credential_id`, for uniformity with other resources? Or should we prefer uniformity with the API? Not sure...

stackit/internal/services/loadbalancer/credential/resource.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// Get status of load balancer functionality
+	statusResp, err := r.client.GetServiceStatus(ctx, projectId).Execute()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error getting status of load balancer functionality", fmt.Sprintf("Calling API: %v", err))
+		return
+	}
+
+	// If load balancer functionality is not enabled, enable it
+	if *statusResp.Status != wait.FunctionalityStatusReady {
```
hcsa73, line 177: It feels wrong to import this from the `wait` package, isn't it in a better place in the SDK?

stackit/internal/services/loadbalancer/credential/resource.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// Create new credentials
+	createResp, err := r.client.CreateCredentials(ctx, projectId).CreateCredentialsPayload(*payload).XRequestID(uuid.NewString()).Execute()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating credential", fmt.Sprintf("Calling API: %v", err))
+		return
+	}
+
+	// Map response body to schema
+	err = mapFields(createResp.Credential, &model)
```
hcsa73, line 206: Add `ctx = tflog.SetField(ctx, "credentials_ref", [CREDENTIALS REF FROM RESPONSE])` after credential creation

stackit/internal/services/loadbalancer/credential/datasource.go
```diff
@@ -0,0 +1,200 @@
...
+
+// Schema defines the schema for the data source.
+func (r *credentialDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	descriptions := map[string]string{
+		"main":            "Load balancer credential data source schema. Must have a `region` specified in the provider configuration.",
+		"id":              "Terraform's internal resource ID. It is structured as \"`project_id`\",\"`credentials_ref`\".",
+		"project_id":      "STACKIT project ID to which the load balancer credential is associated.",
+		"display_name":    "Credential name.",
+		"username":        "The username used for the ARGUS instance.",
+		"password":        "The password used for the ARGUS instance.",
```
hcsa73, line 92: There's no password for data source
Does it make sense to have it as a data source?

stackit/internal/services/loadbalancer/loadbalancer_acc_test.go
```diff
@@ -83,7 +86,14 @@ func configResources(targetPort string) string {
 			options = {
 				private_network_only = %s
 			}
-			}
+		}
+
+		resource "stackit_loadbalancer_credential" "credential" {
+			project_id   = "%[3]s"
```
hcsa73, line 92: Didn't know about this, usually we just put the project ID multiple times

stackit/internal/services/loadbalancer/credential/datasource.go
joaopalet: Good point, probably not right? Was in autopilot implementing it here

stackit/internal/services/loadbalancer/credential/resource.go
joaopalet: 1. I'll move it to the top
2. Good point. But I think generally when we have `...Id` the field is a UUID and in this case it's not, and for some resources the identifier is a `name`

stackit/internal/services/loadbalancer/credential/resource.go
joaopalet: Unfortunately not, we are just defining the possible response strings in `wait.go`

stackit/internal/services/loadbalancer/credential/resource.go
joaopalet: Moved it to the top but it didn't produce any difference in the docs (it's ordered alphabetically)

stackit/internal/services/loadbalancer/credential/datasource.go
joaopalet: Removed

PR #264, opened 2024-02-13
Fix loadbalancer internal models to use TF types

hcsa73:


stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -642,31 +671,42 @@ func (r *loadBalancerResource) Update(ctx context.Context, req resource.UpdateRe
...
-	for _, targetPool := range model.TargetPools {
+	targetPools := []targetPool{}
+	diags = model.TargetPools.ElementsAs(ctx, &targetPools, false)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	for i := range targetPools {
+		targetPool := targetPools[i]
+		targetPoolName := model.Name.ValueString()
```
vicentepinto98, line 682:
```suggestion
		targetPoolName := targetPool.Name.ValueString()
```

stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -642,31 +671,42 @@ func (r *loadBalancerResource) Update(ctx context.Context, req resource.UpdateRe
 	ctx = tflog.SetField(ctx, "project_id", projectId)
 	ctx = tflog.SetField(ctx, "name", name)
 
-	for _, targetPool := range model.TargetPools {
+	targetPools := []targetPool{}
```
vicentepinto98, line 674: In some of the other fields, you call this variable  <field>Model but here it doesn't have the "Model" suffix

I would actually remove the "Model" suffix in the other variable names, seems redundant

vicentepinto98 (changes requested):

stackit/internal/services/loadbalancer/loadbalancer/resource.go
hcsa73: I've renamed this variable to `targetPoolModel`, let me know if you want me to revert this and rename the other `...Model` variables

vicentepinto98: fine for me

vicentepinto98 (approved):

Issue #261, opened 2024-02-13
`stackit_dns_record_set` enforces a valid IP Address always


JorTurFer:
`stackit_dns_record_set` can be used for multiple record types, and not all of them are IP Address (e.g: CNAME, TXT,..) but the resource requires always the IP validation:
https://github.com/stackitcloud/terraform-provider-stackit/blob/28907ba33e0a1211ff03ee5a44729d66eccf6ed2/stackit/internal/services/dns/recordset/resource.go#L154-L163

The problem is that the validation doesn't check if the type of the record requires it either:
https://github.com/stackitcloud/terraform-provider-stackit/blob/28907ba33e0a1211ff03ee5a44729d66eccf6ed2/stackit/internal/validate/validate.go#L59-L74

As result of this, I'm getting a validation error when I try to set a TXT record:
`Attribute records[0] value must be an IP address, got: v=spf1 a mx include:xxxxxxxxx ~all`
and almost the same for a CNAME record:
`Attribute records[0] value must be an IP address, got: xxxxxxxxxx.stackit.org`

JorTurFer:
you can assign it to me. I'lll work on it asap xD

vicentepinto98:
This issue is fixed on the [v0.13.0](https://github.com/stackitcloud/terraform-provider-stackit/releases/tag/v0.13.0) release.

PR #267, opened 2024-02-14
fix:  validates the record conent based on record type

JorTurFer:
I've added a new validation `RecordSet` which takes into account the type type for validating the record value

Fixes https://github.com/stackitcloud/terraform-provider-stackit/issues/261

### DIFF
--- a/stackit/internal/validate/validate.go
+++ b/stackit/internal/validate/validate.go
@@ -115,7 +115,6 @@ func RecordSet() *Validator {
 					req.ConfigValue.ValueString(),
 				))
 			}
-
 		},
 	}
 }
--- a/stackit/internal/validate/validate.go
+++ b/stackit/internal/validate/validate.go
@@ -108,12 +108,8 @@ func RecordSet() *Validator {
 			case "ALIAS":
 			case "DNAME":
 			case "CAA":
+			case "PTR":
 			default:
-				resp.Diagnostics.Append(validatordiag.InvalidAttributeValueDiagnostic(
-					req.Path,
-					"invalid record type",
-					req.ConfigValue.ValueString(),
-				))
 			}
 		},
 	}
--- a/stackit/internal/validate/validate.go
+++ b/stackit/internal/validate/validate.go
@@ -10,7 +10,9 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-framework-validators/helpers/validatordiag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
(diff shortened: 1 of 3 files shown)


stackit/internal/validate/validate.go
```diff
@@ -73,6 +75,50 @@ func IP() *Validator {
...
+						req.ConfigValue.ValueString(),
+					))
+				}
+			case "CNAME":
+			case "NS":
+			case "MX":
+			case "TXT":
+			case "ALIAS":
+			case "DNAME":
+			case "CAA":
```
vicentepinto98, line 110:
```suggestion
			case "CAA":
			case "PTR":
```

There's one other supported record type

stackit/internal/validate/validate.go
```diff
@@ -73,6 +75,50 @@ func IP() *Validator {
...
+			default:
```
vicentepinto98, line 111: I would leave the default as empty instead of throwing error. We are not validating the type field in the schema, which means we also cannot do it here.

There may be other supported record types introduced by the API that we can't guarantee to maintain here

stackit/internal/validate/validate_test.go
```diff
@@ -101,6 +104,157 @@ func TestIP(t *testing.T) {
...
+			true,
+		},
+		{
+			"CAA record",
+			"some-record",
+			"CAA",
+			true,
+		},
+		{
+			"random record",
```
vicentepinto98, line 217: following my comment about default validation, this should be true

stackit/internal/validate/validate_test.go
```diff
@@ -101,6 +104,157 @@ func TestIP(t *testing.T) {
...
+			"some-record",
+			"DNAME",
+			true,
+		},
+		{
+			"CAA record",
+			"some-record",
+			"CAA",
+			true,
+		},
```
vicentepinto98, line 215: following my comment about adding PTR type, add the test case here

vicentepinto98 (changes requested):

stackit/internal/validate/validate.go
JorTurFer: I've checked and StackIT portal doesn't support it yet, I can add it to the code, but I prefer to ask before just in case

stackit/internal/validate/validate.go
JorTurFer: It's a good point!

stackit/internal/validate/validate.go
vicentepinto98: sorry my bad, then let's leave it like this. Might be an unreleased feature

vicentepinto98: can you remove it? with no validation by default there is no problem

JorTurFer: done

vicentepinto98 (approved):

JorTurFer:
Could I ask for a release including this feature? We are using my own fork to not be blocked xD

Issue #273, opened 2024-02-21
Enhancment: Improve SKE objects


PeterStolz:
I had some unexpected friction when creating a cluster via terraform.
Here is a list of points I noticed:
- When specifying an invalid kubernetes version I am not told which versions are available. If you do that check please log them so I don't need to try multiple times which one is the latest one. `│ Creating cluster config API payload: getting latest matching kubernetes version: provided version is not one of the available kubernetes versions`. For the versions in the node pool I had to go to the UI and look at the values in the dropdown. That is not the best DX.
- Resource type stackit_ske_cluster needs the id of a stackit_ske_project not a stackit_project_id. This should be better documented

vicentepinto98:
Hello @PeterStolz, thanks for opening the issue and the suggestions. We have put this in our backlog and will try to tackle it soon, it should be relatively easy to do.

vicentepinto98:
Regarding your second point, which `id` are you referring to?
`stackit_ske_cluster.project_id` should be set to a STACKIT project ID, so the current description of the field is accurate: [`STACKIT project ID to which the cluster is associated.`](https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/resources/ske_cluster#project_id)

The `stackit_ske_project` resource has a field `project_id` and a field `id` (internal Terraform resource identifier) but they should all have the same value. So you can either get the project ID from the project resource or set it directly.

Do you have an example where this didn't work?

PeterStolz:
> Regarding your second point, which id are you referring to?

You are correct. I thought I needed the id of the `stackit_ske_project` and not the stackit `project_id` because I need to create a `stackit_ske_project` to be able to create a cluster. This is not explicitly mentioned in the `stackit_ske_cluster` docs so I got it working by creating an stackit_ske_project assuming its id was then needed(in case I have multiple ske_projects?). Looking into the terraform state the `id` and `project_id` are identical.

vicentepinto98:
The `stackit_ske_project` resource allows you to enable the SKE service for your project, you can only have one per project. If you had already enabled the SKE service (e.g. through the Portal) then you wouldn't need to create this resource and you could just provide the `project_id` in the cluster configuration.

As the concept of service enablement is coming to STACKIT and will impact all services, we will probably move the enablement logic into the `stackit_ske_cluster` resource directly and deprecate `stackit_ske_project` in a future release.

PeterStolz:
[...]

PR #278, opened 2024-02-23
Return available ske versions in the error message

vicentepinto98:
Related to #273

### DIFF
--- a/stackit/internal/services/ske/cluster/resource.go
+++ b/stackit/internal/services/ske/cluster/resource.go
@@ -1337,11 +1337,13 @@ func latestMatchingVersion(availableVersions []ske.KubernetesVersion, providedVe
 	}
 
 	var versionUsed *string
+	var availableVersionsArray []string
 	// Get the higher available version that matches the major and minor version provided by the user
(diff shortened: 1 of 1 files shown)


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1337,11 +1337,13 @@ func latestMatchingVersion(availableVersions []ske.KubernetesVersion, providedVe
 	}
 
 	var versionUsed *string
+	var availableVersionsArray []string
```
joaopalet, line 1340: Maybe rename to something more descriptive, like `validAvailableVersions`?

`availableVersionsArray` doesn't tell more than the already existing array `availableVersions`

joaopalet (approved):

stackit/internal/services/ske/cluster/resource.go
vicentepinto98: It's literally just a string array of the `availableVersions`. We just need it to print in the error message. 
`availableVersions` includes more than just the `version`, but I don't think it's necessary to go into those details here

Issue #294, opened 2024-03-08
Data Loss: Deleting one SKE Project resource deletes all SKE Clusters in the STACKIT Project


jebreuer:
Hi all! Me again. :)

I experienced a data loss issue.

Please see the example tf file attached for a minimal example reproducing the issue: [ske-bug.tf.txt](https://github.com/stackitcloud/terraform-provider-stackit/files/14536895/ske-bug.tf.txt)

I define a SKE Project resource for each SKE Cluster. After `terraform apply` the result is as follows:
```
$ stackit ske cluster list -p $redacted

 NAME  │ STATE         │ VERSION │ POOLS │ MONITORING
───────┼───────────────┼─────────┼───────┼────────────
 ske-1 │ STATE_HEALTHY │ 1.27.11 │     1 │ Disabled
 ske-2 │ STATE_HEALTHY │ 1.27.11 │     1 │ Disabled
```

Deleting the resources stackit_ske_cluster.ske-2 and stackit_ske_project.ske-project-2 causes an error after a couple of minutes:
```
[...]
stackit_ske_cluster.ske-2: Destruction complete after 8m12s
stackit_ske_project.ske-project-2: Destroying... [id=e1c540fe-97de-4fa0-a244-c640c98e6021]
╷
│ Error: Error deleting credential
│
│ Calling API: 409 Conflict, status code 409, Body: {"code":"AlreadyExists","message":"already exists: projects","details":""}
│
╵
```

 And the resource stackit_ske_project.ske-1 gets deleted implicitly without consent:
```
stackit ske cluster list -p $redacted

 NAME  │ STATE          │ VERSION │ POOLS │ MONITORING
───────┼────────────────┼─────────┼───────┼────────────
 ske-1 │ STATE_DELETING │ 1.27.11 │     1 │ Disabled
```

Not only do I experience data loss the terraform state is also corrupt and needs manual cleanup (`terraform state rm ...`) for the remaining SKE related resources.

**Observation:**
The resource stackit_ske_project seems to be unique for every STACKIT project. Subsequent instances (here: stackit_ske_project.ske-project-2) seem to be referencing the exact same thing.

**Expectation 1:**
If the observation is true then creating a second instance of stackit_ske_project should fail.

**Expectation 2:**
Deleting a stackit_ske_project that still contains clusters should fail. 
The current behavior is an equivalent of `rm -Rf *` which can cause implicit and harmful data loss.

**Alternatively:**
[...]

Issue #276, opened 2024-02-22
Enhancment: set URI in `stackit_redis_credential` to use SSL and uri to sensitive


PeterStolz:
Right no the `stackit_redis_credential` returns an uri using no ssl but you can enble ssl using `rediss://` instead of `redis://`

Rabbitmq already uses ssl with `amqps`.

Also password is marked as sensitive, but the uri is not. That pattern is consistent across all credentials resources i have used so far including mariadb, redis, rabbitmq.

vicentepinto98:
Hi @PeterStolz, we are looking into it and will get back to you :)

vicentepinto98:
Hello, 
I have asked our colleagues and the returned URI is already the SSL one, for all of the services that you mentioned. 
For redis specifically, you just need to provide the --tls suffix on the connection client, but the connection string is the same.

If you can provide more details of your setup, we can support you better.

PeterStolz:
I know I can enable TLS for my redis connection using --tls. However there are a bunch of libraries that infer this based on the provided connection URI. Some don't support additional flags and rely entirely on the provided URI. Also `rediss` is an official URI scheme.
https://www.iana.org/assignments/uri-schemes/prov/rediss
From a security point of view, as stackit redis supports TLS it should be the default and I should not have to opt in to communicate securely.
Most developers will likely just use the provided uri and do not bother to enable tls on their redis client, as they expect secure defaults. The app I am working with only takes a redis URI as an environment variable meaning that I would have to do brittle string manipulation inside terraform to use secure communication.

I am confident that this is the exact reason that the rabbimq resource defaults to `amqps://` instad of `amqp://` which would also work. It is the same thing with http vs https. You would not post http links nowdays if the server can speak https as well

vicentepinto98:
Thanks for the detailed reply, we will replace the redis URI and return `rediss`. We hope to work on that soon

jebreuer:
Just ran into the same issue as well.

I tried to configure our app with a redis connection. The data from the terraform provider said the uri was `redis://...`. Connecting to that endpoint didn't work in our app so I tried with the redis-cli which just complained about an IO error. Luckily Peter sits across the room so I could ask him for his experience and he tells me that he also stumbled into the same problem and an issue is already open.

Sad to say but currently it's not fun. If you can't provide a fix in a timely manner please **update the documentation** accordingly so it at least saves other people's time.

vicentepinto98:
[...]

PR #298, opened 2024-03-11
Update documentation

vicentepinto98:
Related to #276 and #294

### DIFF
--- a/stackit/internal/services/redis/credential/resource.go
+++ b/stackit/internal/services/redis/credential/resource.go
@@ -103,6 +103,7 @@ func (r *credentialResource) Schema(_ context.Context, _ resource.SchemaRequest,
 		"credential_id": "The credential's ID.",
 		"instance_id":   "ID of the Redis instance.",
 		"project_id":    "STACKIT Project ID to which the instance is associated.",
+		"uri":           "Connection uri. Currently the returned uri doesn't support ssl, but you can enable it by using rediss:// instead of redis://. We will provide the ssl connection in a future release",
 	}
 
 	resp.Schema = schema.Schema{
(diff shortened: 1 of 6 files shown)


stackit/internal/services/redis/credential/resource.go
```diff
@@ -103,6 +103,7 @@ func (r *credentialResource) Schema(_ context.Context, _ resource.SchemaRequest,
 		"credential_id": "The credential's ID.",
 		"instance_id":   "ID of the Redis instance.",
 		"project_id":    "STACKIT Project ID to which the instance is associated.",
+		"uri":           "Connection uri. Currently the returned uri doesn't support ssl, but you can enable it by using rediss:// instead of redis://. We will provide the ssl connection in a future release",
```
joaopalet, line 106:
```suggestion
		"uri":           "Connection URI. Currently the returned uri doesn't support SSL, but you can enable it by using `rediss://` instead of `redis://`. We will provide the SSL connection in a future release",
```

stackit/internal/services/ske/project/resource.go
```diff
@@ -86,7 +86,7 @@ func (r *projectResource) Configure(ctx context.Context, req resource.ConfigureR
 // Schema returns the Terraform schema structure
 func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
 	resp.Schema = schema.Schema{
-		Description: "SKE project resource schema. Must have a `region` specified in the provider configuration.",
+		Description: "SKE project resource schema. Must have a `region` specified in the provider configuration. Deleting this resource will destroy any SKE clusters associated to the project",
```
joaopalet, line 89: Probably it's also worth documenting that only one `stackit_ske_project` resource should be created, and that this only enables the SKE functionality for the given STACKIT project

joaopalet (dismissed):

stackit/internal/services/redis/credential/resource.go
joaopalet: The `datasource` file also has this

joaopalet (approved):

Issue #270, opened 2024-02-15
Document how to use vault provider with stackit secrets manager


not22day:
The following code worked for me
1) how to carry over user/pw from secretsmanager_user into provider
2) skip_child_token = true required to make it work
3) mount is the instance id

```
provider "vault" {
  address = "https://prod.sm.eu01.stackit.cloud"
  skip_child_token = true
  auth_login_userpass {
    username = stackit_secretsmanager_user.example.username
    password = stackit_secretsmanager_user.example.password

  }
}

resource "vault_kv_secret_v2" "example" {
  mount                      = stackit_secretsmanager_instance.example.instance_id
  name                       = "my-secret"
  cas                        = 1
  delete_all_versions        = true
  data_json                  = jsonencode(
  {
    zip       = "zap",
    foo       = "bar"
  }
  )
}
```

DiogoFerrao:
Documentation for this case is available in the [Terraform Registry](https://registry.terraform.io/providers/stackitcloud/stackit/0.15.0/docs/guides/vault_secrets_manager)

PR #310, opened 2024-03-25
Create guide to document Vault usage with Secrets Manager

DiogoFerrao:


docs/guides/vault_secrets_manager.md
```diff
@@ -0,0 +1,98 @@
...
+page_title: "Using Vault Provider with STACKIT Secrets Manager"
+subcategory: ""
+description: |-
+   Guide on how to write secrets from Terraform deployed resources on STACKIT cloud to the STACKIT Secrets Manager 
+---
+# Using Vault Provider with STACKIT Secrets Manager
+
+### Overview
+
+This guide outlines the process of utilizing the HashiCorp Vault provider alongside the STACKIT provider to write secrets into the STACKIT Secrets Manager. Due to the current limitation in the STACKIT Terraform provider where direct writing of secrets isn't supported, this workaround is necessary.
```
vicentepinto98, line 11: I wouldnt say this is a limitation of the provider. And this is not really a workaround. 
In general, the STACKIT services are providing control-plane APIs, that don't support performing operations in the instances themselves such as `write` operations. It's the same for writing stuff into an Object Storage bucket: you need to use the `s3` provider

So I would rephrase this

vicentepinto98: This would be a workaround if the Secrets Manager would plan to add this functionality to their API but I think this will not happen

docs/guides/vault_secrets_manager.md
```diff
@@ -0,0 +1,98 @@
...
+      skip_child_token = true
+
+      auth_login_userpass {
+        username = stackit_secretsmanager_user.example.username
+        password = stackit_secretsmanager_user.example.password
+      }
+    }
+    ```
+
+5. **Define Terraform Resource (Example: DNS Zone)**
```
vicentepinto98, line 65: :+1: for the example

But maybe you could replace it with a more meaningful one such as some user or credentials resource

vicentepinto98 (commented):
Do you know how this will show-up in the Terraform Registry? 

Is this the standard location (`docs/guides`) for` such guides in terraform documentation?

DiogoFerrao:
> Do you know how this will show-up in the Terraform Registry?
> 
> Is this the standard location (`docs/guides`) for` such guides in terraform documentation?

Yes, this is the standard location for Guides, I followed the documentation here: https://developer.hashicorp.com/terraform/registry/providers/docs#format

docs/guides/vault_secrets_manager.md
```diff
@@ -8,15 +8,7 @@ description: |-
...
-This guide outlines the process of utilizing the HashiCorp Vault provider alongside the STACKIT provider to write secrets into the STACKIT Secrets Manager. Due to the current limitation in the STACKIT Terraform provider where direct writing of secrets isn't supported, this workaround is necessary.
-
-### Problem
-
-Writing credentials from provisioned STACKIT services into a Secrets Manager instance is currently not supported by the STACKIT Terraform provider.
-
-### Solution
-
-To overcome this limitation, you can integrate the HashiCorp Vault provider with the STACKIT provider. This solution involves configuring the Vault provider to interface with the STACKIT Secrets Manager.
+This guide outlines the process of utilizing the HashiCorp Vault provider alongside the STACKIT provider to write datasources from STACKIT cloud resources (deployed with Terraform) as secrets in the STACKIT Secrets Manager.
```
vicentepinto98, line 8: It's not restricted to data from STACKIT resources. Any secret can be written into the instance

docs/guides/vault_secrets_manager.md
```diff
@@ -87,12 +74,12 @@ To overcome this limitation, you can integrate the HashiCorp Vault provider with
...
+         grafana_password = stackit_argus_instance.example.grafana_initial_admin_password,
         }
       )
     }
     ```
 
 ### Note
 
-This example can be adapted for various resources within the provider by replacing the DNS zone with the appropriate resource and associated variables.
+This example can be adapted for various resources within the provider by replacing the Argus Monitoring Grafana password with the appropriate resource and associated variables.
```
vicentepinto98: For various resources or any other secret that the user wants to set into the Secrets Manager instance

docs/guides/vault_secrets_manager.md
DiogoFerrao: I updated the wording, I think it still makes sense to focus on stackit resources, since I think this will be the main usage of the guide, otherwise its just a guide for the Vault provider. Does this make sense?

vicentepinto98: Yeah that makes sense :+1:

vicentepinto98 (approved):

DiogoFerrao:
Closes #270

Issue #297, opened 2024-03-11
stackit_ske_cluster version change not working


PeterStolz:
Heyho I think an API change might have broken the terraform provider or version changes need recreation of the cluster.

Changes to apply:
```
  # stackit_ske_cluster.ske will be updated in-place
  ~ resource "stackit_ske_cluster" "ske" {
        id                      = "1027e3d4-e94c-41d7-b497-6804e4cc7e0b,stoffstrom"
      + kube_config             = (sensitive value)
      ~ kubernetes_version      = "1.27" -> "1.26"
      ~ kubernetes_version_used = "1.27.11" -> (known after apply)
        name                    = "stoffstrom"
        # (3 unchanged attributes hidden)
    }
 ```
 
 Error log:
 ```

stackit_ske_cluster.ske: Modifying... [id=1027e3d4-e94c-41d7-b497-6804e4cc7e0b,stoffstrom]
╷
│ Error: Error creating/updating cluster
│ 
│   with stackit_ske_cluster.ske,
│   on main.tf line 83, in resource "stackit_ske_cluster" "ske":
│   83: resource "stackit_ske_cluster" "ske" {
│ 
│ Calling API: 400 Bad Request, status code 400, Body: {"code":"InvalidArgument","message":"invalid input: Kubernetes version \"1.26.14\" is not available. Supported: [], Preview: []","details":""}
│ 
╵
```

vicentepinto98:
Hi @PeterStolz,
Kubernetes version downgrade for a cluster is not supported by SKE.
The error could be more descriptive but it's not a bug.

PeterStolz:
That is totally fine for me but then terraform should recreate it, instead of failing in place replacment

vicentepinto98:
I agree, we should replace the resource in case of a downgrade. I re-opened the issue and we will provide a fix soon

Kumm-Kai:
The SKE-API now also returns a proper error when trying to downgrade the kubernetes version.

PR #315, opened 2024-03-27
Bugfix SKE downgrade replace cluster

DiogoFerrao:


stackit/internal/services/ske/ske_acc_test.go
```diff
@@ -509,6 +509,50 @@ func TestAccSKE(t *testing.T) {
 					resource.TestCheckNoResourceAttr("stackit_ske_cluster.cluster", "kube_config"), // when using the kubeconfig resource, the kubeconfig field becomes null
 				),
 			},
+			// 7) Downgrade kubernetes version
+			{
+				Config: getConfig(clusterResource["kubernetes_version"], utils.Ptr(clusterResource["maintenance_end_new"])),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// cluster data
+					resource.TestCheckResourceAttr("stackit_ske_cluster.cluster", "project_id", clusterResource["project_id"]),
```
joaopalet, line 517: Maybe it's enough to test the kubernetes version fields here

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -269,6 +269,17 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+					stringplanmodifier.RequiresReplaceIf(stringplanmodifier.RequiresReplaceIfFunc(func(ctx context.Context, sr planmodifier.StringRequest, rrifr *stringplanmodifier.RequiresReplaceIfFuncResponse) {
+						if sr.StateValue.IsNull() || sr.PlanValue.IsNull() {
+							return
+						}
+						planVersion := fmt.Sprintf("v%s", sr.PlanValue.ValueString())
+						stateVersion := fmt.Sprintf("v%s", sr.StateValue.ValueString())
+
+						rrifr.RequiresReplace = semver.Compare(planVersion, stateVersion) < 0
+					}), "Kubernetes version", "If the Kubernetes version is a downgrade, the cluster will be replaced"),
+				},
```
joaopalet, line 282: I was wondering if it's worth implementing this is a new package like we did for the `validate` package (for validators, e.g. `validate.NoSeparator()`) but I think we can keep it here for now and then extract when we need to re-use it

DiogoFerrao: It would be a cleaner approach, but I'm not sure the overhead is worth the payoff here. Plan modifiers should be rare, especially compared to input validation.

joaopalet: Yes I agree, if we later decide we want to this for more services we would extract it

joaopalet (approved):

DiogoFerrao:
Closes #297

Issue #326, opened 2024-04-09
DNS: Multi txt records can not be used


Slm0n87:
I try to set a multi-record TXT record:

```
resource "stackit_dns_zone" "example" {
  project_id  = var.project_id
  name        = "domain-example-stackit-zone"
  description = "Just an example to manage dns"
  dns_name    = "domain-example.stackit.zone"
}

resource "stackit_dns_record_set" "example" {
  project_id = var.project_id
  zone_id    = stackit_dns_zone.example.zone_id
  name       = "example-a-record.domain-example.stackit.zone."
  type       = "TXT"
  comment    = "My example Service"
  records    = ["foo", "bar"]
}

```
but get the following error after the apply:

```
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to stackit_dns_record_set.example, provider
│ "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new value:
│ .records[0]: was cty.StringVal("foo"), but now cty.StringVal("bar").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to stackit_dns_record_set.example, provider
│ "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new value:
│ .records[1]: was cty.StringVal("bar"), but now cty.StringVal("foo").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
➜  
```

vicentepinto98:
Hi @Slm0n87 , we have encountered a similar issue before and we will work on it soon

DiogoFerrao:
This issue was addressed in Release [v0.15.3](https://github.com/stackitcloud/terraform-provider-stackit/releases/tag/v0.15.3).
If the problem persists, feel free to reach out and reopen this Issue.

PR #328, opened 2024-04-12
Bugfix inconsistent applies of ListAttributes

DiogoFerrao:


stackit/internal/services/dns/recordset/resource.go
```diff
@@ -417,15 +417,25 @@ func mapFields(recordSetResp *dns.RecordSetResponse, model *Model) error {
 	if recordSet.Records == nil {
 		model.Records = types.ListNull(types.StringType)
 	} else {
-		records := []attr.Value{}
+		respRecords := []string{}
```
vicentepinto98, line 420: :+1:

stackit/internal/utils/utils.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// ReconcileStringLists reconciles two string lists by removing elements from the
+// first list that are not in the second list and appending elements from the
+// second list that are not in the first list.
+// This preserves the order of the elements in the first list that are also in
+// the second list, which is useful when using ListAttributes in Terraform.
+// The source of truth is the second list.
+func ReconcileStringLists(list1, list2 []string) []string {
```
vicentepinto98, line 17:
```suggestion
func ReconcileStringSlices(slice1, slice2 []string) []string {
```
More accurate like this

stackit/internal/utils/utils.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// ReconcileStringLists reconciles two string lists by removing elements from the
+// first list that are not in the second list and appending elements from the
+// second list that are not in the first list.
+// This preserves the order of the elements in the first list that are also in
+// the second list, which is useful when using ListAttributes in Terraform.
+// The source of truth is the second list.
```
vicentepinto98, line 16:
```suggestion
// The source of truth for the order is the first list and the source of truth for the content is the second list.
```

stackit/internal/utils/utils_test.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+		{
+			"common elements with empty string",
+			[]string{"d", "", "c"},
+			[]string{"", "c", "d"},
+			[]string{"d", "", "c"},
+		},
+		{
+			"common elements with duplicates",
+			[]string{"a", "b", "c", "c"},
```
vicentepinto98, line 57: did you get to test some resource with duplicates to check if there is any error?

stackit/internal/services/dns/recordset/resource.go
```diff
@@ -417,15 +417,25 @@ func mapFields(recordSetResp *dns.RecordSetResponse, model *Model) error {
...
+
+		modelRecords, err := utils.ListValuetoStringSlice(model.Records)
+		if err != nil {
+			return err
 		}
-		recordsList, diags := types.ListValue(types.StringType, records)
+
+		reconciledRecords := utils.ReconcileStringLists(modelRecords, respRecords)
+
+		recordsTF, diags := types.ListValueFrom(context.Background(), types.StringType, reconciledRecords)
```
vicentepinto98, line 433: we should use the context coming from the caller method, instead of creating a new one. Pass it as an argument to `mapFields`

stackit/internal/services/rabbitmq/credential/resource.go
```diff
@@ -356,48 +356,66 @@ func mapFields(credentialsResp *rabbitmq.CredentialsResponse, model *Model) erro
...
-			}
-			httpApiUrisList, diags := types.ListValue(types.StringType, httpApiUris)
+			respHttpApiUris := *credentials.HttpApiUris
+
+			reconciledHttpApiUris := utils.ReconcileStringLists(modelHttpApiUris, respHttpApiUris)
+
+			httpApiUrisTF, diags := types.ListValueFrom(context.Background(), types.StringType, reconciledHttpApiUris)
 			if diags.HasError() {
-				return fmt.Errorf("failed to map httpApiUris: %w", core.DiagsToError(diags))
+				return fmt.Errorf("failed to map hosts: %w", core.DiagsToError(diags))
```
vicentepinto98, line 396:
```suggestion
				return fmt.Errorf("failed to map httpApiUris: %w", core.DiagsToError(diags))
```

stackit/internal/utils/utils_test.go
DiogoFerrao: Yes, probably depends on API, but DNS allows for repeated items in the request and just ignores the fact that it is repeated.

vicentepinto98 (approved):

DiogoFerrao:
Closes #326

PR #340, opened 2024-04-19
Remove deleted resources from Terraform state on `Read` (LB and DSA)

joaopalet:


stackit/internal/services/loadbalancer/loadbalancer/datasource.go
```diff
@@ -319,6 +321,11 @@ func (r *loadBalancerDataSource) Read(ctx context.Context, req datasource.ReadRe
 
 	lbResp, err := r.client.GetLoadBalancer(ctx, projectId, name).Execute()
 	if err != nil {
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
```
vicentepinto98, line 324: for datasources we should throw error if it doesn't exist

stackit/internal/services/logme/instance/resource.go
```diff
@@ -299,6 +301,11 @@ func (r *instanceResource) Read(ctx context.Context, req resource.ReadRequest, r
 
 	instanceResp, err := r.client.GetInstance(ctx, projectId, instanceId).Execute()
 	if err != nil {
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
+		if ok && (oapiErr.StatusCode == http.StatusNotFound) || (oapiErr.StatusCode == http.StatusGone) {
+			resp.State.RemoveResource(ctx)
```
vicentepinto98, line 306: if the resource was never in the state, this function doesn't do anything and doesn't throw error right?

stackit/internal/services/logme/instance/resource.go
```diff
@@ -299,6 +301,11 @@ func (r *instanceResource) Read(ctx context.Context, req resource.ReadRequest, r
 
 	instanceResp, err := r.client.GetInstance(ctx, projectId, instanceId).Execute()
 	if err != nil {
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
+		if ok && (oapiErr.StatusCode == http.StatusNotFound) || (oapiErr.StatusCode == http.StatusGone) {
```
vicentepinto98, line 305: `if ok && ((oapiErr.StatusCode == http.StatusNotFound) || (oapiErr.StatusCode == http.StatusGone))`

This is the same right? Because Go is lazy so I was wondering if it would check the first AND and ignore the last part

vicentepinto98 (changes requested):

stackit/internal/services/logme/instance/resource.go
joaopalet: Yes

stackit/internal/services/logme/instance/resource.go
joaopalet: Actually I think yes, but I meant to have the parenthesis wrapping the two last statements (for readability and to not rely on that prioritization of operations), good catch

stackit/internal/services/loadbalancer/loadbalancer/datasource.go
vicentepinto98: this is still missing

stackit/internal/services/mariadb/instance/datasource.go
```diff
@@ -168,9 +168,8 @@ func (r *instanceDataSource) Read(ctx context.Context, req datasource.ReadReques
 	instanceResp, err := r.client.GetInstance(ctx, projectId, instanceId).Execute()
 	if err != nil {
 		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
-		if ok && (oapiErr.StatusCode == http.StatusNotFound) || (oapiErr.StatusCode == http.StatusGone) {
+		if ok && (oapiErr.StatusCode == http.StatusNotFound || oapiErr.StatusCode == http.StatusGone) {
 			resp.State.RemoveResource(ctx)
```
vicentepinto98, line 172: We don't need to remove it from the state, the fix in this PR doesn't apply to datasources in general. 
If the user is reading something that got deleted, it should receive an error and then the user needs to remove the datasource block from his configuration. 
Removing it from the state will not make a difference

vicentepinto98 (changes requested):

stackit/internal/services/mariadb/instance/datasource.go
joaopalet: Discussed offline, we should still throw an error but removing it from the state is also a valid use case so the user doesn't have do to it manually

stackit/internal/services/redis/credential/datasource.go
```diff
@@ -159,6 +161,11 @@ func (r *credentialDataSource) Read(ctx context.Context, req datasource.ReadRequ
 
 	recordSetResp, err := r.client.GetCredentials(ctx, projectId, instanceId, credentialId).Execute()
 	if err != nil {
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
+		if ok && oapiErr.StatusCode == http.StatusNotFound {
+			resp.State.RemoveResource(ctx)
+			return
```
vicentepinto98, line 167: all credential datasources are returning before throwing error, is that on purpose?

stackit/internal/services/loadbalancer/loadbalancer/datasource.go
```diff
@@ -319,6 +321,11 @@ func (r *loadBalancerDataSource) Read(ctx context.Context, req datasource.ReadRe
...
+		if ok && oapiErr.StatusCode == http.StatusNotFound {
+			resp.State.RemoveResource(ctx)
+			return
```
vicentepinto98, line 327: also returning here, should throw error

vicentepinto98 (approved):

PR #349, opened 2024-04-26
Fix Argus scrape config internal models to use Terraform types

joaopalet:


stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -582,47 +709,52 @@ func toCreatePayload(ctx context.Context, model *Model) (*argus.CreateScrapeConf
...
 
-	if model.SAML2 != nil && !model.SAML2.EnableURLParameters.ValueBool() {
+	if !saml2Model.EnableURLParameters.IsNull() && !saml2Model.EnableURLParameters.IsUnknown() {
 		m := make(map[string]interface{})
 		if sc.Params != nil {
 			m = *sc.Params
 		}
-		m["saml2"] = []string{"disabled"}
+		if saml2Model.EnableURLParameters.ValueBool() {
+			m["saml2"] = []string{"enabled"}
```
vicentepinto98, line 720: isn't this already set on `defaultsCreateScrapeConfig`?

vicentepinto98: then it could be added to the if clause like on the original code

stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -582,47 +709,52 @@ func toCreatePayload(ctx context.Context, model *Model) (*argus.CreateScrapeConf
...
-		for _, v := range target.URLs {
-			tgts = append(tgts, v.ValueString())
+
+		urls := []string{}
+		diags := target.URLs.ElementsAs(ctx, &urls, false)
+		if diags.HasError() {
+			return nil, core.DiagsToError(diags)
 		}
-		ti.Targets = &tgts
+		ti.Targets = &urls
```
vicentepinto98, line 743: :+1: for nicer variable names

vicentepinto98 (commented):
did you run acceptance test?

stackit/internal/services/argus/scrapeconfig/resource.go
joaopalet: Yes, but before we were relying on the default to be true, this now works for both possible default cases

vicentepinto98 (approved):

PR #360, opened 2024-05-13
Support automatic kubernetes updates for stackit_ske_cluster

vicentepinto98:


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -262,13 +266,35 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
 				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
 				Validators: []validator.String{
 					validate.NoSeparator(),
 				},
 			},
+			"kubernetes_version_min": schema.StringAttribute{
+				Description: "The minimum Kubernetes version. This field will be used to set the kubernetes version on creation/update of the cluster. " + SKEUpdateDoc + " To get the current kubernetes version being used for your cluster, use the read-only `kubernetes_version_used` field. If unset, the latest supported Kubernetes version will be used.",
```
DiogoFerrao, line 277: I think this description and attribute name is problematic if we allow the user to downgrade versions. Maybe we should call this attribute "kubernetes_version_initial" or something similar, since it isn't really the minimum version if a user can downgrade a version after creation

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1343,41 +1383,74 @@ func toKubernetesPayload(m *Model, availableVersions []ske.KubernetesVersion) (k
...
 
-	if providedVersion == nil {
-		return nil, false, fmt.Errorf("provided version is nil")
+	if providedVersionMin == nil {
+		if providedVersion == nil {
+			// kubernetes_version field deprecation
+			// this if clause should be removed once kubernetes_version field is completely removed
+			latestVersion, err := getLatestSupportedKubernetesVersion(availableVersions)
+			if err != nil {
+				return nil, false, fmt.Errorf("get latest supporter kubernetes version: %w", err)
```
DiogoFerrao, line 1399:
```suggestion
				return nil, false, fmt.Errorf("get latest supported kubernetes version: %w", err)
```

stackit/internal/services/ske/cluster/resource.go
vicentepinto98: The user cannot downgrade a k8s version of a cluster after creation. This is validated with the `RequiresReplaceIf` function

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -46,6 +47,8 @@ const (
 	VersionStateSupported        = "supported"
 	VersionStatePreview          = "preview"
 	VersionStateDeprecated       = "deprecated"
+
+	SKEUpdateDoc = "SKE automatically updates the cluster Kubernetes version if you have set `maintenance.enable_kubernetes_version_updates` to true or if there is a mandatory update, as described in `Updates for Kubernetes versions and Operating System versions in SKE`(https://docs.stackit.cloud/stackit/en/version-updates-in-ske-10125631.html)."
```
joaopalet, line 51: Should this be a link, like this
```suggestion
	SKEUpdateDoc = "SKE automatically updates the cluster Kubernetes version if you have set `maintenance.enable_kubernetes_version_updates` to true or if there is a mandatory update, as described in [Updates for Kubernetes versions and Operating System versions in SKE](https://docs.stackit.cloud/stackit/en/version-updates-in-ske-10125631.html)."
```
?

stackit/internal/services/ske/cluster/resource.go
DiogoFerrao: But can't a user specify a lower version which requires replace? Then the cluster is replaced and the version is updated, right?

My point is that with the attribute name of `kubernetes_version_min` the user may expect that lowering the minimum version won't change anything, whereas it might even replace the whole cluster if it is a downgrade.

stackit/internal/services/ske/cluster/resource.go
vicentepinto98: good point :+1:

stackit/internal/services/ske/cluster/resource.go
vicentepinto98: "kubernetes_version_initial" it's also not accurate because the version can be updated after the cluster is created. 

I got your point but I don't have an idea for a better name. Not being able to downgrade a kubernetes version is a common concept of SKE in general, but I've made it more explicit in the field description that a downgrade will require a replace of the cluster. I think that should be sufficient and clear for the users

DiogoFerrao (approved):

PR #363, opened 2024-05-17
Fix kubernetes_version_min field logic

vicentepinto98:


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -659,11 +648,27 @@ func (r *clusterResource) loadAvailableVersions(ctx context.Context) ([]ske.Kube
 	return *res.KubernetesVersions, nil
 }
 
-func (r *clusterResource) createOrUpdateCluster(ctx context.Context, diags *diag.Diagnostics, model *Model, availableVersions []ske.KubernetesVersion) {
+// getCurrentVersion makes a call to get the details of a cluster and returns the current kubernetes version
+// if the cluster doesn't exist or some error occurs, returns nil
+func (r *clusterResource) getCurrentVersion(ctx context.Context, m *Model) *string {
```
joaopalet, line 653: Maybe worth to write a small test for this?

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -275,19 +275,8 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				},
 			},
 			"kubernetes_version_min": schema.StringAttribute{
-				Description: "The minimum Kubernetes version. This field will be used to set the kubernetes version on creation/update of the cluster and can only by incremented. A downgrade of the version requires a replace of the cluster.  If unset, the latest supported Kubernetes version will be used. " + SKEUpdateDoc + " To get the current kubernetes version being used for your cluster, use the read-only `kubernetes_version_used` field.",
+				Description: "The minimum Kubernetes version. This field will be used to set the minimum kubernetes version on creation/update of the cluster and can only by incremented. If unset, the latest supported Kubernetes version will be used. " + SKEUpdateDoc + " To get the current kubernetes version being used for your cluster, use the read-only `kubernetes_version_used` field.",
 				Optional:    true,
-				PlanModifiers: []planmodifier.String{
```
joaopalet, line 280: This means there is now no way to downgrade the version right? The user would need to create a new cluster

vicentepinto98: Yes, I thought this would make more sense, because with the logic introduced in this PR, you can actually set a lower version and it will cause no effect, it is in fact a "minimum" version, so not having a `RequiresReplace` would make more sense with the field name. Related also to this [comment](https://github.com/stackitcloud/terraform-provider-stackit/pull/360#discussion_r1599795560).

But It's not so aligned with Terraform logic, where the user may want to re-create a whole new cluster with the same parameters but with a lower version, so we can decided to re-add it

joaopalet: I also think it makes more sense and is more aligned with the name of the field that changing the minimum allowed version won't re-create the cluster

joaopalet (approved):

PR #364, opened 2024-05-17
Support SKE OS versions auto and forceful updates

vicentepinto98:


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -848,7 +919,139 @@ func toNodepoolsPayload(ctx context.Context, m *Model) ([]ske.Nodepool, error) {
...
+	return cnps, deprecatedVersionsUsed, nil
+}
+
+// latestMatchingMachineVersion gets the latest machine image version to be used in the create/update payload
+// it's determined based on the available versions for the input OS (machineName), the minimum version configured by the user and the current version in the cluster
+// if the minimum version is not set, get the current one (if exists) or the latest for the input OS
+// if the minimum version is set but it is a downgrade, use the current version
+// for the minimum version selected, if a patch is not specified, get the latest patch for that minor version
+// for the version returned, check the state and return if it is deprecated or not
+func latestMatchingMachineVersion(availableMachineImages []ske.MachineImage, machineVersionMin *string, machineName string, currentMachineImage *ske.Image) (version *string, deprecated bool, err error) {
```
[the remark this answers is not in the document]
DiogoFerrao: I think we can remove the `machine` from all the variables here, the function name makes it clear its all machine related logic inside. Do you agree?
Then it's less confusing with all the variable names

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -848,7 +919,139 @@ func toNodepoolsPayload(ctx context.Context, m *Model) ([]ske.Nodepool, error) {
...
+	}
+
+	if len(availableMachineImages) == 0 {
+		return nil, false, fmt.Errorf("there are no available machine versions for the provided machine image name %s", machineName)
+	}
+
+	if machineVersionMin == nil {
+		// different machine OS have different versions
+		// so if the current machine is nil or the machina name was updated,  get the latest one
+		// otherwise, the current machine version should be used
```
DiogoFerrao, line 952:
```suggestion
		// different machine OSes have different versions
		// so if the current machine image is nil or the current machine image name was updated, get the latest one
		// otherwise, the current machine image version should be used
```

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -848,7 +919,139 @@ func toNodepoolsPayload(ctx context.Context, m *Model) ([]ske.Nodepool, error) {
...
+		// return the currently used version
+		machineVersionUsed := *currentMachineImage.Version
+		machineVersionMinString := *machineVersionMin
+
+		minVersionPrefixed := "v" + machineVersionMinString
+		usedVersionPrefixed := "v" + machineVersionUsed
+
+		if semver.Compare(minVersionPrefixed, usedVersionPrefixed) == -1 {
+			machineVersionMin = currentMachineImage.Version
+		}
```
DiogoFerrao, line 972:
```suggestion
        // If the os_version_min is set but is lower than the current version used in the cluster,
        // retain the current version to avoid downgrading.
		minimumVersion := "v" + *machineVersionMin
		currentVersion := "v" + *currentMachineImage.Version

		if semver.Compare(minimumVersion, currentVersion) == -1 {
			machineVersionMin = currentMachineImage.Version
		}
```

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -848,7 +919,139 @@ func toNodepoolsPayload(ctx context.Context, m *Model) ([]ske.Nodepool, error) {
...
-	return cnps, nil
+	return cnps, deprecatedVersionsUsed, nil
+}
+
+// latestMatchingMachineVersion gets the latest machine image version to be used in the create/update payload
+// it's determined based on the available versions for the input OS (machineName), the minimum version configured by the user and the current version in the cluster
+// if the minimum version is not set, get the current one (if exists) or the latest for the input OS
+// if the minimum version is set but it is a downgrade, use the current version
+// for the minimum version selected, if a patch is not specified, get the latest patch for that minor version
+// for the version returned, check the state and return if it is deprecated or not
```
DiogoFerrao, line 930:
```suggestion
// latestMatchingMachineVersion determines the latest machine image version for the create/update payload.
// It considers the available versions for the specified OS (OSName), the minimum version configured by the user,
// and the current version in the cluster. The function's behavior is as follows:
//
// 1. If the minimum version is not set:
//    - Return the current version if it exists.
//    - Otherwise, return the latest available version for the specified OS.
//
// 2. If the minimum version is set:
//    - If the minimum version is a downgrade, use the current version instead.
//    - If a patch is not specified for the minimum version, return the latest patch for that minor version.
//
// 3. For the selected version, check its state and return it, indicating if it is deprecated or not.

```

stackit/internal/services/ske/cluster/resource.go
DiogoFerrao, line 952:
```suggestion
        // Different machine OS have different versions.
        // If there is no current machine version or the machine name has been updated,
        // retrieve the latest supported version. Otherwise, use the current machine version.
```

DiogoFerrao:
Unit tests are very extensive, good work there

Shouldn't acceptance tests be extended with a case of this new functionality? Or was it already tested before?

stackit/internal/services/ske/cluster/resource.go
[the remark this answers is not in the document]
vicentepinto98: Fully agree, nice suggestion

stackit/internal/services/ske/cluster/resource.go
vicentepinto98: Really cool re-structuring, thanks

vicentepinto98:
> Unit tests are very extensive, good work there
> 
> Shouldn't acceptance tests be extended with a case of this new functionality? Or was it already tested before?

Good point, I will extend the acceptance test

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -848,7 +919,144 @@ func toNodepoolsPayload(ctx context.Context, m *Model) ([]ske.Nodepool, error) {
...
+			}
+		}
+	}
+
+	if versionUsed != nil {
+		if strings.EqualFold(*state, VersionStateDeprecated) {
+			deprecated = true
+		} else {
+			deprecated = false
+		}
```
DiogoFerrao, line 1026:
```suggestion
		deprecated = strings.EqualFold(*state, VersionStateDeprecated)
```

DiogoFerrao: Is simpler but maybe we lose some clarity. Anyway, if you keep the `if`, the `else` clause should be removed, deprecated is already false.

DiogoFerrao (approved):

PR #368, opened 2024-05-21
Allow specifying network ID in SKE cluster

joaopalet:


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -491,6 +502,22 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+				Optional:    true,
+				Attributes: map[string]schema.Attribute{
+					"id": schema.StringAttribute{
+						Description: "ID of the STACKIT Network Area (SNA) network into which the cluster will be deployed.",
+						Optional:    true,
+						Validators: []validator.String{
+							validate.UUID(),
+						},
+						PlanModifiers: []planmodifier.String{
+							stringplanmodifier.RequiresReplace(),
```
vicentepinto98, line 516: why does it require replace? Did you test it or was it confirmed with the team?

vicentepinto98 (approved):

stackit/internal/services/ske/cluster/resource.go
joaopalet: Confirmed with the team that it can never be changed after creation

PR #375, opened 2024-05-24
Fix cluster network mapping

### LINKED ISSUE #374: Error: Provider produced inconsistent result after apply (v0.18.0)
Hi all, 

after executing `tofu apply` the following error is thrown: 

> Error: Provider produced inconsistent result after apply
> 
>  When applying changes to stackit_ske_cluster.kubernetes_cluster, provider 
>  "provider[\"registry.opentofu.org/stackitcloud/stackit\"]" produced an unexpected new value: .network: was null, but now cty.ObjectVal(map[string]cty.Value{"id":cty.NullVal(cty.String)}).
>  
> This is a bug in the provider, which should be reported in the provider's own issue tracker. 

Used `main.tf` file
```yaml
terraform {
  required_providers {
    stackit = {
      source  = "stackitcloud/stackit"
      version = "0.18.0"
    }
}


resource "stackit_ske_cluster" "kubernetes_cluster" {
  project_id             = var.project_id
  name                   = "example-cluster"
  kubernetes_version_min = "1.28"
  node_pools             = [
    {
      name               = "example-node-pool"
      machine_type       = "b1.2"
      os_name            = "flatcar"
      os_version_min     = "3815.2.1"
      minimum            = 1
      maximum            = 2
      availability_zones = ["eu01-3"]
      volume_size        = 20
      volume_type        = "storage_premium_perf0"
    }
  ]
}
```

`tofu plan` prints the following: 

```
  # stackit_ske_kubeconfig.kubernetes_cluster_config will be created
  + resource "stackit_ske_kubeconfig" "kubernetes_cluster_config" {
      + cluster_name   = "example-cluster"
      + expiration     = 3600
      + expires_at     = (known after apply)
      + id             = (known after apply)
      + kube_config    = (sensitive value)
      + kube_config_id = (known after apply)
      + project_id     = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
      + refresh        = true
    }
```

**Note: Error does not happen when downgrading to Version 0.17.0**

vicentepinto98:
Fixes #374

### DIFF
--- a/stackit/internal/services/ske/cluster/resource.go
+++ b/stackit/internal/services/ske/cluster/resource.go
@@ -505,10 +505,6 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 			"network": schema.SingleNestedAttribute{
 				Description: "Network block as defined below.",
 				Optional:    true,
-				Computed:    true,
-				PlanModifiers: []planmodifier.Object{
-					objectplanmodifier.UseStateForUnknown(),
-				},
 				Attributes: map[string]schema.Attribute{
 					"id": schema.StringAttribute{
 						Description: "ID of the STACKIT Network Area (SNA) network into which the cluster will be deployed.",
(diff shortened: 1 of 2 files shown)


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -505,6 +505,10 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 			"network": schema.SingleNestedAttribute{
 				Description: "Network block as defined below.",
 				Optional:    true,
+				Computed:    true,
```
joaopalet, line 508: What is it exactly that the API returns in this case then? I'm not sure if the field should be Computed as it's not really something that the API can set, right?

vicentepinto98: Yeah you're right, I was mislead by the other nested field `maintenance`, which can actually be set by the API. I will update with another fix, similarly to the `extensions` field, which has the same type and is not `Computed`

joaopalet (approved):

Issue #372, opened 2024-05-23
Removing a nodepool in a multi-nodepool SKE cluster deletes wrong nodepool


maxbischoff:
When removing a nodepool in the `node_pools` of a `stackit_ske_cluster`, that has other nodepools after it in the list, terraform generates a plan similar to this:

```
  # stackit_ske_cluster.api_cluster will be updated in-place
  ~ resource "stackit_ske_cluster" "api_cluster" {
        id                      = "<removed>"
      + kube_config             = (sensitive value)
      ~ kubernetes_version_used = "1.26.15" -> (known after apply)
        name                    = "apidev"
      ~ node_pools              = [
          ~ {
              ~ availability_zones = [
                  - "eu01-m",
                  + "eu01-1",
                  + "eu01-2",
                  + "eu01-3",
                ]
              ~ max_surge          = 1 -> 3
              ~ max_unavailable    = 1 -> 0
              ~ name               = "apidevnode" -> "nodepool"
                # (9 unchanged attributes hidden)
            },
          - {
              - availability_zones = [
                  - "eu01-1",
                  - "eu01-2",
                  - "eu01-3",
                ] -> null
              - cri                = "containerd" -> null
              - labels             = {} -> null
              - machine_type       = "m1.3" -> null
              - max_surge          = 1 -> null
              - max_unavailable    = 1 -> null
              - maximum            = 10 -> null
              - minimum            = 3 -> null
              - name               = "nodepool" -> null
              - os_name            = "flatcar" -> null
              - os_version         = "3815.2.1" -> null
              - volume_size        = 50 -> null
              - volume_type        = "storage_premium_perf1" -> null
            },
        ]
        # (4 unchanged attributes hidden)
    }
```

my nodepools were changed from

```hcl
  node_pools = [{
    name               = "apidevnode"
    machine_type       = "m1.3"
    max_surge          = "1"
    max_unavailable    = "1"
    maximum            = "10"
    minimum            = "3"
    volume_size        = "50"
    volume_type        = storage_premium_perf1
    os_version         = "3815.2.1"
    availability_zones = ["eu01-m"]
    }, {
    name            = "nodepool"
    machine_type    = "m1.3"
    max_surge       = "1"
    max_unavailable = "1"
    maximum         = "10"
    minimum         = "3"
    volume_size     = "50"
    volume_type     = storage_premium_perf1
    os_version      = "3815.2.1"
[...]

PR #376, opened 2024-05-24
Fix os_version and os_version_min conflicting fields

vicentepinto98:
Small bug found while investigating #372


joaopalet (approved):
Looks good to me, what was the issue with the `path.MatchRoot("node_pools").AtAnyListIndex().AtName` validation?

vicentepinto98:
> Looks good to me, what was the issue with the `path.MatchRoot("node_pools").AtAnyListIndex().AtName` validation?

I'm not sure why, but the framework was throwing the error that `os_version` was set in some `ListIndex`, which was not the case. So I decided to implement the validation in each individual nodepool, so that we have more control of what we are checking.

Also, it seemed more complex to try to implement that using [path expressions](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/path-expressions#building-expression-steps) which don't support something like validation at the same index out of the box.

PR #388, opened 2024-06-04
Extend scrape config schema

GokceGK:


stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -36,11 +36,19 @@ import (
...
+	DefaultSampleLimit                     = int64(5000)
+	DefaultSAML2EnableURLParameters        = true
+	DefaultOauth2InsecureSkipVerify        = false
+	DefaultHttpSdConfigRefreshInterval     = "60s"
+	DefaultMetricsRelabelConfigAction      = "replace"
+	DefaultMetricsRelabelConfigRegex       = ".*"
+	DefaultMetricsRelabelConfigReplacement = "$1"
+	DefaultMetricsRelabelConfigSeparator   = ";"
+	DefaultHonorLabels                     = false
+	DefaultHonorTimeStamps                 = false
```
DiogoFerrao, line 51: I don't think there should be defaults for these new attributes
I was discussing with João and our guidelines for having hard-coded default values in the Provider are:
 - Field is required in the API and doesn't have a default
 - Field has a "bad" default (the DefaultScheme case in this file)
 - Exception for when a default value makes sense or was requested by the service team/user

 In general, the new fields here are Optional/Computed so they shouldn't have defaults

stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -308,6 +390,239 @@ func (r *scrapeConfigResource) Schema(_ context.Context, _ resource.SchemaReques
...
+				Description: "",
+				Optional:    true,
+				Computed:    true,
+				Attributes: map[string]schema.Attribute{
+					"insecure_skip_verify": schema.BoolAttribute{
+						Description: "Disable validation of the server certificate.",
+						Optional:    true,
+					},
+				},
+			},
```
DiogoFerrao, line 625: The inner attribute should also be computed and it applies to all tls_config

stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -353,8 +665,44 @@ func (r *scrapeConfigResource) Create(ctx context.Context, req resource.CreateRe
...
+		diags = model.TlsConfig.As(ctx, &tlsConfigModel, basetypes.ObjectAsOptions{})
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
+
 	// Generate API request body from model
-	payload, err := toCreatePayload(ctx, &model, &saml2Model, &basicAuthModel, targetsModel)
+	payload, err := toCreatePayload(ctx, &model, &saml2Model, &basicAuthModel, targetsModel, httpSdConfigsModel, metricsRelabelConfigsModel, &oauth2Model, &tlsConfigModel)
```
DiogoFerrao, line 705: I would make them all pointers

stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -792,7 +1535,7 @@ func setDefaultsCreateScrapeConfig(sc *argus.CreateScrapeConfigPayload, model *M
 	}
 }
 
-func toUpdatePayload(ctx context.Context, model *Model, saml2Model *saml2Model, basicAuthModel *basicAuthModel, targetsModel []targetModel) (*argus.UpdateScrapeConfigPayload, error) {
+func toUpdatePayload(ctx context.Context, model *Model, saml2Model *saml2Model, basicAuthModel *basicAuthModel, targetsModel []targetModel, httpSdConfigsModel []httpSdConfigModel, metricsRelabelConfigsModel []metricsRelabelConfigModel, oauth2Model *oauth2Model, tlsConfigModel *tlsConfigModel) (*argus.UpdateScrapeConfigPayload, error) {
```
DiogoFerrao, line 1538: Same here

stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -713,11 +1358,16 @@ func toCreatePayload(ctx context.Context, model *Model, saml2Model *saml2Model,
...
-		Scheme:      conversion.StringValueToPointer(model.Scheme),
+		SampleLimit:     utils.Ptr(float64(model.SampleLimit.ValueInt64())),
+		Scheme:          conversion.StringValueToPointer(model.Scheme),
+		BearerToken:     conversion.StringValueToPointer(model.BearerToken),
+		HonorLabels:     conversion.BoolValueToPointer(model.HonorLabels),
+		HonorTimeStamps: conversion.BoolValueToPointer(model.HonorTimeStamps),
 	}
 	setDefaultsCreateScrapeConfig(&sc, model, saml2Model)
 
+	var diags diag.Diagnostics
```
DiogoFerrao, line 1369: I don't think this is necessary, was there some place you need this to be declared?

stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -758,6 +1408,99 @@ func toCreatePayload(ctx context.Context, model *Model, saml2Model *saml2Model,
...
+
+		mrcsi.Action = conversion.StringValueToPointer(metricsRelabelConfig.Action)
+		mrcsi.Modulus = utils.Ptr(metricsRelabelConfig.Modulus.ValueFloat64())
+		mrcsi.Regex = conversion.StringValueToPointer(metricsRelabelConfig.Regex)
+		mrcsi.Replacement = conversion.StringValueToPointer(metricsRelabelConfig.Replacement)
+		mrcsi.Separator = conversion.StringValueToPointer(metricsRelabelConfig.Separator)
+		mrcsi.TargetLabel = conversion.StringValueToPointer(metricsRelabelConfig.TargetLabel)
+
+		sourceLabels := []string{}
+		diags = metricsRelabelConfig.SourceLabels.ElementsAs(ctx, &sourceLabels, true)
```
DiogoFerrao, line 1430: You could just use `:=` here and its enough, I think `diags` is like the `err`

stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -758,6 +1408,99 @@ func toCreatePayload(ctx context.Context, model *Model, saml2Model *saml2Model,
 	}
 	sc.StaticConfigs = &t
 
+	if sc.TlsConfig == nil && !tlsConfigModel.InsecureSkipVerify.IsNull() && !tlsConfigModel.InsecureSkipVerify.IsNull() {
+		sc.TlsConfig = &argus.CreateScrapeConfigPayloadHttpSdConfigsInnerOauth2TlsConfig{
+			InsecureSkipVerify: conversion.BoolValueToPointer(tlsConfigModel.InsecureSkipVerify),
+		}
+	}
+
+	mrcs := make([]argus.CreateScrapeConfigPayloadMetricsRelabelConfigsInner, len(metricsRelabelConfigsModel))
```
DiogoFerrao, line 1417: We could probably improve these names `sc` is already not the best, but at least its the root level. Here we could call his `metricsRelabelConfigs`

stackit/internal/services/argus/scrapeconfig/resource.go
```diff
@@ -758,6 +1408,99 @@ func toCreatePayload(ctx context.Context, model *Model, saml2Model *saml2Model,
...
+
+	for i, metricsRelabelConfig := range metricsRelabelConfigsModel {
+		mrcsi := argus.CreateScrapeConfigPayloadMetricsRelabelConfigsInner{}
```
DiogoFerrao, line 1420: Same here, probably `metricsRelabelConfigsInner`

stackit/internal/services/argus/scrapeconfig/resource.go
GokceGK: As I was using in a few places, I thought it is better if I declare it in this place. But now I changed them to declaration whereever needed with :=

stackit/internal/services/argus/scrapeconfig/resource.go
GokceGK: Renamed

bahkauv70:
closed due to inactivity

PR #393, opened 2024-06-05
Allow users to set Argus metrics storage retention policy 

DiogoFerrao:


stackit/internal/services/argus/instance/resource.go
```diff
@@ -354,6 +362,49 @@ func (r *instanceResource) Create(ctx context.Context, req resource.CreateReques
...
+	// Get metrics retention policy after update
+	metricsResp, err := r.client.GetMetricsStorageRetentionExecute(ctx, *instanceId, projectId)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating instance", fmt.Sprintf("Getting metrics retention policy: %v", err))
+		return
+	}
+	// Map response body to schema
+	err = mapMetricsRetentionField(metricsResp, &model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating instance", fmt.Sprintf("Processing API payload: %v", err))
```
joaopalet, line 397:
```suggestion
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating instance", fmt.Sprintf("Processing API response for the metrics retention: %v", err))
```

stackit/internal/services/argus/instance/resource.go
```diff
@@ -497,6 +565,47 @@ func (r *instanceResource) Update(ctx context.Context, req resource.UpdateReques
 		return
 	}
 
+	// If any of the metrics retention days are set, set the metrics retention policy
+	if metricsRetentionDays != nil || metricsRetentionDays5mDownsampling != nil || metricsRetentionDays1hDownsampling != nil {
+		// Need to get the metrics retention policy because update endpoint is a PUT and we need to send all fields
+		metricsResp, err := r.client.GetMetricsStorageRetentionExecute(ctx, instanceId, projectId)
+		if err != nil {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating instance", fmt.Sprintf("Getting metrics retention policy: %v", err))
```
joaopalet, line 573:
```suggestion
			core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating instance", fmt.Sprintf("Getting metrics retention policy: %v", err))
```

stackit/internal/services/argus/instance/resource.go
```diff
@@ -497,6 +565,47 @@ func (r *instanceResource) Update(ctx context.Context, req resource.UpdateReques
...
+			return
+		}
+
+		metricsRetentionPayload, err := toUpdateMetricsStorageRetentionPayload(metricsRetentionDays, metricsRetentionDays5mDownsampling, metricsRetentionDays1hDownsampling, metricsResp)
+		if err != nil {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating instance", fmt.Sprintf("Building metrics retention policy payload: %v", err))
```
joaopalet, line 579: Same here, and for the ones below

stackit/internal/services/argus/instance/resource.go
joaopalet: Same for the update

stackit/internal/services/argus/instance/resource.go
```diff
@@ -639,6 +748,42 @@ func mapACLField(aclList *argus.ListACLResponse, model *Model) error {
...
+	if r == nil {
+		return fmt.Errorf("response input is nil")
+	}
+	if model == nil {
+		return fmt.Errorf("model input is nil")
+	}
+
+	if r.MetricsRetentionTimeRaw == nil || r.MetricsRetentionTime5m == nil || r.MetricsRetentionTime1h == nil {
+		return fmt.Errorf("metrics retention time is nil")
+	}
```
joaopalet, line 761: Do we need to error if these are not defined in the API? Maybe we could just not map in this case

stackit/internal/services/argus/instance/resource_test.go
```diff
@@ -46,13 +55,23 @@ func TestMapFields(t *testing.T) {
...
 				Acl: &[]string{
 					"1.1.1.1/32",
 				},
 				Message: utils.Ptr("message"),
 			},
+			&argus.GetMetricsStorageRetentionResponse{
+				MetricsRetentionTimeRaw: utils.Ptr("60d"),
+				MetricsRetentionTime1h:  utils.Ptr("30d"),
+				MetricsRetentionTime5m:  utils.Ptr("7d"),
+			},
```
joaopalet, line 74: These being diferent is to simulate an update, right?

joaopalet: AH but there is already another test case for the update, is that one different in any way?

stackit/internal/services/argus/argus_acc_test.go
```diff
@@ -127,6 +107,22 @@ func resourceConfig(acl *string, instanceName, planName, target, saml2EnableUrlP
...
+}
+
+func resourceConfig(acl, metricsRetentionDays, metricsRetentionDays1hDownsampling, metricsRetentionDays5mDownsampling *string, instanceName, planName, target, saml2EnableUrlParameters string) string {
+	return fmt.Sprintf("%s\n\n%s\n\n%s\n\n%s",
+		testutil.ArgusProviderConfig(),
+		instanceResourceConfig(acl, metricsRetentionDays, metricsRetentionDays1hDownsampling, metricsRetentionDays5mDownsampling, instanceName, planName),
+		scrapeConfigResourceConfig(target, saml2EnableUrlParameters),
+		credentialResourceConfig(),
+	)
+}
```
joaopalet, line 124: 👍

joaopalet: Nice organization

stackit/internal/services/argus/instance/datasource.go
```diff
@@ -233,23 +233,34 @@ func (d *instanceDataSource) Read(ctx context.Context, req datasource.ReadReques
...
 	}
-	err = mapACLField(aclList, &model)
+	err = mapACLField(aclListResp, &model)
 	if err != nil {
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating instance", fmt.Sprintf("Processing API response for the ACL: %v", err))
 		return
 	}
+	err = mapMetricsRetentionField(metricsRetentionResp, &model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating instance", fmt.Sprintf("Processing API response for the metrics retention: %v", err))
```
joaopalet, line 261:
```suggestion
	err = mapACLField(aclListResp, &model)
	if err != nil {
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading instance", fmt.Sprintf("Processing API response for the ACL: %v", err))
		return
	}
	err = mapMetricsRetentionField(metricsRetentionResp, &model)
	if err != nil {
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading instance", fmt.Sprintf("Processing API response for the metrics retention: %v", err))
```

stackit/internal/services/argus/instance/datasource.go
```diff
@@ -233,23 +233,34 @@ func (d *instanceDataSource) Read(ctx context.Context, req datasource.ReadReques
...
 	}
 
-	aclList, err := d.client.ListACL(ctx, instanceId, projectId).Execute()
+	aclListResp, err := d.client.ListACL(ctx, instanceId, projectId).Execute()
 	if err != nil {
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading instance", fmt.Sprintf("Calling API to list ACL data: %v", err))
 		return
 	}
 
+	metricsRetentionResp, err := d.client.GetMetricsStorageRetention(ctx, instanceId, projectId).Execute()
```
joaopalet, line 242: Why do we need this extra call in the datasource? Don't these fields also come in the `GetInstance` call?

stackit/internal/services/argus/instance/resource.go
DiogoFerrao: I don't see a valid case where this would be nil. We could definitely not error, but not mapping could lead to inconsistent applies, where Terraform complains that values are still unknown after apply.

We could do something like mapping to 0, but then I would prefer just erroring

stackit/internal/services/argus/instance/datasource.go
DiogoFerrao: Yes, and also the general map fields is already handling this, I think we can remove the mapping entirely

DiogoFerrao: Nice catch

stackit/internal/services/argus/instance/resource_test.go
DiogoFerrao: Yes, that's the goal. But its a nice catch because I created a specific update test and forgot to change this back

stackit/internal/services/argus/instance/resource.go
DiogoFerrao: Because the metrics storage retention is always set, starting with default values. So the get endpoint returning nil, has to be an error on the API side

stackit/internal/services/argus/instance/resource.go
joaopalet: This one is missing, right?

DiogoFerrao: You are right, I missed it. Thanks

PR #403, opened 2024-06-13
Onboard SQLServer Flex user resource

GokceGK:


stackit/internal/services/sqlserverflex/user/resource.go
```diff
@@ -0,0 +1,452 @@
...
+		model.ProjectId.ValueString(),
+		model.InstanceId.ValueString(),
+		userId,
+	}
+	model.Id = types.StringValue(
+		strings.Join(idParts, core.Separator),
+	)
+	model.UserId = types.StringValue(userId)
+	model.Username = types.StringPointerValue(user.Username)
+
```
DiogoFerrao, line 369: Is model.Database being mapped?

stackit/internal/services/sqlserverflex/user/resource.go
```diff
@@ -0,0 +1,452 @@
...
```
DiogoFerrao, line 420: Same here

stackit/internal/services/sqlserverflex/user/resource.go
GokceGK: in the response of the api, the field is named as default_database. Service team will fix it. Until then, we will leave this field out.

stackit/internal/services/sqlserverflex/user/resource.go
```diff
@@ -0,0 +1,453 @@
...
+			"username": schema.StringAttribute{
+				Required: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"roles": schema.SetAttribute{
+				ElementType: types.StringType,
+				Required:    true,
```
joaopalet, line 165: Is this field really required by the API? I think we can create users without setting it

GokceGK: Roles field is now Optional

joaopalet (approved):

PR #412, opened 2024-06-20
Support beta functionality

DiogoFerrao:


stackit/provider.go
```diff
@@ -103,6 +103,7 @@ type providerModel struct {
 	ResourceManagerCustomEndpoint types.String `tfsdk:"resourcemanager_custom_endpoint"`
 	TokenCustomEndpoint           types.String `tfsdk:"token_custom_endpoint"`
 	JWKSCustomEndpoint            types.String `tfsdk:"jwks_custom_endpoint"`
+	EnableBeta                    types.Bool   `tfsdk:"enable_beta"`
```
joaopalet, line 106: I think it would be a bit clearer if the name is more explicit
```suggestion
	EnableBetaResources                    types.Bool   `tfsdk:"enable_beta_resources"`
```

stackit/internal/features/beta.go
```diff
@@ -0,0 +1,39 @@
...
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/core"
+)
+
+// BetaFeatureFlagEnabled returns whether this provider is running with the BETA feature flag (enable_beta) enabled.
+func BetaFeatureFlagEnabled(data core.ProviderData) bool {
+	return data.EnableBeta
+}
```
joaopalet, line 15: This only used in the function below right? I would just inline it then

stackit/internal/features/beta.go
```diff
@@ -0,0 +1,39 @@
...
+
+// BetaResourcesEnabled returns whether this provider has BETA functionality enabled.
+//
+// In order of precedence, beta functionality can be managed by:
+//   - Environment Variable `STACKIT_TERRAFORM_BETA_RESOURCES` - `true` is enabled, unset or any other value is disabled.
+//   - Provider configuration feature flag `enable_beta` - `true` is enabled, `false` is disabled.
+func BetaResourcesEnabled(data core.ProviderData) bool {
+	value, ok := os.LookupEnv("STACKIT_TERRAFORM_BETA_RESOURCES")
```
joaopalet, line 23: And would maybe rename this to match the provider field
```suggestion
	value, ok := os.LookupEnv("STACKIT_TF_ENABLE_BETA_RESOURCES")
```

stackit/internal/features/beta.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	return strings.ToLower(value) == "true"
+}
+
+// CheckBetaResourcesEnabled is a helper function to log and add a warning or error if the BETA functionality is not enabled.
+//
+// Should be called in the Configure method of a BETA resource and you need to check for Errors in the diags after calling.
+func CheckBetaResourcesEnabled(ctx context.Context, data core.ProviderData, diags *diag.Diagnostics) {
+	if !BetaResourcesEnabled(data) {
+		core.LogAndAddErrorBeta(ctx, diags, "DNS record set")
```
joaopalet, line 35: This should be replaced by the actual name, can we get it here?

stackit/internal/features/beta.go
```diff
@@ -0,0 +1,54 @@
...
+	return data.EnableBetaResources
+}
+
+// BetaResourcesEnabled returns whether this provider has BETA functionality enabled.
+//
+// In order of precedence, beta functionality can be managed by:
+//   - Environment Variable `STACKIT_TF_ENABLE_BETA_RESOURCES` - `true` is enabled, `false` is disabled.
+//   - Provider configuration feature flag `enable_beta` - `true` is enabled, `false` is disabled.
+func BetaResourcesEnabled(ctx context.Context, data *core.ProviderData, diags *diag.Diagnostics) bool {
+	value, ok := os.LookupEnv("STACKIT_TF_ENABLE_BETA_RESOURCES")
```
joaopalet, line 27: *Nitpick:* I think using `set` is slighlty more clear
```suggestion
	value, set := os.LookupEnv("STACKIT_TF_ENABLE_BETA_RESOURCES")
```

stackit/internal/features/beta.go
```diff
@@ -0,0 +1,54 @@
...
+	if ok {
+		if strings.EqualFold(value, "true") {
+			return true
+		}
+		if strings.EqualFold(value, "false") {
+			return false
+		}
+		warnDetails := fmt.Sprintf("The value of the environment variable that enables BETA functionality must be either 'true' or 'false', got %q. \nDefaulting to the provider feature flag.", value)
+		core.LogAndAddWarning(ctx, diags, "Invalid value for STACKIT_TF_ENABLE_BETA_RESOURCES environment variable.", warnDetails)
```
joaopalet, line 36: 👌

joaopalet:
```suggestion
		warnDetails := fmt.Sprintf(`The value of the environment variable that enables BETA functionality must be either "true" or "false", got %q. \nDefaulting to the provider feature flag.`, value)
		core.LogAndAddWarning(ctx, diags, "Invalid value for STACKIT_TF_ENABLE_BETA_RESOURCES environment variable.", warnDetails)
```

stackit/internal/features/beta.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	// ProviderData should always be set, but we check just in case
+	if data == nil {
+		return false
+	}
+	return BetaFeatureFlagEnabled(data)
```
joaopalet, line 42:
```suggestion
	// ProviderData should always be set, but we check just in case
	if data == nil {
		return false
	}
	return data.EnableBetaResources
```

joaopalet: And then the function could be removed, right?

stackit/internal/features/beta.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// CheckBetaResourcesEnabled is a helper function to log and add a warning or error if the BETA functionality is not enabled.
+//
+// Should be called in the Configure method of a BETA resource and you need to check for Errors in the diags after calling.
```
joaopalet, line 47: This means that the caller needs to do something like 
```go
if diags.HasError() {
    return
}
```
after calling this function, right?

stackit/internal/features/beta_test.go
```diff
@@ -0,0 +1,285 @@
...
+			var err error
+			var lastEnv string
+			var envWasSet bool
+			if tt.envSet {
+				lastEnv, envWasSet = os.LookupEnv("STACKIT_TF_ENABLE_BETA_RESOURCES")
+				err = os.Setenv("STACKIT_TF_ENABLE_BETA_RESOURCES", tt.envValue)
+				if err != nil {
+					t.Fatalf("Error setting environment variable: %s", err)
+				}
+			}
```
joaopalet, line 202: :+1: for the complete test!

joaopalet: One suggestion: I think you can use [`t.Setenv`](https://pkg.go.dev/testing#B.Setenv), which sets the environment variable for the duration of the current test only

stackit/internal/features/beta_test.go
```diff
@@ -0,0 +1,285 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.description, func(t *testing.T) {
+			lastEnv, envWasSet := os.LookupEnv("STACKIT_TF_ENABLE_BETA_RESOURCES")
+			err := os.Unsetenv("STACKIT_TF_ENABLE_BETA_RESOURCES")
+			if err != nil {
+				t.Fatalf("Error setting environment variable: %s", err)
+			}
```
joaopalet, line 256: Same here

stackit/internal/features/beta.go
DiogoFerrao: Yes, exactly. I added that more explicitely in the comments

stackit/internal/features/beta_test.go
DiogoFerrao: Thanks, that actually simplifies the test a lot

joaopalet (approved):

PR #416, opened 2024-06-24
ref 624723: server backup schedules

aeter:
Adds a terraform provider for the STACKIT server-backup-schedules API. 

ref: 624723

---

Testing Done:
* Tested locally with a `.tf` file:
  ** tested terraform import, 
  ** tested datasource
  ** tested resource creation (with already enabled service / without already enabled service), resource update, resource deletion (tested disabling the service in case no more backups are present, tested that the service is not disabled in case any backups or backup schedules are present).
* Ran unit tests (`make test`)
* Ran linters/formatters (`make fmt`, `make lint`)


stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+// scheduleResource is the resource implementation.
+type scheduleResource struct {
+	client *serverbackup.APIClient
+}
+
+// Metadata returns the resource type name.
+func (r *scheduleResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_serverbackup_schedule"
```
joaopalet, line 62: I would prefer `stackit_server_backup_schedule`, what do you think?

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+		)
+	}
+
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error configuring API client", fmt.Sprintf("Configuring client: %v. This is an error related to the provider configuration, not to the resource configuration", err))
+		return
+	}
+
+	r.client = apiClient
+	tflog.Info(ctx, "Serverbackup schedule client configured")
```
joaopalet, line 98: Same for these logs, could be `Server backup schedule ...` instead

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
+				},
+			},
+			"server_id": schema.StringAttribute{
+				Description: "Server ID for the backup schedule",
```
joaopalet, line 147:
```suggestion
				Description: "Server ID for the backup schedule.",
```

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(1, 255),
+				},
+			},
+			"backup_schedule_id": schema.Int64Attribute{
+				Description: "Backup schedule ID",
```
joaopalet, line 125:
```suggestion
				Description: "Backup schedule ID.",
```

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+				Validators: []validator.String{
+					validate.Rrule(),
+					validate.NoSeparator(),
+				},
+			},
+			"enabled": schema.BoolAttribute{
+				Description: "Is the backup schedule enabled or disabled.",
+				Required:    true,
+			},
+			"backup_properties": schema.ObjectAttribute{
```
joaopalet, line 174: We are using `SingleNestedAttribute`s for these kind of objects as the `ObjectAttribute`s type as been discouraged (see [here](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/object))

joaopalet: You can check the implementation in some of the existing resources, e.g. https://github.com/stackitcloud/terraform-provider-stackit/blob/610b65f1ccd85427a464e030a5a956d8d895016d/stackit/internal/services/mariadb/instance/resource.go#L183

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+				Description: "Backup schedule details for the backups",
```
joaopalet, line 175:
```suggestion
				Description: "Backup schedule details for the backups.",
```

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+	projectId := model.ProjectId.ValueString()
+	serverId := model.ServerId.ValueString()
+	ctx = tflog.SetField(ctx, "project_id", projectId)
+	ctx = tflog.SetField(ctx, "server_id", serverId)
+
+	// Enable backups if not already enabled
+	err := enableBackupsService(ctx, &model, r.client)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating schedule", fmt.Sprintf("Enabling serverbackup project before creation: %v", err))
```
joaopalet, line 205:
```suggestion
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating server backup schedule", fmt.Sprintf("Enabling serverbackup project before creation: %v", err))
```

joaopalet: This is applicable everywhere in the file, to avoid confusions with other possible backups existing in the provider

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+	model.Name = types.StringPointerValue(schedule.Name)
+	model.Rrule = types.StringPointerValue(schedule.Rrule)
+	model.Enabled = types.BoolPointerValue(schedule.Enabled)
+	if schedule.BackupProperties == nil {
+		model.BackupProperties = nil
+		return nil
+	}
+	ids, diags := types.ListValueFrom(ctx, types.StringType, schedule.BackupProperties.VolumeIds)
+	if diags.HasError() {
+		return fmt.Errorf("failed to map hosts: %w", core.DiagsToError(diags))
```
joaopalet, line 406: In general we try to keep error messages short and avoid repeating patterns such as `failed to` or `error` in, because this is already clear from the start.

E.g. we use `Error creating schedule: process payload: map hosts: ...` instead of `Error creating schedule: failed to process payload: failed to map hosts: failed to ...`

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+
+	tflog.Debug(ctx, "Enabling server backup service")
+	err := client.EnableService(ctx, projectId, serverId).EnableServicePayload(enableServicePayload).Execute()
+	if err != nil {
+		if strings.Contains(err.Error(), "Tried to activate already active service") {
+			tflog.Debug(ctx, "Service for server backup already enabled")
+			return nil
+		}
+		return fmt.Errorf("Failed to enable server backup service: %w", err)
+	}
```
joaopalet, line 430: So enabling the service fails if the service is already enabled? I am asking because usually other APIs are idempotent in this case

joaopalet: But if this one fails, couldn't we get the status of the service first, and only call the enable endpoint if the API is not yet enabled?

This would avoid having to rely on the error message, which can be more error prone

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+func disableBackupsService(ctx context.Context, model *scheduleResourceModel, client *serverbackup.APIClient) error {
+	tflog.Debug(ctx, "Disabling server backup service (in case there are no backups and no backup schedules)")
+
+	projectId := model.ProjectId.ValueString()
+	serverId := model.ServerId.ValueString()
+
+	tflog.Debug(ctx, "Checking for existing backup schedules")
+	schedules, err := client.ListBackupSchedules(ctx, projectId, serverId).Execute()
+	if err != nil {
+		return fmt.Errorf("Failed to disable server backup service: %w", err)
```
joaopalet, line 445: E.g. this error message could be simply `list existing backup schedules`, because the context of the caller is already indicating that we are trying to disable the service

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+			return nil
+		}
+		return fmt.Errorf("Failed to enable server backup service: %w", err)
+	}
+	tflog.Info(ctx, "Enabled server backup service")
+	return nil
+}
+
+// Disables only if no backup schedules are present and no backups are present
+func disableBackupsService(ctx context.Context, model *scheduleResourceModel, client *serverbackup.APIClient) error {
```
joaopalet, line 436: We usually do not disable the service in the provider, but it's also fine if you feel like this is needed in this case

stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -0,0 +1,539 @@
...
+	if model.BackupProperties != nil {
+		ids := []string{}
+		if !(model.BackupProperties.VolumeIds.IsNull() || model.BackupProperties.VolumeIds.IsUnknown()) {
+			for _, elem := range model.BackupProperties.VolumeIds.Elements() {
+				val, err := conversion.ToString(ctx, elem)
+				if err != nil {
+					return nil, fmt.Errorf("Error by converting volume id: %w", err)
+				}
+				ids = append(ids, val)
+			}
```
joaopalet, line 485: We have a util `ListValuetoStringSlice` that might be useful to simplify this logic

stackit/internal/services/serverbackup/schedule/datasource.go
```diff
@@ -0,0 +1,240 @@
...
+					validate.NoSeparator(),
+				},
+			},
+			"items": schema.ListNestedAttribute{
+				Computed: true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"backup_schedule_id": schema.Int64Attribute{
+							Computed: true,
+						},
```
joaopalet, line 110: We have never had a case where the datasource is so different than the resource. So I'm not sure I understand the use case, is there a reason why this datasource does not take the `backup_schedule_id` as a required field and reads the data of a single backup schedule?

stackit/internal/services/serverbackup/schedule/resource_test.go
joaopalet, line 1: The unit tests for `toCreatePayload` and `toUpdatePayload` are missing

stackit/internal/validate/validate.go
```diff
@@ -225,3 +226,30 @@ func CIDR() *Validator {
 		},
 	}
 }
+
+func Rrule() *Validator {
```
joaopalet, line 230: Could you also add unit tests to this validator? You can find examples for all others in `validate_test.go`

joaopalet (commented):
The PR is missing the end-to-end tests for the service, these are in files in the form of `<service>_acc_test.go`, for each of the services present in the SDK.

After you are done with the changes, you can also run `make generate-docs` to generate the docs that will be shown in the Terraform registry page (based in the field descriptions, examples, etc)

stackit/internal/services/serverbackup/schedule/resource.go
aeter: ok

stackit/internal/services/serverbackup/schedule/resource.go
aeter: thank you

stackit/internal/services/serverbackup/schedule/resource.go
aeter: Yes, enabling the service fails if already enabled. It seems this should be fixed in the service itself. This code will continue to work after it's fixed.
I think there's no API endpoint to query the status of the service, so for now it seems this error message is the only way to do it. Sorry. I don't know if we can request changing the service.

stackit/internal/services/serverbackup/schedule/resource.go
aeter: It seemed the right way to do it, for the users.

stackit/internal/services/serverbackup/schedule/datasource.go
aeter: Yes. So, I will do as you suggest (`backup_schedule_id` as a required field, populate only 1 schedule) to conform with the rest of the project.

---

In the future, probably the project will need to think of how to filter a list of backup schedules. Like, with a datasource - adding a bunch of `ids` into terraform. There's probably no need to d this as part of this PR. 

For completeness, to explain what I mean: it seems at other terraform providers over the internet, this happens in the following ways (maybe more, this is what I found):

1. a `filter` attribute with an optional `id` attribute (like: [aws_instance](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/instance), etc.)  or
2. adding a plural version of the datasource. Like, having `stackit_server_backup_schedule` and a `stackit_server_backup_schedules` datasource. 
Examples:
[oci_core_subnets](https://registry.terraform.io/providers/oracle/oci/latest/docs/data-sources/core_subnets) <-> [oci_core_subnet](https://registry.terraform.io/providers/oracle/oci/latest/docs/data-sources/core_subnet) 
[aws_vpcs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpcs) <-> [aws_vpc](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpc)

With the required `id` it seems like you have chosen the second way - it's ok with me. If you haven't decided on it yet, this can be added at a later PR maybe, etc.

Thanks for noticing it in the comment, too.

joaopalet: > ok - I will do as you suggest (backup_schedule_id as a required field, populate only 1 schedule) to conform with the rest of the project.

Alright, thank you

> a filter attribute with an optional id attribute (like: [aws_instance](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/instance), etc.) or

I didn't know this approach, but it also says that `If anything other than a single match is returned by the search, Terraform will fail. Ensure that your search is specific enough to return a single Instance ID only.`

> adding a plural version of the datasource. Like, having stackit_server_backup_schedule and a stackit_server_backup_schedules datasource.

So I would prefer this one if we want to allow getting data of multiple backups 👍

joaopalet: You have already implemented it, right? So you could even leave it and change the name to be plural, and implement the singular one. What do you think @aeter

stackit/internal/services/server/backup/schedule/resource.go
```diff
@@ -0,0 +1,551 @@
...
+		)
+	}
+
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error configuring API client", fmt.Sprintf("Configuring client: %v. This is an error related to the provider configuration, not to the resource configuration", err))
+		return
+	}
+
+	r.client = apiClient
+	tflog.Info(ctx, "Server backup schedule client configured.")
```
joaopalet, line 106:
```suggestion
	tflog.Info(ctx, "Server backup client configured.")
```

stackit/internal/services/server/backup/schedule/resource.go
```diff
@@ -0,0 +1,551 @@
...
+	ctx = tflog.SetField(ctx, "project_id", projectId)
+	ctx = tflog.SetField(ctx, "server_id", serverId)
+	ctx = tflog.SetField(ctx, "backup_schedule_id", backupScheduleId)
+
+	err := r.client.DeleteBackupSchedule(ctx, projectId, serverId, strconv.FormatInt(backupScheduleId, 10)).Execute()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error deleting backup schedule", fmt.Sprintf("Calling API: %v", err))
+		return
+	}
+	tflog.Info(ctx, "Server backup schedule deleted ")
```
joaopalet, line 356:
```suggestion
	tflog.Info(ctx, "Server backup schedule deleted.")
```

stackit/internal/services/server/backup/schedule/resource.go
```diff
@@ -0,0 +1,551 @@
...
+	serverId := model.ServerId.ValueString()
+	backupScheduleId := model.BackupScheduleId.ValueInt64()
+	ctx = tflog.SetField(ctx, "project_id", projectId)
+	ctx = tflog.SetField(ctx, "server_id", serverId)
+	ctx = tflog.SetField(ctx, "backup_schedule_id", backupScheduleId)
+
+	// Update schedule
+	payload, err := toUpdatePayload(&model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating schedule", fmt.Sprintf("Creating API payload: %v", err))
```
joaopalet, line 312:
```suggestion
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating server backup schedule", fmt.Sprintf("Creating API payload: %v", err))
```

stackit/internal/services/server/backup/schedule/resource.go
```diff
@@ -0,0 +1,551 @@
...
+		return
+	}
+
+	scheduleResp, err := r.client.UpdateBackupSchedule(ctx, projectId, serverId, strconv.FormatInt(backupScheduleId, 10)).UpdateBackupSchedulePayload(*payload).Execute()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating schedule", fmt.Sprintf("Calling API: %v", err))
```
joaopalet, line 318:
```suggestion
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating server backup schedule", fmt.Sprintf("Calling API: %v", err))
```

stackit/internal/services/server/backup/schedule/resource.go
```diff
@@ -0,0 +1,551 @@
...
+		return
+	}
+
+	// Map response body to schema
+	err = mapFields(ctx, scheduleResp, &model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating schedule", fmt.Sprintf("Processing API payload: %v", err))
```
joaopalet, line 325:
```suggestion
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating server backup schedule", fmt.Sprintf("Processing API payload: %v", err))
```

examples/data-sources/stackit_server_backup_schedule/schedule-data-source.tf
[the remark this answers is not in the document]
joaopalet: In this case it would be `docs/data-sources/stackit_server_backup_schedule/data-source.tf` and `docs/data-sources/stackit_server_backup_schedules/data-source.tf`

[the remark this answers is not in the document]
joaopalet: Sorry, I meant "**examples**/data-sources/<data_source_name>/data-source.tf"

joaopalet (approved):

PR #422, opened 2024-06-25
Fix formatting of the beta configuration guide

DiogoFerrao:


templates/guides/opting_into_beta_resources.md.tmpl
```diff
@@ -24,20 +24,18 @@ provider "stackit" {
...
 Set the `STACKIT_TF_ENABLE_BETA_RESOURCES` environment variable to `"true"` or `"false"`. Other values will be ignored and will produce a warning.
 
 ```sh
 export STACKIT_TF_ENABLE_BETA_RESOURCES=true
 ```
 
-> **Note**: **The environment variable takes precedence over the provider configuration option.**
->
-> This means that if the `STACKIT_TF_ENABLE_BETA_RESOURCES` environment variable is set to a valid value (`"true"` or `"false"`), it will override the `enable_beta_resources` option specified in the provider configuration.
+-> The environment variable takes precedence over the provider configuration option.
```
joaopalet, line 35: 👍

templates/guides/opting_into_beta_resources.md.tmpl
```diff
@@ -24,20 +24,18 @@ provider "stackit" {
...
 
-### Listing Beta Resources
+This means that if the `STACKIT_TF_ENABLE_BETA_RESOURCES` environment variable is set to a valid value (`"true"` or `"false"`), it will override the `enable_beta_resources` option specified in the provider configuration.
```
joaopalet, line 37: Doesn't this line need to be right after the previous one so also be included in the highlighted box?

joaopalet (commented):
Could you also fix the header for the other guide?

joaopalet (approved):

PR #423, opened 2024-06-25
Remove CRUD method logic from PostgreSQL (first step of removal)

DiogoFerrao:


stackit/internal/services/postgresql/credential/resource.go
```diff
@@ -103,8 +98,8 @@ func (r *credentialResource) Schema(_ context.Context, _ resource.SchemaRequest,
 		"main": "PostgreSQL credential resource schema. Must have a `region` specified in the provider configuration.",
 		"deprecation_message": strings.Join(
 			[]string{
-				"The STACKIT PostgreSQL service will reach its end of support on June 30th 2024.",
-				"Resources of this type will stop working after that.",
+				"The STACKIT PostgreSQL service has reached its end of support on June 30th 2024.",
+				"Resources of this type have stopped working since then.",
 				"Use stackit_postgresflex_user instead.",
 				"For more details, check https://docs.stackit.cloud/stackit/en/bring-your-data-to-stackit-postgresql-flex-138347648.html",
```
joaopalet, line 104: We could also add this line to the operation error messages

stackit/internal/services/postgresql/credential/resource.go
joaopalet, line 1: And we can also remove the acc test, right?

DiogoFerrao: Yep, missed that

joaopalet (approved):

PR #438, opened 2024-06-28
Extend LogMe instance parameters

joaopalet:


stackit/internal/services/logme/instance/resource.go
```diff
@@ -124,6 +170,17 @@ func (r *instanceResource) Schema(_ context.Context, _ resource.SchemaRequest, r
...
+	parametersDescriptions := map[string]string{
+		"sgw_acl":                "Comma separated list of IP networks in CIDR notation which are allowed to access this instance.",
+		"enable_monitoring":      "Enable monitoring.",
+		"graphite":               "If set, monitoring with Graphite will be enabled. Expects the host and port where the Graphite metrics should be sent to (host:port).",
+		"max_disk_threshold":     "The maximum disk threshold in MB. If the disk usage exceeds this threshold, the instance will be stopped.",
+		"metrics_frequency":      "The frequency in seconds at which metrics are emitted.",
+		"metrics_prefix":         "The prefix for the metrics. Could be useful when using Graphite monitoring to prefix the metrics with a certain value, like an API key",
+		"monitoring_instance_id": "The monitoring instance ID.",
+		"syslog":                 "List of syslog servers to send logs to.",
+	}
```
DiogoFerrao, line 182: There are missing descriptions here right? Things like `java_heapspace`, etc

stackit/internal/services/logme/instance/resource.go
```diff
@@ -242,6 +413,8 @@ func (r *instanceResource) Create(ctx context.Context, req resource.CreateReques
 		if resp.Diagnostics.HasError() {
 			return
 		}
+	} else {
+		parameters = nil
```
DiogoFerrao, line 417: Why is this needed? And if so, couldn't this be avoided with a `var parameters *parametersModel` declaration above? Then create a parametersModel struct inside the if

stackit/internal/services/logme/instance/resource.go
```diff
@@ -52,12 +54,56 @@ type Model struct {
...
+	IsmJitter              types.Float64 `tfsdk:"ism_jitter"`
+	IsmJobInterval         types.Int64   `tfsdk:"ism_job_interval"`
+	JavaHeapspace          types.Int64   `tfsdk:"java_heapspace"`
+	JavaMaxmetaspace       types.Int64   `tfsdk:"java_maxmetaspace"`
+	MaxDiskThreshold       types.Int64   `tfsdk:"max_disk_threshold"`
+	MetricsFrequency       types.Int64   `tfsdk:"metrics_frequency"`
+	MetricsPrefix          types.String  `tfsdk:"metrics_prefix"`
+	MonitoringInstanceId   types.String  `tfsdk:"monitoring_instance_id"`
+	OpensearchTlsCiphers   types.List    `tfsdk:"opensearch_tls_ciphers"`
+	OpensearchTlsProtocols types.List    `tfsdk:"opensearch_tls_protocols"`
```
DiogoFerrao, line 77: What was the conclusion regarding the inclusion of the `Groks`?

stackit/internal/services/logme/instance/resource.go
DiogoFerrao: I see that this being nil is useful to check in the `toInstanceParams`

stackit/internal/services/logme/instance/resource_test.go
```diff
@@ -199,7 +295,18 @@ func TestToCreatePayload(t *testing.T) {
...
+			var parameters = &parametersModel{}
+			if tt.input != nil {
+				if !(tt.input.Parameters.IsNull() || tt.input.Parameters.IsUnknown()) {
+					diags := tt.input.Parameters.As(context.Background(), parameters, basetypes.ObjectAsOptions{})
+					if diags.HasError() {
+						t.Fatalf("Error converting parameters: %v", diags.Errors())
+					}
+				} else {
+					parameters = nil
+				}
```
DiogoFerrao, line 307: I believe that the same comment from the resource applies here

stackit/internal/services/logme/logme_acc_test.go
```diff
@@ -62,7 +91,13 @@ func TestAccLogMeResource(t *testing.T) {
...
 			// Creation
 			{
-				Config: resourceConfig(instanceResource["sgw_acl-1"]),
+				Config: resourceConfig(
+					map[string]string{
+						"sgw_acl":            instanceResource["sgw_acl-1"],
+						"fluentd_tcp":        instanceResource["fluent_tcp"],
+						"max_disk_threshold": instanceResource["max_disk_threshold"],
+						"enable_monitoring":  instanceResource["enable_monitoring"],
+					}),
```
DiogoFerrao, line 100: Wouldn't it be better to make sure that all parameters work? Although I believe there are some parameters that we don't know valid values for, I think we can exclude those

stackit/internal/services/logme/instance/resource.go
joaopalet: Yes, I was missing some of the ones that are set in the API docs. Added them

joaopalet: Some are still not present but they're not documented anywhere, we can add them when the team updates the docs

stackit/internal/services/logme/instance/resource.go
joaopalet: Good suggestion, applied it

stackit/internal/services/logme/logme_acc_test.go
joaopalet: I'm extend it with parameters as I find out some values that work, but creation and update of the instance is taking almost 1 hour

stackit/internal/services/logme/instance/resource.go
joaopalet: I won't add support for that field for now, we can consider it if we get a request or once the team improves their "GetInstance" endpoint response to actually return the correct types instead of the generic object

stackit/internal/services/logme/logme_acc_test.go
joaopalet: Added a `syslog` list field for type completion

joaopalet: And the `ism_jitter` float parameter

DiogoFerrao (approved):

PR #439, opened 2024-06-28
Onboard Argus alert config receivers

DiogoFerrao:


stackit/internal/services/argus/instance/resource.go
```diff
@@ -64,6 +66,100 @@ type Model struct {
...
+
+type routeModel struct {
+	Receiver types.String `tfsdk:"receiver"`
+}
+
+var routeTypes = map[string]attr.Type{
+	"receiver": types.StringType,
+}
+
+// Struct corresponding to DataSourceModel.AlertConfig.receivers
```
joaopalet, line 107:
```suggestion
// Struct corresponding to ResourceModel.AlertConfig.receivers
```

stackit/internal/services/argus/instance/resource.go
```diff
@@ -405,6 +623,40 @@ func (r *instanceResource) Create(ctx context.Context, req resource.CreateReques
 		return
 	}
 
+	// Alert Config
+
+	if model.AlertConfig.IsUnknown() || model.AlertConfig.IsNull() {
+		alertConfig = mockAlertConfig
+	}
```
joaopalet, line 630: We should have some explaination here on why we need to do this

joaopalet: Actually, do we need to set the mock on creation? Couldn't it be done just on the update, when the user is trying to delete their alert configs?

stackit/internal/services/argus/instance/resource.go
```diff
@@ -784,6 +1092,174 @@ func mapMetricsRetentionField(r *argus.GetMetricsStorageRetentionResponse, model
...
+	receiversList := []attr.Value{}
+
+	for i := range respReceivers {
+		receiver := respReceivers[i]
+
+		// emailConfigs
+		emailConfigList := []attr.Value{}
+		if receiver.EmailConfigs != nil {
+			for _, emailConfig := range *receiver.EmailConfigs {
+				emailConfigList = append(emailConfigList, types.ObjectValueMust(emailConfigsTypes, map[string]attr.Value{
```
joaopalet, line 1116: We shouldn't use `...ValueMust` function in the provider code, as they panic in case of an error in the values instead of returning an error. The function description also says: "This creation function is only recommended to create List values which will not potentially affect practitioners, such as testing, or exhaustively tested provider logic."

joaopalet: I think we only use them in tests so far

stackit/internal/services/argus/instance/resource.go
```diff
@@ -784,6 +1092,174 @@ func mapMetricsRetentionField(r *argus.GetMetricsStorageRetentionResponse, model
...
+
+	modelMockAlertConfig, diags := types.ObjectValueFrom(ctx, alertConfigTypes, mockAlertConfig)
+	if diags.HasError() {
+		return fmt.Errorf("mapping mock alert config: %w", core.DiagsToError(diags))
+	}
+
+	if alertConfig.Equal(modelMockAlertConfig) {
+		model.AlertConfig = types.ObjectNull(alertConfigTypes)
+		return nil
+	}
```
joaopalet, line 1173: This should also have a comment explaining the why behind it

stackit/internal/services/argus/instance/resource.go
```diff
@@ -469,6 +727,13 @@ func (r *instanceResource) Read(ctx context.Context, req resource.ReadRequest, r
 		return
 	}
 
+	// Map response body to schema
+	err = mapAlertConfigField(ctx, alertConfigResp, &model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating instance", fmt.Sprintf("Processing API response for the alert config: %v", err))
+		return
+	}
```
joaopalet, line 735: Since the field is only Optional (and not Computed), do we need to map the alert configs to the state in the Read? Isn't this lead to the plan always showing changes?

stackit/internal/services/argus/instance/resource.go
```diff
@@ -784,6 +1092,174 @@ func mapMetricsRetentionField(r *argus.GetMetricsStorageRetentionResponse, model
 	return nil
 }
 
+func mapUpdateAlertConfigField(ctx context.Context, resp *argus.UpdateAlertConfigsResponse, model *Model) error {
```
joaopalet, line 1095: This function and `mapAlertConfigField` only differ in the signature (because of the `resp` type) and in the first if block, right?

joaopalet: If yes can't we encapsulate everything else after
```go
// Receivers
	respReceivers := *resp.Data.Receivers
```
in a sub function?

stackit/internal/services/argus/instance/resource.go
```diff
@@ -64,6 +66,100 @@ type Model struct {
...
+			"name":             types.StringValue("change-me"),
+			"email_configs":    types.ListValueMust(types.ObjectType{AttrTypes: emailConfigsTypes}, []attr.Value{}),
+			"opsgenie_configs": types.ListValueMust(types.ObjectType{AttrTypes: opsgenieConfigsTypes}, []attr.Value{}),
+			"webhooks_configs": types.ListValueMust(types.ObjectType{AttrTypes: webHooksConfigsTypes}, []attr.Value{}),
+		}),
+	}),
+	Route: types.ObjectValueMust(routeTypes, map[string]attr.Value{
+		"receiver": types.StringValue("change-me"),
+	}),
+}
```
joaopalet, line 84: It looks a little more natural to define the mock in the actual type from the SDK, instead of in terms of the model. Or does this makes it easier in any way?

joaopalet: Also because if you remove the `...ValueMust` the function will possibly return an error and will make it a bit weirder

stackit/internal/services/argus/argus_acc_test.go
```diff
@@ -45,6 +46,28 @@ var credentialResource = map[string]string{
 	"project_id": testutil.ProjectId,
 }
 
+const alertConfigResource = `{
```
joaopalet, line 49:
```suggestion
const alertConfigConfig = `{
```

joaopalet: Oh maybe that's not the best name also

stackit/internal/services/argus/argus_acc_test.go
joaopalet, line 1: This file seems incomplete, or are you planning to update it in the end of the thirs story only?

stackit/internal/services/argus/instance/resource.go
DiogoFerrao: Agreed, I added a comment

stackit/internal/services/argus/instance/resource.go
DiogoFerrao: The advantage of setting our mock on creation instead of relying on the API's mock is that in the first update command after creating the instance only has the changes that the user intends. If we don't set the our mock on creation the user would get the Alert config changes in the plan (trying to change to our mock Alert Config), which may be confusing

DiogoFerrao: I think the best middle ground is to have our mock match their current default Alert Config. This way, it will work even if they change their defaults but if they don't, then there will never be "hidden" changes

stackit/internal/services/argus/instance/resource.go
DiogoFerrao: That was an oversight on my side, you are 100% right. Fixed now

stackit/internal/services/argus/instance/resource.go
DiogoFerrao: The mock configuration is only used for Terraform models, which means that it is not very practical to set it to an SDK payload, since I would always have to convert it. 

Also, setting it to the locally defined `alertConfigModel` allows it to be directly called in the `Create` method and the conversion is less complex in the `alertConfigModel` -> `basetypes.ObjectValue` direction than in reverse. But for this I can see a case for either type, let me know if you have a strong opinion

stackit/internal/services/argus/argus_acc_test.go
DiogoFerrao: Yes, it will be easier to test when the entire `Alert Config` functionality is implemented, here I just make sure that the mapping for the lists is working

stackit/internal/services/argus/instance/resource.go
DiogoFerrao: This still needs to be mapped, there are some fields that are set by default by the API and are not required, for instance the `globalOptions` or some fields inside the `emailConfigs` etc, so we would need to update the state accordingly

The part about leading to the plan always showing changes I don't completely understand. What does happen is that if only one of the optional fields is set, the plan will show `(state known after apply)` in the others, but the plan after an apply is empty since the fields already match what was read, right?
Even in the case of the mock config, since we just set it to `Null` it always matches as well

stackit/internal/services/argus/instance/resource.go
joaopalet: I see, good idea

stackit/internal/services/argus/instance/resource.go
joaopalet: Right, because we map to null objects if the value is the mocked config

joaopalet (approved):

PR #444, opened 2024-07-01
Fix SKE cluster maintenance updated outside TF causing error

joaopalet:
- The API response always contains a mock date part in front of the time, so in case these are the same we just keep the current value in state (this part hasn't changed)
- What I changed was that now if the API response is different (not only in format but also in the actual time), this takes precedent over the state value as usual (instead of erroring)

### DIFF
--- a/stackit/internal/services/ske/cluster/resource.go
+++ b/stackit/internal/services/ske/cluster/resource.go
@@ -1512,6 +1512,7 @@ func getMaintenanceTimes(ctx context.Context, cl *ske.Cluster, m *Model) (startT
 		return "", "", fmt.Errorf("parsing end time '%s' from API response as RFC3339 datetime: %w", *cl.Maintenance.TimeWindow.End, err)
 	}
 
+	// return startTimeAPI.Format("15:04:05Z07:00"), endTimeAPI.Format("15:04:05Z07:00"), nil
 	if m.Maintenance.IsNull() || m.Maintenance.IsUnknown() {
 		return startTimeAPI.Format("15:04:05Z07:00"), endTimeAPI.Format("15:04:05Z07:00"), nil
 	}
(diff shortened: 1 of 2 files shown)


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1529,10 +1529,12 @@ func getMaintenanceTimes(ctx context.Context, cl *ske.Cluster, m *Model) (startT
...
 			return "", "", fmt.Errorf("parsing start time '%s' from TF config as RFC time: %w", maintenance.Start.ValueString(), err)
 		}
-		if startTimeAPI.Format("15:04:05Z07:00") != startTimeTF.Format("15:04:05Z07:00") {
-			return "", "", fmt.Errorf("start time '%v' from API response doesn't match start time '%v' from TF config", *cl.Maintenance.TimeWindow.Start, maintenance.Start.ValueString())
+		// If the start time from the API just differs in time format, we keep the current model value
+		if startTimeAPI.Format("15:04:05Z07:00") == startTimeTF.Format("15:04:05Z07:00") {
+			startTime = maintenance.Start.ValueString()
+		} else {
+			startTime = startTimeAPI.Format("15:04:05Z07:00")
 		}
```
DiogoFerrao, line 1537: This makes sense, but then don't we return `startTimeAPI.Format("15:04:05Z07:00")` every time?

If maintenance.Start is Null or Unknown we return `startTimeAPI.Format("15:04:05Z07:00")`
Otherwise, if the API and Model times match, we return `maintenance.Start.ValueString()`, which should be the same as `startTimeAPI.Format("15:04:05Z07:00")`
Otherwise, if they don't match, we return `startTimeAPI.Format("15:04:05Z07:00")`

I feel like I'm missing something, but isn't all this just `startTime = startTimeAPI.Format("15:04:05Z07:00")`? (and the same for endTime)

joaopalet: As discussed offline, in the TF model we allow two time formats, and when the start times from the API and the TF are equal in content (string compare is the same), we keep the current TF model value to avoid "inconsistent result after apply"

DiogoFerrao (approved):

PR #445, opened 2024-07-02
Add description for the parameters field of DSA instances

joaopalet:


docs/data-sources/postgresql_credential.md
```diff
@@ -4,7 +4,7 @@ page_title: "stackit_postgresql_credential Data Source - stackit"
 subcategory: ""
 description: |-
   PostgreSQL credential data source schema. Must have a region specified in the provider configuration.
-  !> The STACKIT PostgreSQL service has reached its end of support on June 30th 2024. Resources of this type have stopped working since then. Use stackitpostgresflexuser instead. For more details, check https://docs.stackit.cloud/stackit/en/bring-your-data-to-stackit-postgresql-flex-138347648.html
+  !> The STACKIT PostgreSQL service has reached its end of support on June 30th 2024. Resources of this type have stopped working since then. Use stackit_postgresflex_user instead. For more details, check
```
joaopalet, line 7: Not sure why these are here but they are always there when I generate the docs, and only update the files metadata so the content is unchanged

stackit/internal/services/redis/instance/resource.go
```diff
@@ -165,6 +165,7 @@ func (r *instanceResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 		"version":     "The service version.",
 		"plan_name":   "The selected plan name.",
 		"plan_id":     "The selected plan ID.",
+		"parameters":  "Configuration parameters. Since the API only allows updating these through a PATCH request, removing a previsouly configured field from your Terraform configuration won't replace its value in the API. To update a previously configured field, please explicitly set a new value for it.",
```
DiogoFerrao, line 168:
```suggestion
		"parameters":  "Configuration parameters. Since the API only allows updating these through a PATCH request, removing a previously configured field from your Terraform configuration won't replace its value in the API. To update a previously configured field, please explicitly set a new value for it.",
```

DiogoFerrao: I'm not sure we should include the `PATCH request` part, I think it will confuse a lot of users

DiogoFerrao:
```suggestion
		"parameters":  "Configuration parameters. Please note that removing a previously configured field from your Terraform configuration won't replace its value in the API. To update a previously configured field, explicitly set a new value for it.",
```

docs/data-sources/postgresql_credential.md
DiogoFerrao: This is really weird, not sure what is happening. But anyway, changing the metadata description shouldn't be a problem

stackit/internal/services/redis/instance/resource.go
DiogoFerrao: I don't think we should include the `PATCH endpoint` part, it will confuse a lot of users

DiogoFerrao (approved):

PR #447, opened 2024-07-04
Onboard Alert Config route

DiogoFerrao:


stackit/internal/conversion/conversion.go
```diff
@@ -58,6 +59,22 @@ func ToTerraformStringMap(ctx context.Context, m map[string]string) (basetypes.M
 	return res, nil
 }
 
+// FromTerraformStringMapToInterfaceMap converts a basetypes.MapValue of Strings to a map[string]interface{}.
+func FromTerraformStringMapToInterfaceMap(ctx context.Context, m basetypes.MapValue) (map[string]interface{}, error) {
```
joaopalet, line 63: [Nitpick] Isn't `ToStringInterfaceMap` (or something similar) enough for the name? We can already tell by the signature it is from a TF map

stackit/internal/services/argus/instance/resource_test.go
```diff
@@ -162,6 +201,18 @@ func fixtureWebHooksConfigsGetPayload() argus.WebHook {
 	}
 }
 
+func fixtureRouteGetPayload() *argus.Route {
```
joaopalet, line 204: Just noticed it now, aren't these `fixtureRouteGetResponse` instead of `Payload`?

DiogoFerrao: Yes, thats a better name. I was using `GetPayload` and `UpdatePayload`

stackit/internal/conversion/conversion.go
DiogoFerrao: I was trying to match the naming of the `ToTerraformStringMap`, but you are right, it doesn't have the `From` part

stackit/internal/services/argus/argus_acc_test.go
```diff
@@ -233,13 +326,119 @@ func TestAccResource(t *testing.T) {
...
+					resource.TestCheckResourceAttrPair(
+						"stackit_argus_instance.instance", "instance_id",
+						"stackit_argus_credential.credential", "instance_id",
+					),
+					resource.TestCheckResourceAttrSet("stackit_argus_credential.credential", "username"),
+					resource.TestCheckResourceAttrSet("stackit_argus_credential.credential", "password"),
+				),
+				// This is needed because the GET request to the alert config endpoint doesn't return some of the global options
+				// and the emailConfigs. Therefore, the state file doesn't contain them and the diff will fail.
+				ExpectNonEmptyPlan: true,
```
joaopalet, line 432: :+1:

stackit/internal/services/argus/argus_acc_test.go
```diff
@@ -233,13 +326,119 @@ func TestAccResource(t *testing.T) {
...
+			},
+			// Creation without ACL, partial metrics retention days and NO alert configs
```
joaopalet, line 434: So is this a creation or update of the previously created?

stackit/internal/services/argus/argus_acc_test.go
```diff
@@ -439,8 +640,7 @@ func TestAccResource(t *testing.T) {
 					resource.TestCheckResourceAttr("data.stackit_argus_scrapeconfig.scrapeconfig", "saml2.enable_url_parameters", scrapeConfigResource["saml2_enable_url_parameters"]),
 				),
 			},
-
-			// Import
+			// Import 1
```
joaopalet, line 643: And is the import importing the instance from the last step? The that one does not have alert configs, right?

DiogoFerrao: It is importing from the last step, but the last step has alert configs. It doesn't have the emailConfigs because those create non-empty plans and I wanted to minimize the usage of `ExpectNonEmptyPlan`

stackit/internal/services/argus/argus_acc_test.go
DiogoFerrao: They are all updates, I will rename them

stackit/internal/services/argus/argus_acc_test.go
DiogoFerrao: The last test before importing is the `Data source` one, which has alert configs

joaopalet (approved):

PR #455, opened 2024-07-09
Document possible values of schema fields

joaopalet:


stackit/internal/utils/utils.go
```diff
@@ -76,3 +76,17 @@ func SimplifyBackupSchedule(schedule string) string {
 	})
 	return simplifiedSchedule
 }
+
+func PossibleValuesDocumentation(values []string) string {
+	if len(values) == 0 {
+		return ""
+	}
+	return "Possible values are: " + strings.Join(quoteValues(values), ", ") + "."
```
DiogoFerrao, line 84: `Possible values` doesn't sound strict enough, especially if we are then verifying that values are exactly one of them.

Something like `Value must be one of the following:`

DiogoFerrao: `Supported values:` sounds good

DiogoFerrao (approved):

PR #461, opened 2024-07-10
Fix inconsistent result after creating SKE cluster with empty node po…

joaopalet:
…ol taints

### DIFF
--- a/stackit/internal/services/ske/cluster/resource.go
+++ b/stackit/internal/services/ske/cluster/resource.go
@@ -1323,7 +1323,11 @@ func mapNodePools(ctx context.Context, cl *ske.Cluster, m *Model) error {
 			nodePool["cri"] = types.StringPointerValue(nodePoolResp.Cri.Name)
 		}
 
-		err := mapTaints(nodePoolResp.Taints, nodePool)
+		modelNodePoolTaints := types.ListNull(types.ObjectType{AttrTypes: taintTypes})
+		if i < len(modelNodePools) {
+			modelNodePoolTaints = modelNodePools[i].Taints
+		}
+		err := mapTaints(modelNodePoolTaints, nodePoolResp.Taints, nodePool)
 		if err != nil {
 			return fmt.Errorf("mapping index %d, field taints: %w", i, err)
 		}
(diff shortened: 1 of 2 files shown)


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1323,7 +1323,11 @@ func mapNodePools(ctx context.Context, cl *ske.Cluster, m *Model) error {
 			nodePool["cri"] = types.StringPointerValue(nodePoolResp.Cri.Name)
 		}
 
-		err := mapTaints(nodePoolResp.Taints, nodePool)
+		modelNodePoolTaints := types.ListNull(types.ObjectType{AttrTypes: taintTypes})
+		if i < len(modelNodePools) {
+			modelNodePoolTaints = modelNodePools[i].Taints
+		}
```
DiogoFerrao, line 1329: Can you explain what this does? I'm not sure I understand

DiogoFerrao:
Is it possible to set the taints to an empty list in the API?

stackit/internal/services/ske/cluster/resource.go
joaopalet: `i` is the index of the nodepool we are mapping. If `i < len(modelNodePools)` then that node pool is already in the model (we are in a create or update). Otherwise we can be are in a read (e.g. an import or a data source, where the model was previously `nil`)

joaopalet: But maybe there is a better was to do it

joaopalet:
> Is it possible to set the taints to an empty list in the API?

Yes, and we are doing that. The issue was that the API response is `nil` when the payload is `[]`

stackit/internal/services/ske/cluster/resource.go
DiogoFerrao: Ok I get it now., thanks!

But then even if we are in a read operation, the model should have a Taints property that would be Null or Unknown. 
I think the mapTaints function should receive the contents of the model directly and then do the mapping internally depending on the current state. This could be achieved with an if `.IsUnknown` or `.IsNull`. Right? This external mapping is breaking the separation of concerns and I don't think there is any gain

joaopalet: Adjusted the code. I think the logic to _map_ should be kept be in the `mapTaints` function, so I am just passing a bool to indicate if the current node pool model has anything in the taints field or not (not meaning that is either null or unknown)

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1323,7 +1323,14 @@ func mapNodePools(ctx context.Context, cl *ske.Cluster, m *Model) error {
...
 		}
 
-		err := mapTaints(nodePoolResp.Taints, nodePool)
+		taintsInModel := false
+		if i < len(modelNodePools) {
+			modelNodePoolTaints := modelNodePools[i].Taints
+			if !modelNodePoolTaints.IsNull() && !modelNodePoolTaints.IsUnknown() {
+				taintsInModel = true
+			}
+		}
```
DiogoFerrao, line 1332: I think this is simpler:

```golang
if i < len(modelNodePools) && !modelNodePools[i].Taints.IsNull() && !modelNodePools[i].Taints.IsUnknown() {
        taintsInModel = true
}
```

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1362,8 +1369,16 @@ func mapNodePools(ctx context.Context, cl *ske.Cluster, m *Model) error {
 	return nil
 }
 
-func mapTaints(t *[]ske.Taint, nodePool map[string]attr.Value) error {
+func mapTaints(t *[]ske.Taint, nodePool map[string]attr.Value, existInModel bool) error {
 	if t == nil || len(*t) == 0 {
+		if existInModel {
+			taintsTF, diags := types.ListValue(types.ObjectType{AttrTypes: taintTypes}, []attr.Value{})
+			if diags.HasError() {
+				return fmt.Errorf("convert empty taints list: %w", core.DiagsToError(diags))
```
DiogoFerrao, line 1377:
```suggestion
				return fmt.Errorf("create empty taints list: %w", core.DiagsToError(diags))
```

DiogoFerrao (approved):

PR #469, opened 2024-07-16
Onboard network-area resource

GokceGK:


stackit/internal/services/iaas/networkarea/resource.go
```diff
@@ -0,0 +1,678 @@
...
+			},
+			"network_ranges": schema.SetAttribute{
+				Description: "List of Network ranges.",
+				ElementType: types.StringType,
+				Required:    true,
+				Validators: []validator.Set{
+					setvalidator.SizeAtLeast(1),
+					setvalidator.SizeAtMost(64),
+				},
+			},
```
joaopalet, line 185: The network range field in the API is a list of objects, each with a `prefix` and a `networkRangeId` fields. Shouldn't we replicate this here in the provider instead o making it a list of just the string prefixes?

stackit/internal/services/iaas/networkarea/resource.go
```diff
@@ -0,0 +1,678 @@
...
+	payload, err := toCreatePayload(ctx, &model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating network area", fmt.Sprintf("Creating API payload: %v", err))
+		return
+	}
+
+	// Create new network area
+	var httpResp *http.Response
+	ctxWithHTTPResp := runtime.WithCaptureHTTPResponse(ctx, &httpResp)
+	area, err := r.client.CreateNetworkArea(ctxWithHTTPResp, organizationId).CreateNetworkAreaPayload(*payload).Execute()
```
joaopalet, line 250: Do we need to capture the raw HTTP response? I think it is not used anywhere here

stackit/internal/services/iaas/networkarea/resource.go
```diff
@@ -0,0 +1,678 @@
...
+		return
+	}
+	networkAreaId := *networkArea.AreaId
+	ctx = tflog.SetField(ctx, "network_area_id", networkAreaId)
+
+	networkAreaRangesResp, err := r.client.ListNetworkAreaRanges(ctx, organizationId, networkAreaId).Execute()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading network area ranges", fmt.Sprintf("Calling API for network area  data: %v", err))
+		return
+	}
```
joaopalet, line 268: Is this extra call needed? I think the network area ranges already come in the response of `CreateNetworkArea`, in this case they will be under `networkArea`

stackit/internal/services/iaas/networkarea/resource.go
```diff
@@ -0,0 +1,678 @@
...
+		}
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading network area", fmt.Sprintf("Calling API: %v", err))
+		return
+	}
+
+	networkAreaRangesResp, err := r.client.ListNetworkAreaRanges(ctx, organizationId, networkAreaId).Execute()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading network area ranges", fmt.Sprintf("Calling API for network area  data: %v", err))
+		return
+	}
```
joaopalet, line 313: Same here

stackit/internal/services/iaas/networkarea/datasource.go
```diff
@@ -0,0 +1,254 @@
...
+						},
+						"prefix": schema.StringAttribute{
+							Computed: true,
+						},
+						"next_hop": schema.StringAttribute{
+							Computed: true,
+						},
+					},
+				},
+			},
```
joaopalet, line 180: Won't this fail because the `routes` fields is not yet in the model? (because it's part of a follow-up story, right?)

joaopalet: Even if not, since this is not populated yet, I think we could remove this from the schema and add it in that story only

stackit/provider.go
```diff
@@ -425,6 +426,7 @@ func (p *Provider) Resources(_ context.Context) []func() resource.Resource {
 		dnsZone.NewZoneResource,
 		dnsRecordSet.NewRecordSetResource,
 		iaasNetwork.NewNetworkResource,
+		iaasNetworkArea.NewNetworkAreaResource,
```
joaopalet, line 429: The datasource also needs to be added to the provider

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -25,25 +25,54 @@ var networkResource = map[string]string{
...
+	)
+}
+
+func resourceConfig(name, nameservers, areaname, networkranges string) string {
+	return fmt.Sprintf("%s\n\n%s\n\n%s",
+		testutil.IaaSProviderConfig(),
+		networkResourceConfig(name, nameservers),
+		networkAreaResourceConfig(areaname, networkranges),
+	)
+}
```
joaopalet, line 74: Nice :+1:

stackit/internal/services/iaas/networkarea/resource.go
GokceGK: @joaopalet -> It is only used in create network area function, nowhere else. 

Should we also remove raw resp from the `network` resource? https://github.com/stackitcloud/terraform-provider-stackit/blob/af7d7899455de0348679c91d05cc95b18f0e788b/stackit/internal/services/iaas/network/resource.go#L186C20-L186C33

joaopalet: There we need it to get the `x-request-id` header, that is needed to get more details about the network

stackit/internal/services/iaas/networkarea/resource.go
```diff
@@ -0,0 +1,702 @@
...
+					listvalidator.SizeAtLeast(1),
+					listvalidator.SizeAtMost(64),
+				},
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"prefix": schema.StringAttribute{
+							Required: true,
+						},
+					},
+				},
```
joaopalet, line 198:
```suggestion
				NestedObject: schema.NestedAttributeObject{
					Attributes: map[string]schema.Attribute{
						"prefix": schema.StringAttribute{
                            Description: "Classless Inter-Domain Routing (CIDR)."
							Required: true,
						},
					},
				},
```

stackit/internal/services/iaas/networkarea/resource.go
joaopalet: We could also add a field for the computed `networkRangeId` field, right? (it is returned in the response of the [`CreateArea`](https://docs.api.stackit.cloud/documentation/iaas/version/v1#tag/Area/operation/v1CreateArea) endpoint)

GokceGK: thx for the catch, I had it in datasource but not in resource :(

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -25,25 +25,55 @@ var networkResource = map[string]string{
 	"nameserver1":        "5.6.7.8",
 }
 
-func resourceConfig(name, nameservers string) string {
-	return fmt.Sprintf(`
-				%s
+var networkAreaResource = map[string]string{
+	"organization_id":  testutil.OrganizationId,
+	"name":             fmt.Sprintf("acc-test-gg-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)),
```
joaopalet, line 30:
```suggestion
	"name":             fmt.Sprintf("acc-test-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)),
```

joaopalet (approved):

Issue #462, opened 2024-07-10
Creating a loadbalancer with private_network_only will result in an error


h3adex:
Creating a loadbalancer with private_network_only will result in an error. I've used the code supplied in the provider [documentation](https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/resources/loadbalancer).

Code:
```tf
resource "stackit_loadbalancer" "lb_example_1" {
  project_id = var.project_id
  name       = "lb-example-1"

  ...

  options = {
    private_network_only = false
  }
}
```

![Screenshot 2024-07-10 at 14 37 40](https://github.com/stackitcloud/terraform-provider-stackit/assets/39736813/00814910-91de-411a-a6b9-2f6f9ae65ea0)

joaopalet:
Hello @h3adex,

Thanks for reporting the issue! We will be looking into it soon.

In the meantime, if the API returns `nil` if you set the value to `false` it probably means they are defaulting to `false`. So your configuration would be equivalent to one where the field is not specified.

h3adex:
tyvm

joaopalet:
Hi @h3adex,

This was fixed in [`v0.25.1`](https://github.com/stackitcloud/terraform-provider-stackit/releases/tag/v0.25.1).

PR #477, opened 2024-07-22
Adjust mapping of LB API response when private network only field is …

joaopalet:
…null

### DIFF
--- a/stackit/internal/services/loadbalancer/loadbalancer/resource.go
+++ b/stackit/internal/services/loadbalancer/loadbalancer/resource.go
@@ -591,7 +591,7 @@ func (r *loadBalancerResource) Create(ctx context.Context, req resource.CreateRe
 	}
 
 	// Map response body to schema
-	err = mapFields(waitResp, &model)
+	err = mapFields(ctx, waitResp, &model)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


joaopalet:
Addresses #462

stackit/internal/services/loadbalancer/loadbalancer/resource.go
```diff
@@ -1192,14 +1192,29 @@ func mapNetworks(loadBalancerResp *loadbalancer.LoadBalancer, m *Model) error {
...
+func mapOptions(ctx context.Context, loadBalancerResp *loadbalancer.LoadBalancer, m *Model) error {
 	if loadBalancerResp.Options == nil {
 		m.Options = types.ObjectNull(optionsTypes)
 		return nil
 	}
 
+	privateNetworkOnlyTF := types.BoolPointerValue(loadBalancerResp.Options.PrivateNetworkOnly)
+
+	// If the private_network_only field is nil in the response but is explicitly set to false in the model,
+	// we set it to false in the TF state to prevent an inconsistency result after apply error
```
DiogoFerrao, line 1204:
```suggestion
	// we set it to false in the TF state to prevent an inconsistent result after apply error
```

DiogoFerrao (approved):

PR #480, opened 2024-07-25
Allow managing members in the project resource

joaopalet:


stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -421,42 +522,123 @@ func mapFields(ctx context.Context, projectResp *resourcemanager.GetProjectRespo
...
+		return &[]authorization.Member{}, nil
+	}
 
-	owner := projectOwner
-	serviceAccountSubject := serviceAccountEmail
-	members := []resourcemanager.Member{
-		{
-			Subject: &serviceAccountSubject,
-			Role:    &owner,
-		},
```
joaopalet, line 438: The service account was being silently set as an owner of the project before, but this was not documented anywhere and it would case an inconsistent result after apply in the `members` field afterwards unless we filtered it out (which would be misleading)

joaopalet: I think the simplest approach is to let the users choose all members of the project themselves, as he would do in the API

stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -220,14 +296,20 @@ func (r *projectResource) Create(ctx context.Context, req resource.CreateRequest
...
 	if err != nil {
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating project", fmt.Sprintf("Instance creation waiting: %v", err))
 		return
 	}
 
+	membersResp, err := r.authorizationClient.ListMembersExecute(ctx, projectResourceType, *waitResp.ProjectId)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating project", fmt.Sprintf("Reading members: %v", err))
+		return
+	}
```
DiogoFerrao, line 309: We should probably split the mapping here as well, if for some reason this call fails, the resource is created and not mapped

stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -297,21 +385,34 @@ func (r *projectResource) Update(ctx context.Context, req resource.UpdateRequest
...
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating project", fmt.Sprintf("Calling API for updated data: %v", err))
+		return
+	}
+
+	members, err := toMembersPayload(ctx, &model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating project", fmt.Sprintf("Processing members: %v", err))
+		return
+	}
```
DiogoFerrao, line 405: Also here, we should set the state before we updating the members

stackit/internal/services/resourcemanager/project/resource.go
DiogoFerrao: I agree

stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -478,3 +660,104 @@ func toUpdatePayload(model *Model) (*resourcemanager.PartialUpdateProjectPayload
...
+	if err != nil {
+		return fmt.Errorf("get members: %w", err)
+	}
+
+	type memberState struct {
+		isInModel bool
+		isCreated bool
+		subject   string
+		role      string
+	}
```
DiogoFerrao, line 681: Nice 👍

stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -478,3 +660,104 @@ func toUpdatePayload(model *Model) (*resourcemanager.PartialUpdateProjectPayload
...
+	if len(membersToRemove) > 0 {
+		payload := authorization.RemoveMembersPayload{
+			Members:      &membersToRemove,
+			ResourceType: sdkUtils.Ptr(projectResourceType),
+		}
+		_, err := client.RemoveMembers(ctx, projectId).RemoveMembersPayload(payload).Execute()
+		if err != nil {
+			return fmt.Errorf("remove members: %w", err)
+		}
+	}
```
DiogoFerrao, line 755: Good job on the logic!

As mentioned previously, I'm worried about the terraform state if any of these fail. Here we could set the state inside these methods and between calls. It should also address the previous comments

stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -78,42 +98,65 @@ func (r *projectResource) Configure(ctx context.Context, req resource.ConfigureR
...
+		"name":                            "Project name.",
+		"labels":                          "Labels are key-value string pairs which can be attached to a resource container. A label key must match the regex [A-ZÄÜÖa-zäüöß0-9_-]{1,64}. A label value must match the regex ^$|[A-ZÄÜÖa-zäüöß0-9_-]{1,64}",
+		"owner_email":                     "Email address of the owner of the project. This value is only considered during creation. Changing it afterwards will have no effect.",
+		"owner_email_deprecation_message": "The \"owner_email\" field has been deprecated in favor of the \"members\" field. Please use the \"members\" field to assign the owner role to a user, by setting the \"role\" field to `owner`.",
+		"members":                         "The members assigned to the project. At least one subject needs to be a user, and not a client or service account.",
+		"members.role":                    fmt.Sprintf("A valid role defined for the resource. At least one users must have the `owner` role. Legacy roles (%s) are not supported.", strings.Join(utils.QuoteValues(utils.LegacyProjectRoles), ", ")),
```
DiogoFerrao, line 158:
```suggestion
		"members.role":                    fmt.Sprintf("The role of the member in the project. At least one user must have the `owner` role. Legacy roles (%s) are not supported.", strings.Join(utils.QuoteValues(utils.LegacyProjectRoles), ", ")),
```

stackit/internal/services/resourcemanager/project/resource_test.go
```diff
@@ -166,22 +206,72 @@ func TestToCreatePayload(t *testing.T) {
...
+						Subject: utils.Ptr("reader_email"),
+						Role:    utils.Ptr("reader"),
 					},
 				},
-				Name: nil,
+				Name: utils.Ptr("name"),
 			},
 			true,
 		},
 		{
```
DiogoFerrao, line 260: I think its better to keep the test case where only the `OwnerEmail` is provided, since this will still be possible, right?

stackit/internal/services/resourcemanager/project/resource_test.go
```diff
@@ -322,3 +408,309 @@ func TestToUpdatePayload(t *testing.T) {
...
+		},
+		{
+			description:        "remove fails",
+			modelMembers:       fixtureMembers[:1],
+			removeMembersFails: true,
+			isValid:            false,
+		},
+	}
+
+	for _, tt := range tests {
```
DiogoFerrao, line 563: Great job on the testing logics, looks nice 😊

stackit/internal/services/resourcemanager/project/datasource.go
```diff
@@ -153,13 +167,37 @@ func (d *projectDataSource) Schema(_ context.Context, _ datasource.SchemaRequest
 					),
 				},
 			},
+			"owner_email": schema.StringAttribute{
+				Description:        descriptions["owner_email"],
+				DeprecationMessage: "This field has been deprecated in favor of the \"members\" field. Please use the \"members\" field to assign the owner role to a user, by setting the \"role\" field to `owner`.",
```
DiogoFerrao, line 172: This could be in the `descriptions` as in the resource

stackit/internal/services/resourcemanager/project/resource.go
joaopalet: I think to update the state after adding the members we would need to make an extra API call anyway (list members), which could also fail

GokceGK (approved):

PR #510, opened 2024-08-22
Rename argus to observability

GokceGK:


README.md
```diff
@@ -11,6 +11,15 @@
...
 
+> [!WARNING]
+> On August 26 2024, The STACKIT Argus service was renamed to STACKIT Observability.
+>
+> This means that there are new resources for `STACKIT Observability`, which offers the same functionality as the deprecated `STACKIT Argus` resources.
+>
+> Please make sure to **update your STACKIT Terraform Provider to the latest version after August 26 2024** to ensure that you start using STACKIT Observability resources.
+> 
+> You can migrate your existing STACKIT Argus resources by using the `terraform import`. https://developer.hashicorp.com/terraform/language/import
+
```
joaopalet, line 22: I think it's enough to add this message to the changelog of the release instead of the README

joaopalet: Also we should probably also give a hint on how to remove the existing Argus resource without destroying it: remove from TF config and remove from state (`terraform state rm stackit_argus_instance.RESOURCE_NAME`)

MIGRATION.md
```diff
@@ -128,36 +128,39 @@ resource "stackit_logme_credential" "example-credential" {
...
-| stackit_rabbitmq_instance | stackit_rabbitmq_instance | :white_check_mark: | [project_id],[instance_id] |  |
-| stackit_redis_credential | stackit_redis_credential | :white_check_mark: | [project_id],[credentials_group_id],[credential_id] |  |
-| stackit_redis_instance | stackit_redis_instance | :white_check_mark: | [project_id],[instance_id] |  |
-| stackit_secrets_manager_instance | stackit_secretsmanager_instance | :white_check_mark: | [project_id],[instance_id] |  |
-| stackit_secrets_manager_user | stackit_secretsmanager_user | :warning: | [project_id],[instance_id],[user_id] | `password` field will be empty |
+| stackit_object_storage_project           |                                         | | | Resource deprecated |
+| stackit_observability_credential         | stackit_observability_credential        | :x: | |  |
+| stackit_observability_instance           | stackit_observability_instance          | :white_check_mark: | [project_id],[instance_id] |  |
+| stackit_observability_job                | stackit_observability_scrapeconfig      | :white_check_mark: | [project_id],[instance_id],[name] |  |
```
joaopalet, line 153: These are not part of the community provider, so the hint should be added to lines 133-135 (something like "Service deprecated, use stackit_observability_...", and keep the "`id` format" column

README.md
GokceGK: removed the info from readme and left a comment in the work item to add it in the release changelog. I will also add the second part in theere.

MIGRATION.md
GokceGK: added back the id infos and extended the notes part as suggested

joaopalet (approved):

PR #531, opened 2024-09-12
Handle project members

GokceGK:


stackit/internal/services/resourcemanager/resourcemanager_acc_test.go
```diff
@@ -56,6 +56,7 @@ func resourceConfig(name string, label *string, members string) string {
 					members = [
 						%[7]s
 					]
```
joaopalet, line 58: The `members` field (initial members now) will still have priority when specified, right? So in this case the `owner_email` field has no effect, right?test

joaopalet: We can remove the `members` field from the test when really removing the field, but was just asking

stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -147,17 +147,18 @@ func (r *projectResource) Configure(ctx context.Context, req resource.ConfigureR
...
+		"id":                          "Terraform's internal resource ID. It is structured as \"`container_id`\".",
+		"project_id":                  "Project UUID identifier. This is the ID that can be used in most of the other resources to identify the project.",
+		"container_id":                "Project container ID. Globally unique, user-friendly identifier.",
+		"parent_container_id":         "Parent resource identifier. Both container ID (user-friendly) and UUID are supported",
+		"name":                        "Project name.",
+		"labels":                      "Labels are key-value string pairs which can be attached to a resource container. A label key must match the regex [A-ZÄÜÖa-zäüöß0-9_-]{1,64}. A label value must match the regex ^$|[A-ZÄÜÖa-zäüöß0-9_-]{1,64}",
+		"owner_email":                 "Email address of the owner of the project. This value is only considered during creation. Changing it afterwards will have no effect.",
+		"members":                     "The members assigned to the project. At least one subject needs to be a user, and not a client or service account.",
```
joaopalet, line 158: Since now this field only specifies the _initial_ members, I think we should write it in the description

stackit/internal/services/resourcemanager/project/resource.go
```diff
@@ -224,12 +225,13 @@ func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 			},
 			"owner_email": schema.StringAttribute{
 				Description: descriptions["owner_email"],
-				// When removing the owner_email field, we should mark the members field as required and add a listvalidator.SizeAtLeast(1) validator to it
-				Optional: true,
+				Required:    true,
```
joaopalet, line 228: The goal of deprecating the `members` field instead of removing is to not cause a breaking change immediatelly, right?

joaopalet: If so we should keep the `owner_email` as optional during the deprecation period, otherwise it will be breaking anyway

stackit/internal/services/resourcemanager/resourcemanager_acc_test.go
GokceGK: yes, that is true. members field still have precedence over owner_email. 
I kind of added owner_email to be prepared for the removal of members field after deprecation is over.

stackit/internal/services/resourcemanager/project/resource.go
GokceGK: @joaopalet  Changed description to _"The members assigned to the project. At least one subject needs to be a user, and not a client or service account. This value is only considered during creation. Changing it afterwards will have no effect."_

PR #546, opened 2024-10-03
IaaS Server baseline configuration

vicentepinto98:


stackit/internal/services/iaas/server/resource.go
```diff
@@ -0,0 +1,770 @@
...
+			},
+			"boot_volume": schema.SingleNestedAttribute{
+				Description: "The boot volume for the server",
+				Optional:    true,
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.RequiresReplace(),
+				},
+				Attributes: map[string]schema.Attribute{
+					"delete_on_termination": schema.BoolAttribute{
+						Description: "Delete the volume during the termination of the server. Defaults to `false`.",
```
joaopalet, line 289: Weren't we going to leave this off the provider for now?

joaopalet: Because the volume might also be configured in TF

stackit/internal/services/iaas/server/resource.go
```diff
@@ -0,0 +1,770 @@
...
+						Optional:    true,
+						PlanModifiers: []planmodifier.Bool{
+							boolplanmodifier.RequiresReplace(),
```
joaopalet, line 292: If we want to keep it, does it really requires replace? Isn't it some flag we can toggle that will change the behaviour just on deletion?

joaopalet: Ah I see it's only used in creation

stackit/internal/services/iaas/server/resource.go
vicentepinto98: If the source is an image, the volume can be deleted safely on termination because there is no volume resource on the user configuration. 

But we should validate and only allow setting this field if `source_type` is `image`

stackit/internal/services/iaas/server/resource.go
```diff
@@ -0,0 +1,770 @@
...
+				Validators: []validator.String{
+					stringvalidator.LengthAtLeast(1),
+					stringvalidator.LengthAtMost(63),
+					stringvalidator.RegexMatches(
+						regexp.MustCompile(`^[A-Za-z0-9]+((-|\.)[A-Za-z0-9]+)*$`),
+						"must match expression"),
+				},
+			},
+			"machine_type": schema.StringAttribute{
+				Description: "Name of the machine type the server will belong to.",
```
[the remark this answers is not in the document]
vicentepinto98: Yes, but it's a huge list. Returned by the GET machine-types endpoint. I will add a link to a documentation page where all machine types are listed: https://docs.stackit.cloud/stackit/en/virtual-machine-flavors-75137231.html

stackit/internal/services/iaas/server/resource.go
```diff
@@ -0,0 +1,770 @@
...
+// needs to exist because the Configure method is called twice.
+var resourceBetaCheckDone bool
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource                = &serverResource{}
+	_ resource.ResourceWithConfigure   = &serverResource{}
+	_ resource.ResourceWithImportState = &serverResource{}
+
+	SupportedSourceTypes = []string{"volume", "image"}
```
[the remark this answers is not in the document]
vicentepinto98: it's also used on the `datasource` that's why it must be global. But I made it private :+1:

stackit/internal/services/iaas/server/resource.go
vicentepinto98: decided to remove it. It would require a custom validator, we can add later if needed

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -44,13 +49,26 @@ var networkAreaRouteResource = map[string]string{
...
+var serverResource = map[string]string{
+	"project_id":        testutil.ProjectId,
+	"availability_zone": "eu01-1",
+	"size":              "64",
+	"source_type":       "image",
+	"source_id":         testutil.IaaSImageId,
+	"name":              fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlpha)),
+	"machine_type":      serverMachineType,
+	"label1":            "value",
+	"user_data":         "#!/bin/bash",
```
joaopalet, line 69: In the description it says it must be base64 encoded, isn't it a requirement on the API side then?

vicentepinto98: Ups, I need to update the description. We decided to use a string value on the field and we do the conversion in the provider logic

joaopalet (approved):

PR #553, opened 2024-10-10
Onboard iaas security group rule

GokceGK:


stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,658 @@
...
+	err = mapProtocol(securityGroupRuleResp, model)
+	if err != nil {
+		return fmt.Errorf("map protocol: %w", err)
+	}
+
+	return nil
+}
+
+func mapIcmpParameters(securityGroupRuleResp *iaasalpha.SecurityGroupRule, m *Model) error {
+	if securityGroupRuleResp.IcmpParameters == nil {
```
vicentepinto98, line 532: Did you test the case where the user sets the object fields, e.g. `icmp_parameters` to an empty object? In this case they would be returned by the API as nil, but we should map them to the empty object otherwise there will be value conversion error. 
 
 This doesn't apply when the object field has all fields inside as `Required` because it is not allowed to set the empty object in the configuration, but please double check for all of these object fields
 
 See how we are doing it for the labels field for example

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -83,6 +83,14 @@ var securityGroupResource = map[string]string{
 	"label1":      "value",
 }
 
+// Security Group rule resource data
+var securityGroupRuleResource = map[string]string{
+	"project_id":        testutil.ProjectId,
+	"security_group_id": securityGroupResource["security_group_id"],
```
vicentepinto98, line 89: this can be removed right? I think it doesnt even exist :thinking:

examples/resources/stackit_security_group_rule/resource.tf
```diff
@@ -0,0 +1,5 @@
+resource "stackit_security_group_rule" "example" {
```
vicentepinto98, line 1: it would be nice to add examples for the object fields that define a rule, otherwise the example is not that useful

GokceGK: added object fields in the example

stackit/internal/services/iaas/iaas_acc_test.go
GokceGK: removed the obsolete field

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,653 @@
...
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
+				},
+			},
+			"security_group_id": schema.StringAttribute{
+				Description: "The security group ID.",
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
```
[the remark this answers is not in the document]
GokceGK: removed

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,653 @@
...
+				},
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
+				},
+			},
+			"description": schema.StringAttribute{
+				Description: "The rule description.",
+				Optional:    true,
+				Computed:    true,
```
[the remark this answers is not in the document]
GokceGK: I have verified again. 
`Description, Direction, EtherType, IpRange, PortRange (Min, Max) and Protocol (Name)` are returned even if I do not send them in the request. I kept `Optional` and `Computed` in these fields and added `UseStateForUnknown` (as there is no update method for security group rules).

Other fields where I had a `Computed`, I removed them.

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,658 @@
...
+	diags = resp.State.Set(ctx, model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	tflog.Info(ctx, "security group rule read")
+}
+
+// Update updates the resource and sets the updated Terraform state on success.
+func (r *securityGroupRuleResource) Update(ctx context.Context, _ resource.UpdateRequest, resp *resource.UpdateResponse) { // nolint:gocritic // function signature required by Terraform
```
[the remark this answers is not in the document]
GokceGK: Added `RequiresReplace` to every field

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,697 @@
...
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"icmp_parameters": schema.SingleNestedAttribute{
+				Description: "ICMP Parameters.",
+				Optional:    true,
```
vicentepinto98, line 232: you can also add `RequiresReplace` at this level, just for safety

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,697 @@
...
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					validate.IP(),
+				},
+			},
+			"port_range": schema.SingleNestedAttribute{
+				Description: "The range of ports.",
+				Optional:    true,
+				Computed:    true,
```
vicentepinto98, line 273: same here

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,697 @@
...
+					"name": schema.StringAttribute{
+						Description: "The protocol name which the rule should match.",
+						Optional:    true,
+						Computed:    true,
+						PlanModifiers: []planmodifier.String{
+							stringplanmodifier.UseStateForUnknown(),
+							stringplanmodifier.RequiresReplace(),
+						},
+					},
+					"protocol": schema.Int64Attribute{
```
vicentepinto98, line 317: would it make more sense to name this field `number` instead? `protocol.protocol` path seems weird

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,697 @@
...
+							int64validator.AtLeast(0),
+							int64validator.AtMost(65535),
+						},
+					},
+				},
+			},
+			"protocol": schema.SingleNestedAttribute{
+				Description: "The internet protocol which the rule should match.",
+				Optional:    true,
+				Computed:    true,
```
vicentepinto98, line 306: same here

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,697 @@
...
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error configuring API client", fmt.Sprintf("Configuring client: %v. This is an error related to the provider configuration, not to the resource configuration", err))
+		return
+	}
+
+	r.client = apiClient
+	tflog.Info(ctx, "iaasalpha client configured")
+}
+
+// Schema defines the schema for the resource.
+func (r *securityGroupRuleResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
```
vicentepinto98, line 153: in general, can you check the openstack provider resource and see if it makes sense to have all these object fields instead of a more flat schema? https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/networking_secgroup_rule_v2

You could also check this with the IaaS team.

Maybe something we could have checked in the refinement but we are still in time to make the best user experience here

vicentepinto98: It would also remove a lot of the object mapping complexity, including that empty object case

vicentepinto98: After aligning with the IaaS colleagues, we decided to leave them as objects

stackit/internal/services/iaas/securitygrouprule/resource.go
vicentepinto98: rename to number, they will probably do the same in the API

stackit/internal/services/iaas/securitygrouprule/resource.go
vicentepinto98: Beware that some of these fields should only be set if a specific protocol is set. E.g:
- `icmp_parameters` **should only** be provided if the protocol is `icmp`
- `port_range` **shouldn't** be provided if protocol is `icmp`

We can't easily validate this in the schema I think, because it would require checking other fields values, but at least we should add to the description

stackit/internal/services/iaas/securitygrouprule/resource.go
GokceGK: Added empty field checks for the optional fields. Also extended unit tests

stackit/internal/services/iaas/securitygrouprule/resource.go
GokceGK: Added `RequiresReplace`

stackit/internal/services/iaas/securitygrouprule/resource.go
GokceGK: Added `RequiresReplace`

stackit/internal/services/iaas/securitygrouprule/resource.go
GokceGK: Added `RequiresReplace`

stackit/internal/services/iaas/securitygrouprule/resource.go
GokceGK: I renamed the sub `protocol` field to `number`. 
In api it is still `protocol`, that is why some places it still looks like `protocol.protocol`. But from terraform implementation side, it is `protocol.number`

stackit/internal/services/iaas/securitygrouprule/resource.go
GokceGK: Added infos to the description. 
I used the examples you gave. Did not find any extra infos in API.

stackit/internal/services/iaas/securitygrouprule/resource_test.go
```diff
@@ -0,0 +1,294 @@
...
+				IpRange:               types.StringValue("iprange"),
+				RemoteSecurityGroupId: types.StringValue("remote"),
+				IcmpParameters:        fixtureModelIcmpParameters,
+				PortRange:             fixtureModelPortRange,
+				Protocol:              fixtureModelProtocol,
+			},
+			true,
+		},
+		{
+			"empty_port_range",
```
vicentepinto98, line 114: this test case is no longer relevant because port_range can't be empty

stackit/internal/services/iaas/securitygrouprule/resource_test.go
```diff
@@ -0,0 +1,294 @@
...
+				PortRange: types.ObjectValueMust(portRangeTypes, map[string]attr.Value{
+					"max": types.Int64Null(),
+					"min": types.Int64Null(),
+				}),
+				Protocol: types.ObjectNull(protocolTypes),
+			},
+			true,
+		},
+		{
+			"empty_protocol",
```
vicentepinto98, line 144: what this test case should cover is if the user has set an empty object but the API returns nil, which should not throw value conversion error, so it should be the following: 
```
Model{
				ProjectId:           types.StringValue("pid"),
				SecurityGroupId:     types.StringValue("sgid"),
				SecurityGroupRuleId: types.StringValue("sgrid"),
                                Protocol: types.ObjectValueMust(protocolTypes, map[string]attr.Value{
					"name":   types.StringNull(),
					"number": types.Int64Null(),
				})
			},
			&iaasalpha.SecurityGroupRule{
				Id:       utils.Ptr("sgrid"),
			},
			Model{
				Id:                    types.StringValue("pid,sgid,sgrid"),
				ProjectId:             types.StringValue("pid"),
				SecurityGroupId:       types.StringValue("sgid"),
				SecurityGroupRuleId:   types.StringValue("sgrid"),
				Direction:             types.StringNull(),
				Description:           types.StringNull(),
				EtherType:             types.StringNull(),
				IpRange:               types.StringNull(),
				RemoteSecurityGroupId: types.StringNull(),
				IcmpParameters:        types.ObjectNull(icmpParametersTypes),
				PortRange:             types.ObjectNull(portRangeTypes),
				Protocol: types.ObjectValueMust(protocolTypes, map[string]attr.Value{
					"name":   types.StringNull(),
					"number": types.Int64Null(),
				}),
			},
```

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -323,6 +323,7 @@ func (r *securityGroupRuleResource) Schema(_ context.Context, _ resource.SchemaR
 					"number": schema.Int64Attribute{
 						Description: "The protocol number which the rule should match.",
 						Optional:    true,
+						Computed:    true,
 						PlanModifiers: []planmodifier.Int64{
 							int64planmodifier.RequiresReplace(),
```
vicentepinto98, line 328: Add UseStateForUnknown

stackit/internal/services/iaas/securitygrouprule/resource.go
```diff
@@ -0,0 +1,721 @@
...
+				},
+			},
+			"protocol": schema.SingleNestedAttribute{
+				Description: "The internet protocol which the rule should match.",
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.RequiresReplace(),
+				},
+				Attributes: map[string]schema.Attribute{
```
vicentepinto98, line 313: just confirmed with IaaS, at least one of `name` and `number` must be provided so we must add `AtLeastOneOf` object validator: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators/objectvalidator#AtLeastOneOf`

stackit/internal/services/iaas/securitygrouprule/resource_test.go
GokceGK: If protocol needs to have name or number, this case will be obsolete, right

vicentepinto98: yes, not relevant anymore

GokceGK: removed test

stackit/internal/services/iaas/securitygrouprule/resource.go
GokceGK: added the `UseStateForUnknown`

stackit/internal/services/iaas/securitygrouprule/resource.go
GokceGK: added the validator. Also added the info in the description

vicentepinto98 (approved):

PR #556, opened 2024-10-14
Remove initial networking field from server

vicentepinto98:
- After aligning withe IaaS team, we decide to remove this field to improve user experience and make it very clear and transparent how to setup networking for a server which is going to be done solely by separate network interface attachment resource


joaopalet (approved):
Just missing doc generation, but no need to update in this specific PR (feature branch)

PR #567, opened 2024-10-18
Fix credential and credential group not found issue

GokceGK:


stackit/internal/services/objectstorage/credentialsgroup/resource.go
```diff
@@ -198,9 +200,14 @@ func (r *credentialsGroupResource) Read(ctx context.Context, req resource.ReadRe
 	ctx = tflog.SetField(ctx, "project_id", projectId)
 	ctx = tflog.SetField(ctx, "credentials_group_id", credentialsGroupId)
 
-	found, err := readCredentialsGroups(ctx, &model, r.client)
-	if err != nil {
-		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading credentialsGroup", fmt.Sprintf("getting credential group from list of credentials groups: %v", err))
+	found, formattedErr, err := readCredentialsGroups(ctx, &model, r.client)
+	if formattedErr != nil {
```
vicentepinto98, line 204: I don't think we need this `formattedErr`, we should make the check for 404 not found error status inside the `readCredentialsGroups` method. And if the error is 404, then we return `found` as `false`, which will already remove the resource from the state in the line 213 below

GokceGK: I have simplified the error handling as suggested

vicentepinto98 (approved):

PR #576, opened 2024-10-24
Extend network resource fields

GokceGK:
**Adds fields:**
- `ipv4_gateway` (String) The IPv4 gateway of a network. If not specified, the first IP of the network will be assigned as the gateway.
- `no_ipv4_gateway` (Boolean) If set to `true`, the network doesn't have a gateway.
- `ipv4_nameservers` (List of String) The IPv4 nameservers of the network.
- `ipv4_prefix` (String) The IPv4 prefix of the network (CIDR).
- `ipv4_prefixes` (List of String) The IPv4 prefixes of the network.
- `ipv6_gateway` (String) The IPv6 gateway of a network. If not specified, the first IP of the network will be assigned as the gateway.
- `no_ipv6_gateway` (Boolean) If set to `true`, the network doesn't have a gateway.
- `ipv6_nameservers` (List of String) The IPv6 nameservers of the network.
- `ipv6_prefix` (String) The IPv6 prefix of the network (CIDR).
- `ipv6_prefix_length` (Number) The IPv6 prefix length of the network.
- `ipv6_prefixes` (List of String) The IPv6 prefixes of the network.
- `routed` (Boolean) If set to `true`, the network is routed and therefore accessible from other networks.

**Deprecates fields:**
- `nameservers` (List of String, Deprecated) The nameservers of the network. This field is deprecated and will be removed soon, use `ipv4_nameservers` to configure the nameservers for IPv4.
- `prefixes` (List of String, Deprecated) The prefixes of the network. This field is deprecated and will be removed soon, use `ipv4_prefixes` to read the prefixes of the IPv4 networks.


stackit/internal/services/iaas/network/resource.go
```diff
@@ -40,10 +41,19 @@ type Model struct {
 	NetworkId        types.String `tfsdk:"network_id"`
 	Name             types.String `tfsdk:"name"`
 	Nameservers      types.List   `tfsdk:"nameservers"`
+	IPv4Gateway      types.String `tfsdk:"ipv4_gateway"`
+	IPv4Nameservers  types.List   `tfsdk:"ipv4_nameservers"`
+	IPv4Prefix       types.String `tfsdk:"ipv4_prefix"`
 	IPv4PrefixLength types.Int64  `tfsdk:"ipv4_prefix_length"`
 	Prefixes         types.List   `tfsdk:"prefixes"`
```
joaopalet, line 48: What is the difference between `Prefixes` and `IPv4Prefix`?

stackit/internal/services/iaas/network/resource.go
```diff
@@ -40,10 +41,19 @@ type Model struct {
...
+	IPv6Gateway      types.String `tfsdk:"ipv6_gateway"`
+	IPv6Nameservers  types.List   `tfsdk:"ipv6_nameservers"`
+	IPv6Prefix       types.String `tfsdk:"ipv6_prefix"`
+	IPv6PrefixLength types.Int64  `tfsdk:"ipv6_prefix_length"`
+	IPv6Prefixes     types.List   `tfsdk:"ipv6_prefixes"`
```
joaopalet, line 53: Same for `IPv6Prefixes` and `IPv6Prefix`

stackit/internal/services/iaas/network/resource.go
joaopalet: Couldn't they be the same field?

joaopalet: One is computed only and the other is optional

docs/resources/network.md
```diff
@@ -16,11 +16,26 @@ Network resource schema. Must have a `region` specified in the provider configur
...
   }
+  routed = false
+}
+resource "stackit_network" "example_ipv6" {
+  project_id         = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+  name               = "example-network"
+  ipv6_nameservers   = ["2001:4860:4860::8888", "2001:4860:4860::8844"]
+  ipv6_prefix_length = 56
+  ipv6_gateway       = "10.1.2.1"
+  ipv6_prefix        = "10.1.2.0/24"
```
joaopalet, line 34: These are IPv4 values, right?

docs/resources/network.md
```diff
@@ -16,11 +16,26 @@ Network resource schema. Must have a `region` specified in the provider configur
...
+  ipv4_nameservers   = ["1.2.3.4", "5.6.7.8"]
   ipv4_prefix_length = 24
+  ipv4_gateway       = "10.1.2.1"
+  ipv4_prefix        = "10.1.2.0/24"
   labels = {
     "key" = "value"
   }
+  routed = false
+}
+resource "stackit_network" "example_ipv6" {
```
joaopalet, line 28:
```suggestion
}

resource "stackit_network" "example_ipv6" {
```

stackit/internal/services/iaas/network/resource.go
```diff
@@ -141,20 +151,78 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+			"ipv4_gateway": schema.StringAttribute{
+				Description: "The IPv4 gateway of a network. If not specified, the first IP of the network will be assigned as the gateway.",
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.String{
+					validate.IP(),
+				},
+			},
+			"ipv4_nameservers": schema.ListAttribute{
+				Description: "The IPv4 prefix length of the network.",
```
joaopalet, line 172: Wrong field description here

stackit/internal/services/iaas/network/resource.go
```diff
@@ -141,20 +151,78 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+			"ipv6_gateway": schema.StringAttribute{
+				Description: "The IPv6 gateway of a network. If not specified, the first IP of the network will be assigned as the gateway.",
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.String{
+					validate.IP(),
+				},
+			},
+			"ipv6_nameservers": schema.ListAttribute{
+				Description: "The IPv6 prefix length of the network.",
```
joaopalet, line 205: Wrong field description here

stackit/internal/services/iaas/network/resource.go
```diff
@@ -167,6 +235,14 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				ElementType: types.StringType,
 				Optional:    true,
 			},
+			"routed": schema.BoolAttribute{
+				Description: "Shows if the network is routed and therefore accessible from other networks.",
```
joaopalet, line 239:
```suggestion
				Description: "If set to `true`, the network is routed and therefore accessible from other networks.",
```

joaopalet: Since the field can also be set by the user

stackit/internal/services/iaas/network/resource_test.go
joaopalet, line 1: We should also extend the acceptance test, replacing the usage of the deprecated fields with the new ones

stackit/internal/services/iaas/network/resource.go
```diff
@@ -450,59 +585,113 @@ func toCreatePayload(ctx context.Context, model *Model) (*iaas.CreateNetworkPayl
...
+	}
+
+	if !(model.IPv4Prefix.IsNull() || model.IPv4PrefixLength.IsNull() || model.IPv4Nameservers.IsNull()) {
+		addressFamily.Ipv4 = &iaas.CreateNetworkIPv4Body{
+			Nameservers:  &modelIPv4Nameservers,
+			Gateway:      iaas.NewNullableString(conversion.StringValueToPointer(model.IPv4Gateway)),
+			Prefix:       conversion.StringValueToPointer(model.IPv4Prefix),
+			PrefixLength: conversion.Int64ValueToPointer(model.IPv4PrefixLength),
+		}
 	}
```
joaopalet, line 624: We should continue supporting the `nameservers` field while it's deprecated. Currently this will be a breaking change for users using the `nameservers`

joaopalet: So we should use the `IPv4Nameservers` if provided, otherwise fallback to `nameservers`

joaopalet: And the same for the update

joaopalet: We can also a test case for this in the unit tests

stackit/internal/services/iaas/network/resource.go
GokceGK: In the API design, there are `two` seperate fields. 
Users can give a single `prefix` while creating a network. But in the response, there is a new field called `prefixes` where list of the prefixes are returned.

docs/resources/network.md
GokceGK: yes, I took the values from the api examples but we can change them.

stackit/internal/services/iaas/network/resource.go
GokceGK: Changed description

stackit/internal/services/iaas/network/resource.go
GokceGK: Changed description

stackit/internal/services/iaas/network/resource.go
GokceGK: In the API design, there are two seperate fields.
Users can give a single prefix while creating a network. But in the response, there is a new field called prefixes where list of the prefixes are returned. 

same like ipv4 :)

docs/resources/network.md
GokceGK: we have decided to remove the ipv6 example as it is not yet supported.

stackit/internal/services/iaas/network/resource.go
joaopalet: But for uniformity shouldn't we call them `IPv4Previx` and `IPv4Prefixes`? Like we do for the IPv6 ones

joaopalet: Instead of`IPv4Previx` and just `Prefixes`

stackit/internal/services/iaas/network/resource.go
```diff
@@ -248,6 +249,21 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+	if req.State.Raw.IsNull() {
+		// Planned to create a network resource
+		var plan Model
+		_ = req.Plan.Get(ctx, &plan)
+
+		if !plan.Nameservers.IsUnknown() && !plan.IPv4Nameservers.IsUnknown() {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Planned to create network resource", "You have provided nameservers and ipv4_nameservers field at the same time. Nameservers field has been deprecated. Please remove this field and try again.")
+		}
+	}
+}
```
joaopalet, line 265: Why do we need the `if req.State.Raw.IsNull()` here? We could just check the plan and see it both `Nameservers` and `IPv4Nameservers` were provided, and if yes throw the error

joaopalet: Then we wouldn't need the check in the `Update` function

stackit/internal/services/iaas/network/resource.go
GokceGK: @joaopalet yes we can do this. In that case, I need to deprecate `Prefixes` (even though it is only computed) field and create a new field `IPv4Prefixes`.

stackit/internal/services/iaas/network/resource.go
GokceGK: I tried this, but in that case, it throws error when we trigger delete operation also

stackit/internal/services/iaas/network/resource.go
GokceGK: Deprecated the `Prefixes` field and created new field `IPv4Prefixes`

stackit/internal/services/iaas/network/resource.go
GokceGK: I have changed the conditions to handle, create // update // delete at the same time. It works for all. 
Only if I have removed nameservers and ipv4_nameserver fields at the same time, it gives alert that I provided both fields. Needs to fix that

stackit/internal/services/iaas/network/resource.go
```diff
@@ -98,6 +110,18 @@ func (r *networkResource) Configure(ctx context.Context, req resource.ConfigureR
...
 
+func (r networkResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	var model Model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &model)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if !model.Nameservers.IsUnknown() && !model.IPv4Nameservers.IsUnknown() && !model.Nameservers.IsNull() && !model.IPv4Nameservers.IsNull() {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Planned to modify network resource", "You have provided nameservers and ipv4_nameservers field at the same time. Nameservers field has been deprecated. Please remove this field and try again.")
```
joaopalet, line 121:
```suggestion
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error configuring network", "You cannot provide both the `nameservers` and `ipv4_nameservers` field simultaneously. Please remove the deprecated `nameservers` field, and use `ipv4_nameservers` to configure IPv4 nameservers.")
```

stackit/internal/services/iaas/network/resource.go
```diff
@@ -141,20 +165,78 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+				},
+			},
 			"ipv4_prefix_length": schema.Int64Attribute{
 				Description: "The IPv4 prefix length of the network.",
 				Optional:    true,
 			},
 			"prefixes": schema.ListAttribute{
-				Description: "The prefixes of the network.",
+				Description:        "This field is deprecated and will be removed after April 28th 2025, use `ipv4_prefixes` to read the prefixes of the IPv4 networks.",
+				DeprecationMessage: "Use `ipv4_prefixes` to read the prefixes of the IPv4 networks.",
```
joaopalet, line 201:
```suggestion
			"prefixes": schema.ListAttribute{
				Description:        "The prefixes of the network. This field is deprecated and will be removed after April 28th 2025, use `ipv4_prefixes` to read the IPv4 prefixes of the network.",
				DeprecationMessage: "Use `ipv4_prefixes` to read the IPv4 prefixes of the network.",
```

stackit/internal/services/iaas/network/resource.go
```diff
@@ -450,59 +601,129 @@ func toCreatePayload(ctx context.Context, model *Model) (*iaas.CreateNetworkPayl
...
+
+	for _, ipv4ns := range modelIPv4List {
+		ipv4NameserverString, ok := ipv4ns.(types.String)
+		if !ok {
+			return nil, fmt.Errorf("type assertion failed")
+		}
+		modelIPv4Nameservers = append(modelIPv4Nameservers, ipv4NameserverString.ValueString())
+	}
+
+	if !model.IPv4Prefix.IsNull() || !model.IPv4PrefixLength.IsNull() || (!model.IPv4Nameservers.IsNull() && len(model.IPv4Nameservers.Elements()) > 0) || (!model.Nameservers.IsNull() && len(model.Nameservers.Elements()) > 0) {
```
joaopalet, line 641:
```suggestion
	if !model.IPv4Prefix.IsNull() || !model.IPv4PrefixLength.IsNull() || !model.IPv4Nameservers.IsNull() || !model.Nameservers.IsNull() {
```

joaopalet: I think this is not needed, otherwise we would not be able to set empty nameservers

stackit/internal/services/iaas/network/resource.go
```diff
@@ -450,59 +601,129 @@ func toCreatePayload(ctx context.Context, model *Model) (*iaas.CreateNetworkPayl
...
+	}
+	for _, ipv4ns := range modelIPv4List {
+		ipv4NameserverString, ok := ipv4ns.(types.String)
+		if !ok {
+			return nil, fmt.Errorf("type assertion failed")
+		}
+		modelIPv4Nameservers = append(modelIPv4Nameservers, ipv4NameserverString.ValueString())
+	}
+
+	if (!model.IPv4Nameservers.IsNull() && len(model.IPv4Nameservers.Elements()) > 0) || (!model.Nameservers.IsNull() && len(model.Nameservers.Elements()) > 0) {
```
joaopalet, line 706:
```suggestion
	if !model.IPv4Nameservers.IsNull() || !model.Nameservers.IsNull() {
```

joaopalet: Same here, otherwise updating to an empty list would fail with an inconsistent result after apply

stackit/internal/services/iaas/network/resource.go
joaopalet: Remember to also update the descriptions in the datasource to match these

examples/resources/stackit_network/resource.tf
```diff
@@ -1,9 +1,12 @@
...
-  nameservers        = ["1.2.3.4", "5.6.7.8"]
+  ipv4_nameservers   = ["1.2.3.4", "5.6.7.8"]
   ipv4_prefix_length = 24
+  ipv4_gateway       = "10.1.2.1"
+  ipv4_prefix        = "10.1.2.0/24"
   labels = {
     "key" = "value"
   }
-}
+  routed = false
```
joaopalet, line 11: Not sure if the example should set routed to false. We could confirm with IaaS

joaopalet: Also, it currently errors when creating a networking without specifying `routed` or when setting it to `true`. This should be possible as it's the default behaviour

stackit/internal/services/iaas/network/resource.go
GokceGK: updated descriptions in resource and datasource :)

examples/resources/stackit_network/resource.tf
GokceGK: as discussed I have implemented the `no_gateway` field checks

GokceGK: also implemented no_ipv6_gateway field :)

stackit/internal/services/iaas/network/resource.go
```diff
@@ -450,59 +632,149 @@ func toCreatePayload(ctx context.Context, model *Model) (*iaas.CreateNetworkPayl
...
+			Prefix:       conversion.StringValueToPointer(model.IPv4Prefix),
+			PrefixLength: conversion.Int64ValueToPointer(model.IPv4PrefixLength),
+		}
+	}
+
+	if model.NoIPv4Gateway.ValueBool() {
+		addressFamily.Ipv4.Gateway = iaas.NewNullableString(nil)
+	} else if !(model.IPv4Gateway.IsUnknown() || model.IPv4Gateway.IsNull()) {
+		addressFamily.Ipv4.Gateway = iaas.NewNullableString(conversion.StringValueToPointer(model.IPv4Gateway))
 	}
```
joaopalet, line 689: Won't this lead to a null pointer exception if the `if` above (line 677) is not entered? Because the `if` does not run, `addressFamily.Ipv4` will be null here, right?

joaopalet: For IPv6 this if inside the other `if`

joaopalet: Same for the update method

stackit/internal/services/iaas/network/resource.go
```diff
@@ -141,20 +182,92 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
 			"nameservers": schema.ListAttribute{
-				Description: "The nameservers of the network.",
+				Description:        "The nameservers of the network. This field is deprecated and will be removed soon, use `ipv4_nameservers` to configure the nameservers for IPv4.",
+				DeprecationMessage: "Use `ipv4_nameservers` to configure the nameservers for IPv4.",
+				Optional:           true,
+				Computed:           true,
+				ElementType:        types.StringType,
+			},
+			"no_ipv4_gateway": schema.BoolAttribute{
+				Description: "If set to `true`, the first IP of the network will be assigned as the IPv4 gateway.",
```
joaopalet, line 192: Shouldn't this be:
```suggestion
				Description: "If set to `true`, the network doesn't have a gateway.",
```

joaopalet: According to the API docs [here](https://docs.api.stackit.cloud/documentation/iaas/version/v1beta1#tag/Network/operation/v1beta1CreateNetwork)

joaopalet: Same for the IPv6 one

stackit/internal/services/iaas/network/resource.go
```diff
@@ -450,59 +632,149 @@ func toCreatePayload(ctx context.Context, model *Model) (*iaas.CreateNetworkPayl
...
+		Name:   conversion.StringValueToPointer(model.Name),
 		Labels: &labels,
-	}, nil
+	}
+
+	if addressFamily.Ipv6 != nil || addressFamily.Ipv4 != nil {
+		payload.AddressFamily = addressFamily
+	}
+
+	return &payload, nil
```
joaopalet, line 779: Why is routed never set in update payloads?

joaopalet: This leads to an inconsistent result after apply when we try to change it. If we don't want to allow updates in the `routed` field for some reason, we should make it `RequiresReplace`

GokceGK: Routed is not yet supported in update operations by API. That is why we have decided to remove it from the payloads. 

Thx for the hint, I added `REquiresReplace`

stackit/internal/services/iaas/network/resource.go
```diff
@@ -450,59 +632,149 @@ func toCreatePayload(ctx context.Context, model *Model) (*iaas.CreateNetworkPayl
...
+	for _, ipv6ns := range model.IPv6Nameservers.Elements() {
+		ipv6NameserverString, ok := ipv6ns.(types.String)
 		if !ok {
 			return nil, fmt.Errorf("type assertion failed")
 		}
-		modelNameservers = append(modelNameservers, nameserverString.ValueString())
+		modelIPv6Nameservers = append(modelIPv6Nameservers, ipv6NameserverString.ValueString())
+	}
+
+	if !model.IPv6Nameservers.IsNull() && len(model.IPv6Nameservers.Elements()) > 0 {
```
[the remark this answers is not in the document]
GokceGK: this is forgotten from the previous implementation, I have changed this condition to an unknown check.

stackit/internal/services/iaas/network/resource.go
GokceGK: It never caused a nil pointer exception during tests. 

But it might make sense to move in the above if block, cause gateway can only be set with prefix set when creating a network.

stackit/internal/services/iaas/network/resource.go
```diff
@@ -141,20 +181,89 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+				Optional:    true,
+				Validators: []validator.String{
+					validate.CIDR(),
+				},
+			},
 			"ipv4_prefix_length": schema.Int64Attribute{
 				Description: "The IPv4 prefix length of the network.",
 				Optional:    true,
 			},
 			"prefixes": schema.ListAttribute{
```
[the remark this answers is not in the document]
GokceGK: added `UseStateForUnknown`

joaopalet (approved):

PR #579, opened 2024-11-04
fix: Fix network interface handling of allowed addresses and security…

joaopalet:
… fields

**Fixes 2 issues in `network_interface` resource where:**
- The `security` field was not being mapped to the API payloads
- Mapping `allowed_addressed` from the API response back to the Terraform state sometimes caused an inconsistent result (the API returns `null` if the payload contains an empty list)

### DIFF
--- a/stackit/internal/services/iaas/networkinterface/resource.go
+++ b/stackit/internal/services/iaas/networkinterface/resource.go
@@ -562,20 +562,18 @@ func toCreatePayload(ctx context.Context, model *Model) (*iaas.CreateNICPayload,
 		}
 	}
 
-	var allowedAddressesPayload *[]iaas.AllowedAddressesInner
+	allowedAddressesPayload := &[]iaas.AllowedAddressesInner{}
 	if !(model.AllowedAddresses.IsNull() || model.AllowedAddresses.IsUnknown()) {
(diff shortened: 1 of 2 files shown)


TC-AVNP (approved):

stackit/internal/services/iaas/networkinterface/resource.go
```diff
@@ -553,19 +562,22 @@ func toCreatePayload(ctx context.Context, model *Model) (*iaas.CreateNICPayload,
 		}
 	}
 
-	allowedAddressesPayload := []iaas.AllowedAddressesInner{}
-
+	var allowedAddressesPayload *[]iaas.AllowedAddressesInner
 	if !(model.AllowedAddresses.IsNull() || model.AllowedAddresses.IsUnknown()) {
+		allowedAddresses := []iaas.AllowedAddressesInner{}
```
TC-AVNP, line 567: Why do you need to instantiate the variable again?
You already declared it in line 564

PhilippSeitz (approved):

TC-AVNP (approved):

TC-AVNP:
Code... LGMT

Could you improve the description of the PR? Thanks.

Issue #532, opened 2024-09-13
provide a way to migrate `stackit_argus_instance` to `stackit_observability_instance`


mouhsen-ibrahim:
The latest version of the provider renamed the resource `stackit_argus_instance` to `stackit_observability_instance`, if I change the resource type in my code and use the moved block to use the new type without re-creating the resources then I get this error from terraform

```
│ Error: Unsupported `moved` across resource types
│ 
│   on ../../modules/argus/argus.tf line 10:
│   10: moved {
│ 
│ The provider "registry.terraform.io/stackitcloud/stackit" does not support moved operations across resource types and providers.
```

My moved block looks like this

```
moved {
  from = stackit_argus_instance.api
  to = stackit_observability_instance.api
}
```

You must in the provider add support for migrating from `stackit_argus_instance` to `stackit_observability_instance`, here is the developer's documentation for doing this
https://developer.hashicorp.com/terraform/plugin/framework/resources/state-move

Can you please handle this ASAP so we can migrate to the new type?

joaopalet:
Hi Mouhsen,

Thanks for the suggestion, we will have a look and plan it it soon.
We will inform you here once it's released.

marceljk:
Hi @mouhsen-ibrahim,

We have now added the ability to move from argus to observability. We will release it with our next release.

PR #593, opened 2024-11-13
feat: Allow move state on new Observability resources

joaopalet:
**Draft PR to allow state move from old Argus to new Observability resources**

- Follows instructions the official [Terraform documentation](https://developer.hashicorp.com/terraform/plugin/framework/resources/state-move#statemover-with-sourceschema) but does not yet work
- Still gives error that provider "does not support moved operations across different resource types"
 
> [!TIP]
> **Workaround for users who want to migrate their Argus instance to Observability instance**
> 1. Create the new `stackit_observability_instance` in the config (should have same fields and values as the old Argus one)
> 2. Import the existing instance to the new `stackit_observability_instance` resource
>     - `$ terraform import stackit_observability_instance.example <PROJECT_ID>,<INSTANCE_ID>`
> 3. Delete the old Argus instance resource from the state
>     - `$ terraform state rm stackit_argus_instance.<INSTANCE_RESOURCE_NAME>`

Related to #532


marceljk:
Moving a argus resource to observability works. Descriptions of the argus resources extended where it's explained how to do it

bahkauv70 (dismissed):
Looks good to me, test worked flawlessly. One minor thing only: Would it make sense to add examples of the `move` blocks from the terraform files to the documentation of the deprecated argus resources?

marceljk:
> Looks good to me, test worked flawlessly. One minor thing only: Would it make sense to add examples of the `move` blocks from the terraform files to the documentation of the deprecated argus resources?

@bahkauv70 Yes, I think it's a good idea. I'll add some examples to the documentation

bahkauv70 (approved):

PR #594, opened 2024-11-14
improve resource documentation for STACKIT LoadBalancer Service

salvage7209:
Add a hint on the Access LoadBalancer Service behaviour to adjust security groups of its targets automatically. 
Also add a suggestion to circumvent this behaviour to not fight over security group ownership.


joaopalet:
Hi, thanks for the contribution but we are currently migrating the example to use STACKIT Terraform provider resources (instead of OpenStack) in #601

joaopalet:
So this PR is no longer necessary.

PR #596, opened 2024-11-20
Add missing fields to datasource

GokceGK:


mhenselin (approved):

docs/data-sources/network.md
```diff
@@ -43,8 +43,6 @@ data "stackit_network" "example" {
 - `labels` (Map of String) Labels are key-value string pairs which can be attached to a resource container
 - `name` (String) The name of the network.
 - `nameservers` (List of String, Deprecated) The nameservers of the network. This field is deprecated and will be removed soon, use `ipv4_nameservers` to configure the nameservers for IPv4.
-- `no_ipv4_gateway` (Boolean) If set to `true`, the network doesn't have an IPv4 gateway.
```
mhenselin: why do we remove these?

stackit/internal/services/iaas/network/datasource.go
```diff
@@ -112,10 +137,6 @@ func (d *networkDataSource) Schema(_ context.Context, _ datasource.SchemaRequest
 				Computed:           true,
 				ElementType:        types.StringType,
 			},
-			"no_ipv4_gateway": schema.BoolAttribute{
```
mhenselin: removing this is causing errors

stackit/internal/services/iaas/network/datasource.go
```diff
@@ -144,10 +165,6 @@ func (d *networkDataSource) Schema(_ context.Context, _ datasource.SchemaRequest
 				Computed:    true,
 				ElementType: types.StringType,
 			},
-			"no_ipv6_gateway": schema.BoolAttribute{
```
mhenselin: removing this causes errors

docs/data-sources/network.md
joaopalet: These fields are not needed in the datasource, as they are not actually used nor returned by the API. They are just used to match the API behaviour on the gateway field:
- `null` --> no gateway
- a value --> this is used for the gateway
- field not present --> default gateway

joaopalet (approved):

PR #604, opened 2024-11-26
Implement new virtual IP and virtual IP member resources

joaopalet:
> [!WARNING]
> Do not merge!
>
> Currently using `iaasalpha` SDK as the Virtual IP endpoints are not yet available in the `iaas` SDK.
> Once the the resources are tested and the endpoints are promoted, we can migrate to using the `iaas` SDK.

- Adds new resources
  - `stackit_virtual_ip`
  - `stackit_virtual_ip_member`


marceljk:
After talking with the IaaS colleagues, this feature will not be released, so this can be closed

PR #605, opened 2024-11-27
Fix public IP update always removing NIC association

joaopalet:
- Fixes an issue where `terraform apply`s detect drifts in the `stackit_public_ip` resource and removes existing network interface association when the user did not specify a `network_interface_id` in the config
- This makes the field `Computed` (besides `Optional`), and changes the update behaviour to not override an existing value if the field is Unknown, which makes it so that if the user does not specify the `network_interface_id` field in the config, Terraform would consider it a Computed field and would not attempt to update it


joaopalet:
Missing docs generation

vicentepinto98:
Aligned with the IaaS team: we won't do this for now. We will document that the user should set `ignore_lifecycle_changes` property to `network_interface_field` if the association is set outside terraform, e.g. when setting the public ip in a load balancer

Issue #602, opened 2024-11-26
SKE Cluster: Provider produced inconsistent result after apply


robertadesso1:
### Provider Version
v0.35.0

### Issue
The apply of ske cluster resource fails with the following error:

```text
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to module.ske_cluster.stackit_ske_cluster.cluster, provider "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new
│ value: .node_pools[1].taints[0].value: was null, but now cty.StringVal("").
```

### Description
We have two node pools, one of which uses taints. I increased the maximum node count in one node pool. Even though I only changed the node count, the provider seemed to confuse the node pools. The first node pool values suddenly got replaced with the second node pool's values, and the second node pool values suddenly got replaced with the first node pool's values. This results in the same valid, desired state but shows up as "replacing" node pool names, OS versions, etc., in the plan.

Since the planned outcome matched the desired state, I applied the plan. However, the apply fails with the given error.

Im not sure if the replacing of valid node pools is connected to this error. There were no changes to the taint itself. (except that the plan unecessarily replaced taints from nodepool A with B, and B with A as described). The hcl code of the taint did not change. 

Node pool configuration:
```hcl
  node_pools = [
    {
      name           = "reg-amd64"
      machine_type = "g1a.4d"     # 4 CPU, 16 GB RAM, amd64 (x86)
      os_version_min = "3975.2.1"
      minimum        = "1"
      maximum        = "2"
      max_surge      = "1"
      availability_zones = ["eu01-3"]
      volume_size    = 30
      volume_type    = "storage_premium_perf0"
      labels = {}
      taints = []
    },
    {
      name           = "licensed-amd64"
      machine_type = "g1a.8d"     # 8 CPU, 16 GB RAM, amd64 (x86)
      os_version_min = "3975.2.1"
      minimum        = "1"
      maximum        = "1"
      max_surge      = "1"
      availability_zones = ["eu01-3"]
      volume_size    = 30
      volume_type    = "storage_premium_perf0"
      labels = {
        "licensed-node" = true
      }
      taints = [{
        key = "licensed-node"
        effect = "NoSchedule"
      }]
    }
  ]
```

robertadesso1:
The bug seems to less involved than what i described in my previous comment.

We created a test cluster, first without any taints, this works fine.
```hcl
  node_pools = [
    {
[...]

PR #606, opened 2024-11-28
Fix inconsistent result after apply in SKE cluster node pool taint value

joaopalet:
- Fixes an inconsistent result after apply in SKE cluster node pool taint value
- Error occured when a `stackit_ske_cluster.nodepool[x].taints[x].value` was left empty in the config, since the API returns `""` in this case (instead of not providing the field in the response)
- Making the field Computed is enough to fix the issue

### DIFF
--- a/stackit/internal/services/ske/cluster/resource.go
+++ b/stackit/internal/services/ske/cluster/resource.go
@@ -485,6 +485,7 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 									"value": schema.StringAttribute{
 										Description: "Taint value corresponding to the taint key.",
 										Optional:    true,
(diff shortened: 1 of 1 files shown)


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -485,6 +485,7 @@ func (r *clusterResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 									"value": schema.StringAttribute{
 										Description: "Taint value corresponding to the taint key.",
 										Optional:    true,
+										Computed:    true,
```
vicentepinto98, line 488: We should also add `UseStateForUnknown`

joaopalet:
Relates to #602

Issue #520, opened 2024-08-28
STACKIT Service Account Email


BackInBash:
Hi,

why do I need to specify the Service Account Email Adresse in combination with the Access Token?

```console
╷
│ Error: Error creating project
│ 
│   with stackit_resourcemanager_project.project-hub,
│   on 03-project.tf line 9, in resource "stackit_resourcemanager_project" "project-hub":
│    9: resource "stackit_resourcemanager_project" "project-hub" {
│ 
│ The service account e-mail cannot be empty: set it in the provider configuration or through the STACKIT_SERVICE_ACCOUNT_EMAIL or in your credentials file (default filepath is ~/.stackit/credentials.json)
```

Because the Service Account Email Adress is part of the JWT Token itself and could be extracted 🤔 

Example Terraform Code to extract Service Account Email:
```hcl
jsondecode(base64decode(split(".", var.STACKIT_SERVICE_ACCOUNT_TOKEN)[1])).email
```

vicentepinto98:
Hi @BackInBash , thanks for opening the issue.
You are right, we could extract it from the JWT Token. We will also check if we need it at all, as it might not be required anymore. 
If that's the case, we'll remove this check.

PR #610, opened 2024-12-12
Deprecate service_account_email config

marceljk:
Deprecated service_account_email config, because it is not required and could be also extracted from the JWT if needed

### DIFF
--- a/stackit/provider.go
+++ b/stackit/provider.go
@@ -94,7 +94,7 @@ func (p *Provider) Metadata(_ context.Context, _ provider.MetadataRequest, resp 
 
 type providerModel struct {
 	CredentialsFilePath             types.String `tfsdk:"credentials_path"`
-	ServiceAccountEmail             types.String `tfsdk:"service_account_email"`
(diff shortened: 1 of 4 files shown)


stackit/provider.go
```diff
@@ -167,8 +167,9 @@ func (p *Provider) Schema(_ context.Context, _ provider.SchemaRequest, resp *pro
 				Description: descriptions["credentials_path"],
 			},
 			"service_account_email": schema.StringAttribute{
-				Optional:    true,
-				Description: descriptions["service_account_email"],
+				Optional:           true,
+				Description:        descriptions["service_account_email"],
+				DeprecationMessage: "service_account_email has been deprecated because it is not required. Will be removed after June 12th 2025.",
```
joaopalet, line 172:
```suggestion
				DeprecationMessage: "The `service_account_email` field has been deprecated because it is not required. Will be removed after June 12th 2025.",
```

joaopalet (approved):

marceljk:
Addresses #520

PR #615, opened 2024-12-17
chore: provide terraform binary to pipeline

bahkauv70:


.github/workflows/ci.yaml
```diff
@@ -4,6 +4,7 @@ on: [pull_request, workflow_dispatch]
 
 env:
   GO_VERSION: "1.23"
+  TERRAFORM_VERSION: "v1.10.2"
```
vicentepinto98, line 7: is there a way to always use the latest one? Would be better

bahkauv70: Yes, the version can just be omitted and then the action will always pull the latest. I'll remove it for now. However, I could imagine, that in the future we need a matrix action here to assure that the provider works with various, explicitly named, terraform actions...

vicentepinto98 (approved):

PR #623, opened 2025-01-08
feat: Allow managing server state in stackit_server resource

bahkauv70:


stackit/internal/services/iaas/server/resource.go
```diff
@@ -450,17 +645,42 @@ func (r *serverResource) Update(ctx context.Context, req resource.UpdateRequest,
...
 		return
 	}
-	// Update existing server
-	updatedServer, err := r.client.UpdateServer(ctx, projectId, serverId).UpdateServerPayload(*payload).Execute()
-	if err != nil {
-		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating server", fmt.Sprintf("Calling API: %v", err))
-		return
+
+	var updatedServer *iaas.Server
+	if model.DesiredStatus.ValueString() != modelStateDeallocated {
```
marceljk, line 670: If a user would change multiple fields of a server, then it could lead here to an invalid state. Because if the user sets the state of the server to deallocated and change some other values, then it would only change the state of the server but not the other values. So maybe you can do the check before updating the server status

stackit/internal/services/iaas/server/resource.go
```diff
@@ -389,6 +451,139 @@ func (r *serverResource) Create(ctx context.Context, req resource.CreateRequest,
...
+	}
+	server, err := wait.StopServerWaitHandler(ctx, client, projectId, serverId).WaitWithContext(ctx)
+	if err != nil {
+		core.LogAndAddError(ctx, diagnostics, "Error updating the server", fmt.Sprintf("cannot check stopped server: %v", err))
+		return nil
+	}
+	return server.Status
+}
+
+func deallocatServer(ctx context.Context, client serverControlClient, projectId, serverId string, diagnostics *diag.Diagnostics) *string {
```
marceljk, line 491: Typo in function name. deallocateServer

docs/resources/server.md
marceljk, line 1: The underscores in resource names and variables have been removed, so the examples here are wrong.

docs/resources/resourcemanager_project.md
marceljk, line 1: Remove these changes, these values should not be public.

docs/data-sources/resourcemanager_project.md
marceljk, line 1: Remove these changes as well. They shouldn't be public as well.

docs/resources/key_pair.md
marceljk, line 1: The underscores here also disappeared, so the examples are not correct.

stackit/internal/services/iaas/server/resource.go
```diff
@@ -333,10 +341,59 @@ func (r *serverResource) Schema(_ context.Context, _ resource.SchemaRequest, res
 				Description: "Date-time when the server was updated",
 				Computed:    true,
 			},
+			"desired_status": schema.StringAttribute{
+				Description: "The desired status of the server resource. Defaults to 'active' " + utils.SupportedValuesDocumentation(desiredStatusOptions),
+				Optional:    true,
+				Computed:    true,
```
vicentepinto98, line 347: I don't think this field should be computed, since we are not mapping the field to any value that is returned by the API

stackit/internal/services/iaas/server/resource.go
```diff
@@ -450,17 +645,42 @@ func (r *serverResource) Update(ctx context.Context, req resource.UpdateRequest,
...
+	var updatedServer *iaas.Server
+	if model.DesiredStatus.ValueString() != modelStateDeallocated {
+		// Update existing server
+		updatedServer, err = r.client.UpdateServer(ctx, projectId, serverId).UpdateServerPayload(*payload).Execute()
+		if err != nil {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating server", fmt.Sprintf("Calling API: %v", err))
+			return
+		}
+	} else {
+		// we cannot update the metadata of a shelved server, read the server state again to update the model
```
vicentepinto98, line 678: I'm not sure if this is the best approach, it will lead to errors if the state doesn't match the user configuration right?

We could start the server, update the metadata and then deallocate it, but what to do in this case should be discussed with the IaaS team

stackit/internal/services/iaas/server/resource.go
```diff
@@ -605,7 +826,15 @@ func mapFields(ctx context.Context, serverResp *iaas.Server, model *Model) error
 
 	model.ServerId = types.StringValue(serverId)
 	model.MachineType = types.StringPointerValue(serverResp.MachineType)
-	model.AvailabilityZone = types.StringPointerValue(serverResp.AvailabilityZone)
+
+	// Proposed fix: If the server is deallocated, it has no availability zone anymore
```
vicentepinto98, line 830: Also something to check with the IaaS team. I think this fix makes sense, but probably what should happen instead is that the API would return the availability zone for deallocated servers. Then we wouldn't need it

docs/data-sources/resourcemanager_project.md
```diff
@@ -3,7 +3,7 @@
 page_title: "stackit_resourcemanager_project Data Source - stackit"
```
vicentepinto98, line 3: this file shouldn't be changed, we are leaving the members field out of the documentation for now

docs/resources/key_pair.md
```diff
@@ -5,26 +5,24 @@ subcategory: ""
 description: |-
   Key pair resource schema. Must have a region specified in the provider configuration. Allows uploading an SSH public key to be used for server authentication.
   Usage with server
-  
-  resource "stackit_key_pair" "keypair" {
+  ```terraform
```
vicentepinto98, line 8: this change also seems wrong

docs/resources/resourcemanager_project.md
```diff
@@ -35,9 +35,20 @@ resource "stackit_resourcemanager_project" "example" {
 ### Optional
```
vicentepinto98, line 35: this file shouldn't be changed, we are leaving the members field out of the documentation for now

docs/data-sources/resourcemanager_project.md
bahkauv70: I'll remove the change from the "other" files and only keep the the ones related to the server resource

docs/resources/key_pair.md
bahkauv70: removed the generated changes

docs/resources/resourcemanager_project.md
bahkauv70: removed the generated changes

docs/resources/server.md
bahkauv70: This was done by the doc generator. I'll have to investigate why this happened.

stackit/internal/services/iaas/server/resource.go
bahkauv70: This relates to [this](https://github.com/stackitcloud/terraform-provider-stackit/pull/623#pullrequestreview-2537421590) comment. The problem here is, that the server config cannot be updated while it is shelved (this will cause a 409). The chek for the state intends to avoid that. The situation is non-trivial, as I found during testing. e.g. if the user has a shelved server and wants to reactivate it and apply changes. So we have to start the server first and afterwards update it. But if we want to deactivate it _and_ apply changes as well, the order has to be reversed (first update and then shelve). So this actually target state specific. The correct approach would probably be:
* check, if desired state is _deallocate_. If so, apply all other changes and update the state afterwards
* if not, apply the state change first (this will block until the server is responsive again). Afterwards apply the "other" attribute changes.

bahkauv70: I will adapt the code according to this proposal and test it

stackit/internal/services/iaas/server/resource.go
bahkauv70: This is the same issue as mentioned [here](https://github.com/stackitcloud/terraform-provider-stackit/pull/623#discussion_r1907443563). I'll adapt the code as proposed below

stackit/internal/services/iaas/server/resource.go
bahkauv70: I'll check back with the Iaas team. Could actually be a bug, because the zone should still be available in the shelved containers configuration (but it is not sent out via the API).

stackit/internal/services/iaas/server/resource.go
vicentepinto98: I agree with the approach, plus like you said, if the server is already in a deallocated state, we need to first start it, apply the changes and deallocate it again, but this should be checked with IaaS

stackit/internal/services/iaas/server/resource.go
vicentepinto98: Ah I missed line 503. With the current implementation it must be computed yes, but I'm wondering if it make sense. I find it confusing that a field which is called "desired_status" would be computed. Maybe it would make more sense to **not** save the value that the API returns for the status in the terraform state. 

Actually this made me realize that you are not using the waiters, see [https://github.com/stackitcloud/stackit-sdk-go/blob/main/services/iaas/wait/wait.go#L315](wait.go). If you replace the current GET calls by the waiters, we are sure that when they succeed the state was correctly updated and don't need to save that value in the state. 

Then the field doesn't need to be computed, but you still need to take care of the case where the field is removed. I would just set it to whatever the user configured, which in this case would be null

stackit/internal/services/iaas/server/resource.go
bahkauv70: The approach described above required a little refactoring, but now this situation is handled correctly

docs/resources/server.md
bahkauv70: reverted the accidental and broken changes

stackit/internal/services/iaas/server/resource.go
bahkauv70: We should discuss in a short meeting, I'm not quite clear to handle the attribute.
The waiters have been added

stackit/internal/services/iaas/server/resource.go
bahkauv70: The IaaS team has confirmed, that a shelved server does not have an availability zone, so we have to keep the fix

stackit/internal/services/iaas/server/resource.go
bahkauv70: Reconfigured the desired_state as a non-computed field, making it effectively a write-only attribute. The incoming value is now stored in the state, not the actual server state. This simplifies the handling of the state considerably, makes the semantics more clear and fixes some problems

bahkauv70:
Regarding the waiters: The retrieval of the server state is the only situation where a waiter is _not_ used. However, function `waiter.CreateServerWaitHandler` cannot be used, as it explicitly blocks until the server state becomes *ACTIVE*. By principle this won't not work for us, as the provider must be able to retrieve the server attributes in any state

stackit/internal/services/iaas/server/resource.go
```diff
@@ -333,10 +341,58 @@ func (r *serverResource) Schema(_ context.Context, _ resource.SchemaRequest, res
 				Description: "Date-time when the server was updated",
 				Computed:    true,
 			},
+			"desired_status": schema.StringAttribute{
+				Description: "The desired status of the server resource. Defaults to 'active' " + utils.SupportedValuesDocumentation(desiredStatusOptions),
```
vicentepinto98, line 345:
```suggestion
				Description: "The desired status of the server resource. " + utils.SupportedValuesDocumentation(desiredStatusOptions),
```

vicentepinto98 (approved):

vicentepinto98 (approved):

marceljk (approved):

PR #629, opened 2025-01-15
fix(postgresflex): continue if database is not found during read

hown3d:
# Context

When the [`Read`](https://github.com/stackitcloud/terraform-provider-stackit/blame/f04ced99815083eccba802830a5f6a4c1b0671c3/stackit/internal/services/postgresflex/database/resource.go#L220) function checks via the API to see if a database exists [here](https://github.com/stackitcloud/terraform-provider-stackit/blame/f04ced99815083eccba802830a5f6a4c1b0671c3/stackit/internal/services/postgresflex/database/resource.go#L234-L243) and the database does not exist yet, the `getDatabase` function returns an error, resulting in terraform returning an error. 
 
I encountered the issue when using the following command with our provider
```sh
$ terraform apply -refresh-only -auto-approve -input=false -lock=false -json
...
error reading database: Calling API: database not found
```

### DIFF
--- a/stackit/internal/services/postgresflex/database/resource.go
+++ b/stackit/internal/services/postgresflex/database/resource.go
@@ -2,6 +2,7 @@ package postgresflex
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


Fyusel (approved):
Thank you for submitting this PR.

Since it was not easy reproducible on my side I tested what happens when getDatabase() directly returns the error. In this case I can see the described behavior and with the fix the resource is cleaned up.

PR #639, opened 2025-01-22
fix: add missing status attribute to schema

bahkauv70:


stackit/internal/services/iaas/server/datasource.go
```diff
@@ -180,6 +180,10 @@ func (r *serverDataSource) Schema(_ context.Context, _ datasource.SchemaRequest,
 				Description: "Date-time when the server was updated",
 				Computed:    true,
 			},
+			"desired_status": schema.StringAttribute{
+				Description: "The desired status of the server resource." + utils.SupportedValuesDocumentation(desiredStatusOptions),
+				Optional:    true,
```
marceljk, line 185:
```suggestion
				Computed:    true,
```

bahkauv70: The semantics of "computed" are not 100% applicable in this case, as the attribute is actually not read from the serverside (it is a "write-only" attribute). But the "computed" setting probably won't hurt and it would be consistent with the other attributes. I'll change it and test it.

marceljk (approved):

PR #640, opened 2025-01-22
fix: separated models for resource and data-source

bahkauv70:
separated the models of resource and datasource to handle individual attributes correctly

### DIFF
--- a/stackit/internal/services/iaas/server/datasource.go
+++ b/stackit/internal/services/iaas/server/datasource.go
@@ -4,11 +4,15 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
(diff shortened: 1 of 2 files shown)


stackit/internal/services/iaas/server/datasource.go
```diff
@@ -233,3 +252,86 @@ func (r *serverDataSource) Read(ctx context.Context, req datasource.ReadRequest,
 	}
 	tflog.Info(ctx, "server read")
 }
+
+func mapDataSourceFields(ctx context.Context, serverResp *iaas.Server, model *DataSourceModel) error {
```
joaopalet, line 256: Just missing unit test for this, otherwise looks good to me

bahkauv70: Added a testcase based on the corresponding one from the resource

joaopalet (approved):

marceljk (approved):

PR #650, opened 2025-01-27
fix: exclude zero address as for inadmissable cases

bahkauv70:


stackit/internal/validate/validate_test.go
```diff
@@ -58,39 +58,87 @@ func TestUUID(t *testing.T) {
...
+			true,
+			"::",
+			true,
+		},
+		{
+			"valid ipv6 zero short notation",
+			true,
+			"::",
+			true,
+		},
```
Fyusel, line 118: Those two tests are the same.

bahkauv70: Should have been a different syntax, fixed

Fyusel (approved):

PR #651, opened 2025-01-27
Allow deleting root volume when deleting server

Fyusel:
Introduce delete_on_termination field.


stackit/internal/services/iaas/server/datasource.go
```diff
@@ -168,6 +168,10 @@ func (r *serverDataSource) Schema(_ context.Context, _ datasource.SchemaRequest,
 						Description: "The ID of the source, either image ID or volume ID",
 						Computed:    true,
 					},
+					"delete_on_termination": schema.BoolAttribute{
```
marceljk, line 171: The whole object "boot_volume" doesn't get set in the `mapDataSourceFields()`function. This needs to be extended

docs/resources/server.md
```diff
@@ -384,7 +384,7 @@ resource "stackit_server" "user-data-from-file" {
 - `affinity_group` (String) The affinity group the server is assigned to.
 - `availability_zone` (String) The availability zone of the server.
 - `boot_volume` (Attributes) The boot volume for the server (see [below for nested schema](#nestedatt--boot_volume))
-- `desired_status` (String) The desired status of the server resource. Defaults to 'active' Supported values are: `active`, `inactive`, `deallocated`.
+- `desired_status` (String) The desired status of the server resource.Supported values are: `active`, `inactive`, `deallocated`.
```
marceljk, line 387: Missing white space after the dot.
```suggestion
- `desired_status` (String) The desired status of the server resource. Supported values are: `active`, `inactive`, `deallocated`.
```

stackit/internal/services/iaas/server/datasource.go
Fyusel: I created a bug for this since we don't get the corresponding values from the server right now. This needs to be fixed first.

marceljk (approved):

PR #653, opened 2025-01-27
fix: Updating values of "stackit_server" leads to an inconsistent result

marceljk:
If you set "network_interfaces" in "stackit_server" and then apply an update to the server (e.g. if the name has changed), you will get an inconsistent result.


bahkauv70 (approved):
Looks good, tested it as well, works as expected

PR #659, opened 2025-01-30
fix: kubeconfig state will not be removed if the cluster does not exists anymore

marceljk:


stackit/internal/services/ske/kubeconfig/resource.go
```diff
@@ -259,6 +261,11 @@ func (r *kubeconfigResource) Read(ctx context.Context, req resource.ReadRequest,
 
 	cluster, err := r.client.GetClusterExecute(ctx, projectId, clusterName)
 	if err != nil {
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
```
bahkauv70, line 264: Looks good to me. Just for my clarification: I saw this pattern occur at other places as well. Is there a specific reason to use this approach, as errors.As(...) will find unwrapped errors as well

bahkauv70 (approved):

Issue #435, opened 2024-06-26
Support for IAM


gfelbing:
We are using the provider to create our [stackit projects](https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/resources/resourcemanager_project).
But the resources only allows us to set the project owner, not the actual access rights to the projects, but has to be maintained by hand.

I see two options fixing that:

1. Extend `stackit_resourcemanager_project` to maintain the list of members (as far as supported by the [api](https://docs.api.eu01.stackit.cloud/documentation/resource-manager/version/unversioned#tag/Projects))
2. Add a resource to support the [membership api](https://docs.api.eu01.stackit.cloud/documentation/authorization/version/v2)

Or obvious 3rd option: both.

joaopalet:
Hi @gfelbing,

Thanks for opening the issue. As far as I could see providing a list of members is supported in the memberships API (`members` field of the [`"create project" endpoint`](https://docs.api.eu01.stackit.cloud/documentation/resource-manager/version/unversioned#tag/Projects/operation/create-project)). Although the list cannot be updated after creation (it's not present in the ["update project" endpoint](https://docs.api.eu01.stackit.cloud/documentation/resource-manager/version/unversioned#tag/Projects/operation/update-project))

So we will plan the integration of the `members` field in the project resource, and will look into what funcionality from the membership API can be onboarded in separate resources

joaopalet:
Hi @gfelbing,

We've added support for `members` in the project resource in version [`v0.26.1`](https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/resources/resourcemanager_project#members). You can now specify a list of `members`, each consisting in a `subject` and `role`, that will be added to the project.

We have deprecated the existing `owner_email` field, but it will still work to set the user as a project owner for backwards compatibility. The new `members` field will take precedence and be used if it's specified in the config.

GokceGK:
Hi @gfelbing,

after discussions with the product team, we agreed that managing IAM resources with TFP is not currently possible and may be supported in the future. Therefore, we need to remove the members field from the stackit_resourcemanager_project resource.

In order not to completely break the TFP settings, we will mark members field as deprecated for a while. Any changes made to the field will not affect existing projects. We strongly recommend against using the members field.

The owner_email field will remain in use so that a user can be added to the resource at project creation.

We apologize for any inconvenience.

PR #665, opened 2025-02-04
IAM Role Assignment

l0wl3vel:
Reattempts implementing Role Assignment as a separate resource instead of a field in an existing resource, which lead to problems with inherited group membership and general composability.

This attempt should mesh better with the internal behaviour of STACKIT API, which creates an anonymous "association object" when assigning roles: https://docs.stackit.cloud/stackit/en/identity-and-access-management-134415813.html

Currently only implements roles for STACKIT Projects and Organizations.

Fixes: #435


stackit/internal/services/resourcemanager/projectroleassignment/resource_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	stackit_sdk_config "github.com/stackitcloud/stackit-sdk-go/core/config"
+	"github.com/stackitcloud/stackit-sdk-go/services/authorization"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/testutil"
+)
+
+
+func TestProjectRoleAssignmentResource(t *testing.T)	{
```
Fyusel, line 19: We keep our acceptance tests in a separate file (in this case resourcemanager_acc_test.go). Could you move this test please and rename it to TestAccPro...?

Fyusel (commented):
Can you add also an example terraform config file please (under /examples folder)?
And you can run `make generate-docs` to generate the documentation

l0wl3vel:
> Can you add also an example terraform config file please (under /examples folder)? And you can run `make generate-docs` to generate the documentation

Sure. Just FYI: The Docs generated have a diff outside of the areas I touched.

stackit/internal/services/resourcemanager/projectroleassignment/resource_test.go
l0wl3vel: moved the projectroleassignment resource into its own authorization subgroup and moved the tests into a acceptance test file.

stackit/internal/features/experiments.go
```diff
@@ -0,0 +1,58 @@
...
+// Check if an experiment is enabled.
+func CheckExperimentEnabled(ctx context.Context, data *core.ProviderData, experiment, resourceType string, diags *diag.Diagnostics) {
+	if !ValidExperiment(experiment, diags) {
+		return
+	}
+	experiment_active := slices.ContainsFunc(data.Experiments, func(e string) bool {
+		return strings.EqualFold(e, experiment)
+	})
+
+	if !experiment_active {
```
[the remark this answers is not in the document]
rubenhoenle: Also when testing this manually, I get the a warning instead of an error when I don't have `iam` experiment enabled, but use one of your new resources. I expect an error in that case. 

Please check this carefully.

rubenhoenle (changes requested):
see above

stackit/internal/features/experiments.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+var Available_experiments []string = []string{"iam"}
+
+// Check if an experiment is valid.
+func ValidExperiment(experiment string, diags *diag.Diagnostics) bool {
+	valid_experiment := slices.ContainsFunc(Available_experiments, func(e string) bool {
+		return strings.EqualFold(e, experiment)
+	})
+	if !valid_experiment {
```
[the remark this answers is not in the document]
rubenhoenle: Unresolving this because snake case is still used here instead of camel case. Any reason for this?

stackit/internal/features/experiments.go
```diff
@@ -0,0 +1,58 @@
...
+
+	return valid_experiment
+}
+
+// Check if an experiment is enabled.
+func CheckExperimentEnabled(ctx context.Context, data *core.ProviderData, experiment, resourceType string, diags *diag.Diagnostics) {
+	if !ValidExperiment(experiment, diags) {
+		return
+	}
+	experiment_active := slices.ContainsFunc(data.Experiments, func(e string) bool {
```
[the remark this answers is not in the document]
rubenhoenle: Unresolving this because snake case is still used here instead of camel case. Any reason for this?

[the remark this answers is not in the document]
l0wl3vel: Have overlooked it. Your suggestions were not including all the sites the variables appeared. So I decided to redo the renaming with ide support and overlooked this one. Fixup is coming up in a sec

stackit/internal/services/authorization/authorization_acc_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	stackitSdkConfig "github.com/stackitcloud/stackit-sdk-go/core/config"
+	"github.com/stackitcloud/stackit-sdk-go/services/authorization"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/testutil"
+)
+
+//go:embed testfiles/prerequisites.tf
+var prerequisites string
```
rubenhoenle, line 22: One additional thing about codestyle: We normally handle our acceptance test resources like this: https://github.com/stackitcloud/terraform-provider-stackit/blob/main/stackit/internal/services/mariadb/mariadb_acc_test.go#L54-L76

Maybe we can also do that here (instead of the file embeddings) to stay consistent across the codebase of our provider.

l0wl3vel:
Sorry did not intend to request a review yet. My bad

stackit/internal/services/authorization/authorization_acc_test.go
l0wl3vel: I had it like this in the beginning but after adding more resources I thought it would be better to have the resource configuration in separate .tf files for readability. But sure, we can do that.

rubenhoenle: You're right. We will also have syntax hightlighting in IDEs then. Let's keep it like it is 👍

rubenhoenle:
> It smells like caching the result of the experiment check does not pass the assumption that can be reused between configure calls. Since the experiment check is derived from the beta resources code, it is also likely happening there. I will try to just remove the caching of the experiment check, since caching the check seemed unnecessary to me in the first place.

@l0wl3vel I will check this for the beta resources and create a bug ticket for it. Since you have removed the caching for experiments, it's all fine from my side.

rubenhoenle (approved):

PR #667, opened 2025-02-05
fix: handle expiration date in regard to changed timezones

bahkauv70:


stackit/internal/services/objectstorage/credential/resource.go
```diff
@@ -52,6 +54,34 @@ type credentialResource struct {
...
+	resp.Diagnostics.Append(utils.GetTimeFromStringAttribute(ctx, p, req.State, time.RFC3339, &stateDate)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	resp.Diagnostics.Append(utils.GetTimeFromStringAttribute(ctx, p, resp.Plan, time.RFC3339, &planDate)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// replace the planned expiration time with the current state date, iff they represent
```
Fyusel, line 74: nitpick: 
```suggestion
	// replace the planned expiration time with the current state date, if they represent
```

stackit/internal/services/objectstorage/credential/resource.go
```diff
@@ -218,6 +249,25 @@ func (r *credentialResource) Create(ctx context.Context, req resource.CreateRequ
...
+	)
+	resp.Diagnostics.Append(utils.ToTime(ctx, time.RFC3339, model.ExpirationTimestamp, &actualDate)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	resp.Diagnostics.Append(utils.GetTimeFromStringAttribute(ctx, path.Root("expiration_timestamp"), req.Plan, time.RFC3339, &planDate)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// replace the planned expiration date with the original date, iff
```
Fyusel, line 265:
```suggestion
	// replace the planned expiration date with the original date, if
```

stackit/internal/services/objectstorage/credential/resource.go
```diff
@@ -251,6 +301,26 @@ func (r *credentialResource) Read(ctx context.Context, req resource.ReadRequest,
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(utils.GetTimeFromStringAttribute(ctx, path.Root("expiration_timestamp"), req.State, time.RFC3339, &stateDate)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// replace the resulting expiration date with the original date, iff
```
Fyusel, line 319:
```suggestion
	// replace the resulting expiration date with the original date, if
```

Fyusel (approved):

stackit/internal/services/objectstorage/credential/resource.go
bahkauv70: This is actually correct: iff == If, and only if (https://dict.leo.org/englisch-deutsch/iff) ;-)

stackit/internal/services/objectstorage/credential/resource.go
bahkauv70: s. above: https://dict.leo.org/englisch-deutsch/iff ;-)

stackit/internal/services/objectstorage/credential/resource.go
bahkauv70: see above

bahkauv70:
Additional clarification: If the expiration_time is updated to the same date, but with a different notation, e.g. 2025-12-31T12:00:00+01:00 -> 2025-12-31T11:00:00Z, terraform will issue an update anyway, because:
* The plan is not empty (it still contains the no-op change to expiration_time)
* The computed attributes have to be updated
There doesn't seem to be a officaly way to remove attributes from the plan explicitly. So an empty update is performed, but it won't change the state

PR #671, opened 2025-02-07
fix: use file inputstream on file upload instead of in-memory buffer

rubenhoenle:
relates to STACKITTPR-70

---

Not sure how this could be unit tested as we would need some file dummy to upload. Would be thankful for some hints on that 😄


rubenhoenle:
Added testing instructions to STACKITTPR-70, thanks @bahkauv70 for the hint.

PR #672, opened 2025-02-07
Feat: add egress_address_ranges attribute to ske resource

h3adex:
Added the new attribute `egress_address_ranges` from the ske api to the terraform provider ressource.

I've adjusted the tests and run the commands:
make fmt
make test

I've also manually tested the changes by building the go binary and creating a new ske cluster. I've also checked that a user is able to downgrade/upgrade tf provider versions (back and forth from v.0.40 to this version and checked the state manually).

Thank you :-)


Fyusel (commented):
Thank you for contributing. Looks good!
Can you run make generate-docs as well please and add the updated ske documentation?

h3adex:
Added the documentation. I've also removed the optional flag, since this attribute is not optional on the API response. Every Cluster will return some sort of value for the egress ips.

Fyusel (approved):

PR #674, opened 2025-02-10
feat: include preview versions as fallback

bahkauv70:


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1865,58 +1897,105 @@ func latestMatchingKubernetesVersion(availableVersions []ske.KubernetesVersion,
...
+		availableVersionsArray = append(availableVersionsArray, *candidateVersion.Version)
+		vPreffixed := "v" + *candidateVersion.Version
+
+		// [MAJOR].[MINOR] version provided, get the latest non-preview patch version
+		if semver.MajorMinor(vPreffixed) == semver.MajorMinor(kubernetesVersionMin) &&
+			(semver.Compare(vPreffixed, kubernetesVersionMin) >= 0) &&
+			(candidateVersion.State != nil) {
+			// take the current version as a candidate, if we have no other version inspected before
+			// OR the previously found version was a preview version
+			if selectedVersion == nil {
```
rubenhoenle, line 1965:
```suggestion
			if (selectedVersion == nil || (isSupported(&candidateVersion) && isPreview(selectedVersion))) {
```

The code inside the `ìf` and the `else if` block is exactly the same, I would prefer to have it combined into one

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1865,58 +1897,105 @@ func latestMatchingKubernetesVersion(availableVersions []ske.KubernetesVersion,
...
+		availableVersionsArray []string
+	)
+	if fullVersion {
+		availableVersionsArray, selectedVersion = selectFullVersion(availableVersions, providedVersionPrefixed)
+	} else {
+		availableVersionsArray, selectedVersion = selectMatchingVersion(availableVersions, providedVersionPrefixed)
+	}
+
+	if selectedVersion != nil {
+		deprecated = strings.EqualFold(*selectedVersion.State, VersionStateDeprecated)
```
rubenhoenle, line 1919:
```go
deprecated = strings.EqualFold(*selectedVersion.State, VersionStateDeprecated)
```

Maybe a newbie question, why do you to `strings.EqualFold(...)` here and in line 1998 you do it like this?

```go
return *v.State == VersionStateSupported
```

Apart from the fact that there is `isPreview()` and `isSupported()`, but not `isDeprecated()`

stackit/internal/services/ske/cluster/resource.go
bahkauv70: Yes, of course. This was a combination of old and new code, which didn't make sense. Applied the suggestion

stackit/internal/services/ske/cluster/resource.go
bahkauv70: Additionally I've introduced a IsDeprecated helper function as well and replaced the checks consistently with the invocation of the helper functions. This makes the code considerably clearer

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1865,58 +1897,115 @@ func latestMatchingKubernetesVersion(availableVersions []ske.KubernetesVersion,
...
+		selectedVersion        *ske.KubernetesVersion
+		availableVersionsArray []string
+	)
+	if fullVersion {
+		availableVersionsArray, selectedVersion = selectFullVersion(availableVersions, providedVersionPrefixed)
+	} else {
+		availableVersionsArray, selectedVersion = selectMatchingVersion(availableVersions, providedVersionPrefixed)
+	}
+
+	if selectedVersion != nil {
```
rubenhoenle, line 1918: I think the nil check is redundant now, nil check is done by `isDeprecated` / `isPreview` now.

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1865,58 +1897,115 @@ func latestMatchingKubernetesVersion(availableVersions []ske.KubernetesVersion,
...
+		deprecated = isDeprecated(selectedVersion)
```
rubenhoenle, line 1919:
```suggestion
		deprecated := isDeprecated(selectedVersion)
```

stackit/internal/services/ske/cluster/resource.go
bahkauv70: Agree

stackit/internal/services/ske/cluster/resource.go
bahkauv70: Applied

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -743,6 +744,37 @@ func (r *clusterResource) Create(ctx context.Context, req resource.CreateRequest
 	tflog.Info(ctx, "SKE cluster created")
 }
 
+func sortK8sVersions(versions []ske.KubernetesVersion) {
+	sort.Slice(versions, func(i, j int) bool {
+		v1, v2 := (versions)[i].Version, (versions)[j].Version
+		if v1 == nil && v2 == nil {
```
rubenhoenle, line 750:
```suggestion
		
```

This whole if statement (`v1 == nil && v2 == nil`) is completely redundant and not needed. 

Let's assume `v1 == nil` is true, then the next if statement (see below) will handle that. We don't need to care about `v2` at that point.

```go
if v1 == nil {
  return false
}
```

Wrapped my head around it multiple times and tried to understand what you want to achieve here 😄 

---

Unrelated from that, I would prefer `else if` statements for the other ones (`v1 == nil` and `v2 == nil`), as they seem to belong together. But maybe that's just my personal preference ^^

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -743,6 +744,37 @@ func (r *clusterResource) Create(ctx context.Context, req resource.CreateRequest
...
+		// otherwise we would be changing the passed elements
+		t1, t2 := *v1, *v2
+
+		if !strings.HasPrefix(t1, "v") {
+			t1 = "v" + t1
+		}
+		if !strings.HasPrefix(t2, "v") {
+			t2 = "v" + t2
+		}
+		isLess := semver.Compare(t1, t2) > 0
```
rubenhoenle, line 770:
```suggestion
		return semver.Compare(t1, t2) > 0
```

rubenhoenle (approved):

stackit/internal/services/ske/cluster/resource.go
bahkauv70: Agreed, the first if statement is actually redundant, but it won't hurt (just confuses) as well. I'll remove it. It came to be as a comparator for pointers commonly must differentiate the cases
* both values are nil
* first one is nil, second is not nil and vice versa
* both are not nill and can be directly compared
I find that it reduces the mental load, if those cases are handled explicitly in the code, but as stated above, this introduced a redundant statement. But it has the advantage that I can just read it top-down without maintaining a boolean model in mind... ;-) 

Regarding the latter point: I try to avoid the if cond {return} else { return } cascades and try to handle those statements as guard clauses, e.g.
```
if cond {
  return
}

if other-cond {
  return
}
...
```
This keeps it more readable imho

stackit/internal/services/ske/cluster/resource.go
bahkauv70: Applied

stackit/internal/services/ske/cluster/resource.go
rubenhoenle: Regarding the latter point: Yes, that's completely fine, my only suggestion was this:

```
if cond {
  return
}
else if other-cond {            <----- use an else if if the statements belong somehow together. As said, personal preference :)
  return
}
...
```

rubenhoenle (approved):

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1865,58 +1895,113 @@ func latestMatchingKubernetesVersion(availableVersions []ske.KubernetesVersion,
...
+	if v.State == nil {
+		return false
 	}
 
-	return versionUsed, deprecated, nil
+	return *v.State == VersionStateDeprecated
+}
+
+func isPreview(v *ske.KubernetesVersion) bool {
+	if v == nil {
```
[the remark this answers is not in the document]
bahkauv70: It is just a matter of personal taste: I prefer a list of guard statements before longer conditions, but in this trivial case it of course makes no difference. Probably the generated code will almost identical anywas

PR #678, opened 2025-02-12
Suggestion: remove list of beta resources

h3adex:
Let's remove the list of beta resources in our docs. This list is not up to date and I don't see a good way to automate it.


marceljk (approved):
Thank you for your contribution. This makes sense as the list is no longer up to date.

PR #679, opened 2025-02-12
fix: state drift of "stackit_server"

marceljk:
fix: state drift of "stackit_server" when using "stackit_server_network_interface_attach"


stackit/internal/services/iaas/server/resource.go
```diff
@@ -883,7 +885,7 @@ func mapFields(ctx context.Context, serverResp *iaas.Server, model *Model) error
 		launchedAtValue := *serverResp.LaunchedAt
 		launchedAt = types.StringValue(launchedAtValue.Format(time.RFC3339))
 	}
-	if serverResp.Nics != nil {
+	if serverResp.Nics != nil && !utils.IsUndefined(model.NetworkInterfaces) {
```
bahkauv70, line 888: Is this sufficient? I could imagine that we have to check the list on element level, i.e. check for every interface if it is already present in the state. This is a case similar to the expiration_timestamp issue for the credentials: The lists could be the same, but with different order. This would probably cause a replacement of the server. So we would have to actually compare a canonical version of the lists, i.e. sorted by ids

Fyusel (approved):

PR #682, opened 2025-02-17
fix: make resource/data-source specific region attribute read-only

bahkauv70:


stackit/internal/services/objectstorage/bucket/datasource.go
marceljk, line 1: The region can still be set for all three datasources. They also need to be set to 
```
Optional:    false,
Computed:    true,
```

bahkauv70: Correct, fixing this

marceljk (approved):

PR #697, opened 2025-02-26
Add model serving

PatrickKoss:
Add model serving to stackit terraform provider


stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,651 @@
...
+			&resp.Diagnostics,
+			"Error configuring API client",
+			fmt.Sprintf("Expected configure type stackit.ProviderData, got %T", req.ProviderData),
+		)
+		return
+	}
+
+	var apiClient *modelserving.APIClient
+	var err error
+	if providerData.DnsCustomEndpoint != "" {
```
h3adex, line 95: Is this correct? DnsCustomEndpoint? Should it not be a custom model-serving endpoint.

PatrickKoss: Got me. Copied from dns and forgot to adjust the value here

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,656 @@
...
+	client *modelserving.APIClient
+}
+
+// Metadata returns the resource type name.
+func (r *tokenResource) Metadata(
+	_ context.Context,
+	req resource.MetadataRequest,
+	resp *resource.MetadataResponse,
+) {
+	resp.TypeName = req.ProviderTypeName + "_model_serving_token"
```
busilina, line 68: I'm not sure whether we should call it `_modelserving_token` (like, `_secretsmanager_` or `_objectstorage_`). I'm not aware whether there is a naming convention or not. Just wanted to point it out.

PatrickKoss: Thanks for pointing out. Renamed all _model_serving_ to _modelserving_

stackit/internal/core/core.go
```diff
@@ -25,6 +25,7 @@ type ProviderData struct {
 	LogMeCustomEndpoint             string
 	MariaDBCustomEndpoint           string
 	MongoDBFlexCustomEndpoint       string
+	ModelServingCustomEndpoint      string
```
rubenhoenle, line 28: You will have to add the custom endpoint to the

- [ ] provider model (https://github.com/stackitcloud/terraform-provider-stackit/blob/b5ce160d13eacdfb2068f4770b19756e85e7878c/stackit/provider.go#L114)
- [ ] schema (https://github.com/stackitcloud/terraform-provider-stackit/blob/b5ce160d13eacdfb2068f4770b19756e85e7878c/stackit/provider.go#L149 and https://github.com/stackitcloud/terraform-provider-stackit/blob/b5ce160d13eacdfb2068f4770b19756e85e7878c/stackit/provider.go#L240-L243)
- [ ] and also read the configuration (https://github.com/stackitcloud/terraform-provider-stackit/blob/b5ce160d13eacdfb2068f4770b19756e85e7878c/stackit/provider.go#L355-L357)

rubenhoenle:
Could you add some examples for your new resource and datasource please?

see https://github.com/stackitcloud/terraform-provider-stackit/tree/main/examples

stackit/internal/services/modelserving/model_serving_acc_test.go
```diff
@@ -0,0 +1,237 @@
+package modelserving_test
```
rubenhoenle, line 1: Maybe rename this file to `modelserving_acc_test.go` to stay consistent.

stackit/internal/core/core.go
PatrickKoss: Thanks for pointing out. I added the model serving stuff everywhere in the file.

PatrickKoss:
> Could you add some examples for your new resource and datasource please?
> 
> see https://github.com/stackitcloud/terraform-provider-stackit/tree/main/examples

Added the examples

stackit/internal/services/modelserving/model_serving_acc_test.go
PatrickKoss: renamed.

examples/resources/stackit_modelserving_token/resource.tf
```diff
@@ -0,0 +1,5 @@
+resource "stackit_modelserving_token" "example" {
+  project_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+  region     = "eu01"
```
rubenhoenle, line 3: - [ ] As said in the chat, let's wait with the merge until we decided about how we want to implement the new multi-region concept in the Terraform provider.

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,656 @@
...
+	}
+
+	return &modelserving.CreateTokenPayload{
+		Name:        conversion.StringValueToPointer(model.Name),
+		Description: conversion.StringValueToPointer(model.Description),
+		TtlDuration: conversion.StringValueToPointer(model.TTLDuration),
+	}, nil
+}
+
+func toUpdatePayload(model *Model) (*modelserving.PartialUpdateTokenPayload, error) {
```
h3adex, line 590: These handlers should be part of the stackit-go-sdk. There is a wait package for each service which our tf provider leverages: https://github.com/stackitcloud/stackit-sdk-go/tree/main/services/iaas/wait. The wait packages are not automatically generated

examples/resources/stackit_modelserving_token/resource.tf
rubenhoenle: Decision was made, we will keep it as it is. 😊

h3adex:
Please also run`make generate-docs`and push the model_serving docs to this PR :-)

PatrickKoss:
will come back to this pr once the stackit-go-sdk is released because I need the wait handler

h3adex:
> will come back to this pr once the stackit-go-sdk is released because I need the wait handler

[Release (2025-03-14)](https://github.com/stackitcloud/stackit-sdk-go/releases/tag/release-2025-03-14)

PatrickKoss:
@h3adex @rubenhoenle can you check again? Is this the right way to implement the default region if not in the provider specified?

stackit/internal/services/modelserving/token/datasource.go
```diff
@@ -0,0 +1,224 @@
...
+			"project_id": schema.StringAttribute{
+				Description: "STACKIT project ID to which the model serving auth token is associated.",
+				Required:    true,
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
+				},
+			},
+			"region": schema.StringAttribute{
+				Description: "STACKIT region to which the model serving auth token is associated.",
```
rubenhoenle, line 120:
```suggestion
				Description: "Region to which the model serving auth token is associated. If not defined, the provider region is used.",
```

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,665 @@
...
+			"project_id": schema.StringAttribute{
+				Description: "STACKIT project ID to which the model serving auth token is associated.",
+				Required:    true,
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
+				},
+			},
+			"region": schema.StringAttribute{
+				Description: "STACKIT region to which the model serving auth token is associated.",
```
rubenhoenle, line 156:
```suggestion
				Description: "Region to which the model serving auth token is associated. If not defined, the provider region is used."",
```

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,665 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectId := model.ProjectId.ValueString()
+	tokenId := model.TokenId.ValueString()
+	region := model.Region.ValueString()
+	if region == "" {
+		region = r.client.GetConfig().Region
+	}
```
rubenhoenle, line 310:
```suggestion
	var region string
	if utils.IsUndefined(model.Region) {
		region = r.providerData.GetRegion()
	} else {
		region = model.Region.ValueString()
	}
```

The region of the API client (from the SDK) will be removed long-term due to the implementation of the multi-region concept. So please get the default value from the provider config instead.

rubenhoenle: Referenced utils pkg: `github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils`

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,665 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectId := state.ProjectId.ValueString()
+	tokenId := state.TokenId.ValueString()
+	region := state.Region.ValueString()
+	if region == "" {
+		region = r.client.GetConfig().Region
+	}
```
rubenhoenle, line 389:
```suggestion
var region string
	if utils.IsUndefined(state.Region) {
		region = r.providerData.GetRegion()
	} else {
		region = state.Region.ValueString()
	}
```

same here

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,665 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectId := model.ProjectId.ValueString()
+	tokenId := model.TokenId.ValueString()
+	region := model.Region.ValueString()
+	if region == "" {
+		region = r.client.GetConfig().Region
+	}
```
rubenhoenle, line 490:
```suggestion
var region string
	if utils.IsUndefined(model.Region) {
		region = r.providerData.GetRegion()
	} else {
		region = model.Region.ValueString()
	}
```
same here

stackit/internal/services/modelserving/token/datasource.go
```diff
@@ -0,0 +1,224 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectId := model.ProjectId.ValueString()
+	tokenId := model.TokenId.ValueString()
+	region := model.Region.ValueString()
+	if region == "" {
+		region = d.client.GetConfig().Region
+	}
```
rubenhoenle, line 176:
```suggestion
	var region string
	if utils.IsUndefined(model.Region) {
		region = r.providerData.GetRegion()
	} else {
		region = model.Region.ValueString()
	}
```

same here

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,665 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectId := model.ProjectId.ValueString()
+	region := model.Region.ValueString()
+	if region == "" {
+		region = r.client.GetConfig().Region
+	}
```
rubenhoenle, line 227:
```suggestion
var region string
	if utils.IsUndefined(model.Region) {
		region = r.providerData.GetRegion()
	} else {
		region = model.Region.ValueString()
	}
```

same here

stackit/internal/services/modelserving/token/resource_test.go
```diff
@@ -0,0 +1,348 @@
...
+			expected: &modelserving.PartialUpdateTokenPayload{
+				Name:        utils.Ptr("name"),
+				Description: utils.Ptr("desc"),
+			},
+			isValid: true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
rubenhoenle, line 326:
```suggestion
```

stackit/internal/services/modelserving/token/resource_test.go
```diff
@@ -0,0 +1,348 @@
...
+				State:             types.StringValue("active"),
+				ValidUntil:        types.StringValue("2099-01-01T00:00:00Z"),
+				RotateWhenChanged: types.MapNull(types.StringType),
+			},
+			isValid: true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
rubenhoenle, line 83:
```suggestion
```

stackit/internal/services/modelserving/token/resource_test.go
```diff
@@ -0,0 +1,348 @@
...
+				ValidUntil:        types.StringValue("2099-01-01T00:00:00Z"),
+				Content:           types.StringValue("content"),
+				RotateWhenChanged: types.MapNull(types.StringType),
+			},
+			isValid: true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
rubenhoenle, line 212:
```suggestion
```

stackit/internal/services/modelserving/token/resource_test.go
```diff
@@ -0,0 +1,348 @@
...
+				Name:        utils.Ptr("name"),
+				Description: utils.Ptr("desc"),
+				TtlDuration: utils.Ptr("1h"),
+			},
+			isValid: true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
rubenhoenle, line 272:
```suggestion
```

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,665 @@
...
+				err,
+				tokenId,
+				region,
+				projectId,
+			),
+		)
+		return
+	}
+
+	if updateTokenResp != nil && updateTokenResp.Token.State != nil &&
```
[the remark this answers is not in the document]
PatrickKoss: added a warning and remove resource from tf state which is now recreating the inactive token. Altough not deleting the inactive token

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,665 @@
...
+	region := model.Region.ValueString()
+	if region == "" {
+		region = r.client.GetConfig().Region
+	}
+
+	ctx = tflog.SetField(ctx, "project_id", projectId)
+	ctx = tflog.SetField(ctx, "token_id", tokenId)
+	ctx = tflog.SetField(ctx, "region", region)
+
+	getTokenResp, err := r.client.GetToken(ctx, region, projectId, tokenId).
```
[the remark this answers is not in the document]
PatrickKoss: added a recreate

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,665 @@
...
+			},
+			"project_id": schema.StringAttribute{
+				Description: "STACKIT project ID to which the model serving auth token is associated.",
+				Required:    true,
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
+				},
+			},
+			"region": schema.StringAttribute{
```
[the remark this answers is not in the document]
PatrickKoss: yes makes sense to recreate if the region changes since the token is region bound

stackit/internal/services/modelserving/token/resource_test.go
PatrickKoss: usually shadows something and some linters complain about it :D

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,738 @@
...
+				Required:    false,
+				Optional:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"token_id": schema.StringAttribute{
+				Description: "The model serving auth token ID.",
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
```
h3adex, line 174: it is computed, so plan-modifier will never be triggered.

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,738 @@
...
+
+	getTokenResp, err := r.client.GetToken(ctx, region, projectId, tokenId).
+		Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		if errors.As(err, &oapiErr) {
+			if oapiErr.StatusCode == http.StatusNotFound {
+				// Remove the resource from the state so Terraform will recreate it
+				resp.State.RemoveResource(ctx)
+				core.LogAndAddWarning(
```
h3adex, line 348: we usually don't print a warning in this case. The user will just see on the plan that a token recreation is triggered

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,738 @@
...
+	updateTokenResp, err := r.client.PartialUpdateToken(ctx, region, projectId, tokenId).
+		PartialUpdateTokenPayload(*payload).
+		Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		if errors.As(err, &oapiErr) {
+			if oapiErr.StatusCode == http.StatusNotFound {
+				// Remove the resource from the state so Terraform will recreate it
+				resp.State.RemoveResource(ctx)
+				core.LogAndAddWarning(
```
h3adex, line 459: same here. Usually no warning

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,738 @@
...
+	ctx = tflog.SetField(ctx, "token_id", tokenId)
+	ctx = tflog.SetField(ctx, "region", region)
+
+	// Delete existing model serving auth token. We will ignore the state 'deleting' for now.
+	_, err := r.client.DeleteToken(ctx, region, projectId, tokenId).Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		if errors.As(err, &oapiErr) {
+			if oapiErr.StatusCode == http.StatusNotFound {
+				core.LogAndAddWarning(
```
h3adex, line 562: same here

PatrickKoss:
@h3adex @rubenhoenle can you check again?

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/stackitcloud/stackit-sdk-go/services/loadbalancer v0.18.0
 	github.com/stackitcloud/stackit-sdk-go/services/logme v0.21.0
 	github.com/stackitcloud/stackit-sdk-go/services/mariadb v0.21.0
+	github.com/stackitcloud/stackit-sdk-go/services/modelserving v0.2.0
```
bahkauv70, line 21: Since creation of the PR the version 0.2.1 has been released, so we need to upgrade the dependency

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/core"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/validate"
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource                = &tokenResource{}
+	_ resource.ResourceWithConfigure   = &tokenResource{}
+	_ resource.ResourceWithImportState = &tokenResource{}
```
bahkauv70, line 36: To handle the regions correctly, the resource must also implement the ResourceWithModifyPlan interface with the following implementation:
```go
// ModifyPlan implements resource.ResourceWithModifyPlan.
// Use the modifier to set the effective region in the current plan.
func (r *credentialsGroupResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) { // nolint:gocritic // function signature required by Terraform
	var configModel Model
	// skip initial empty configuration to avoid follow-up errors
	if req.Config.Raw.IsNull() {
		return
	}
	resp.Diagnostics.Append(req.Config.Get(ctx, &configModel)...)
		return
	}

	var planModel Model
	resp.Diagnostics.Append(req.Plan.Get(ctx, &planModel)...)
		return
	}

	coreutils.AdaptRegion(ctx, configModel.Region, &planModel.Region, r.providerData.GetRegion(), resp)
		return
	}

	resp.Diagnostics.Append(resp.Plan.Set(ctx, planModel)...)
		return
	}
}

```

(see e.g. objectstorage/credentialsgroup/resource for comparision)

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+	if providerData.ModelServingCustomEndpoint != "" {
+		ctx = tflog.SetField(
+			ctx,
+			"modelserving_custom_endpoint",
+			providerData.ModelServingCustomEndpoint,
+		)
+		apiClient, err = modelserving.NewAPIClient(
+			config.WithCustomAuth(providerData.RoundTripper),
+			config.WithEndpoint(providerData.ModelServingCustomEndpoint),
+			config.WithRegion(providerData.GetRegion()),
```
bahkauv70, line 114: this call is now deprecated

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+		)
+	} else {
+		apiClient, err = modelserving.NewAPIClient(
+			config.WithCustomAuth(providerData.RoundTripper),
+			config.WithRegion(providerData.GetRegion()),
```
bahkauv70, line 119: this call is now deprecated

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+	if token.ValidUntil != nil {
+		validUntil = token.ValidUntil.Format(time.RFC3339)
+	}
+
+	if waitResp == nil || waitResp.Token == nil || waitResp.Token.State == nil {
+		return fmt.Errorf("response input is nil")
+	}
+
+	idParts := []string{
+		model.ProjectId.ValueString(),
```
bahkauv70, line 682: With the changes to the region handling, the region identifier should now be part of the ID as well, see postgresflex for comparision

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+	model.Name = types.StringPointerValue(token.Name)
+	model.State = types.StringPointerValue(waitResp.Token.State)
+	model.ValidUntil = types.StringValue(validUntil)
+	model.Content = types.StringPointerValue(token.Content)
+	model.Description = types.StringPointerValue(token.Description)
+
+	return nil
+}
+
+func mapToken(token *modelserving.Token, model, state *Model) error {
```
bahkauv70, line 698: this function is almost identical to mapFields. Shouldn't it be a single function?

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+	}
+	if model == nil {
+		return fmt.Errorf("model input is nil")
+	}
+	if state == nil {
+		return fmt.Errorf("state input is nil")
+	}
+
+	// theoretically, should never happen, but still catch null pointers
+	validUntil := time.Now().Format(time.RFC3339)
```
bahkauv70, line 710: What is the correct default value, if the token does not define a validUntil attribute? Is it really 'now' or should it be undefined, i.e. types.StringNull?

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+	if token.ValidUntil != nil {
+		validUntil = token.ValidUntil.Format(time.RFC3339)
+	}
+
+	idParts := []string{
+		model.ProjectId.ValueString(),
+		model.TokenId.ValueString(),
```
bahkauv70, line 717: see above, the region should be part of the id

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+
+// ImportState imports a resource into the Terraform state on success.
+// The expected format of the resource import identifier is: project_id,zone_id,record_set_id
+func (r *tokenResource) ImportState(
+	ctx context.Context,
+	req resource.ImportStateRequest,
+	resp *resource.ImportStateResponse,
+) {
+	idParts := strings.Split(req.ID, core.Separator)
+	if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
```
bahkauv70, line 633: the id should contain the region as well, see https://github.com/stackitcloud/terraform-provider-stackit/blob/1444376f351aa9909a5c759c6b2a5abe0754b719/stackit/internal/services/postgresflex/instance/resource.go#L541 for comparision

stackit/internal/services/modelserving/token/datasource.go
```diff
@@ -0,0 +1,229 @@
...
+		return
+	}
+
+	var apiClient *modelserving.APIClient
+	var err error
+	if providerData.ModelServingCustomEndpoint != "" {
+		apiClient, err = modelserving.NewAPIClient(
+			config.WithCustomAuth(providerData.RoundTripper),
+			config.WithEndpoint(providerData.ModelServingCustomEndpoint),
+			config.WithRegion(providerData.GetRegion()),
```
bahkauv70, line 74: this function has been deprecated

bahkauv70 (changes requested):
Generally please try not to reformat unrelated code if not strictly required by the linter etc.. This blows up the change and increases the required effort for reviewing... If possible, please revert changes to unrelated files, Thanks!

stackit/internal/services/modelserving/token/resource.go
```diff
@@ -0,0 +1,767 @@
...
+	ProjectId         types.String `tfsdk:"project_id"`
+	Region            types.String `tfsdk:"region"`
+	TokenId           types.String `tfsdk:"token_id"`
+	Name              types.String `tfsdk:"name"`
+	Description       types.String `tfsdk:"description"`
+	State             types.String `tfsdk:"state"`
+	ValidUntil        types.String `tfsdk:"valid_until"`
+	TTLDuration       types.String `tfsdk:"ttl_duration"`
+	Content           types.String `tfsdk:"content"`
+	RotateWhenChanged types.Map    `tfsdk:"rotate_when_changed"`
```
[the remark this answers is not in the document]
bahkauv70: The attribute is still defined, but it is not used in any request/response and it is not computed. Could you please clarify it's purpose?

stackit/internal/services/modelserving/token/datasource.go
bahkauv70: the with region call in line 74 can be removed

stackit/internal/services/modelserving/token/resource.go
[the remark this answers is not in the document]
PatrickKoss: I thought I have commented on it :D
Maybe an user want to rotate the token after a certain amount of time like this example:
```
resource "time_rotating" "rotate" {
  rotation_minutes = 1
}

resource "stackit_modelserving_token" "example" {
  project_id = "722ceecb-ac78-47b8-a7bc-65aa591622ce"
  name       = "Example token"
  ttl_duration = "10m"
#   rotate_when_changed = {
#       rotation = time_rotating.rotate.id
#     }
  region = "eu01"
}
```

I stole the code from Moritz service account rotation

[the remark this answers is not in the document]
h3adex: We might want to add it to the generated docs like I've done it:
https://github.com/stackitcloud/terraform-provider-stackit/blob/main/stackit/internal/services/serviceaccount/key/const.go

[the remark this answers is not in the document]
PatrickKoss: generating docs is pain since it generate a bunch of docs that should not be in the pr. Manually deleting them is annoying so I will skip docs here :D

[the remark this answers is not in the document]
h3adex: this issue got fixed. Just rebase from the latest master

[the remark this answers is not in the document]
bahkauv70: Ok, got the intention of the attribute, and now I understood the description of the attribute as well ;-) Perhaps to avoid confusion it would be good to add a sentence to the description, that this is a client-side only attribute and is not actually sent to the API.

bahkauv70:
Ok, looks good now. Could you please provide a sample terraform file and some testing instructions, so we can do some acceptance testsing for this change. If it works, we can merge it asap. 
For future PR please try to avoid reformatting of existing code when not necessary to keep the diffs as small as possible

h3adex:
```terraform
provider "stackit" {
  default_region           = var.stackit_region
  service_account_key_path = var.stackit_service_account_key_path
  enable_beta_resources    = true
}

resource "time_rotating" "rotate" {
  rotation_minutes = 1
}

resource "stackit_modelserving_token" "example" {
  project_id = var.stackit_project_id
  region     = "eu01"
  name       = "Example token"

  rotate_when_changed = {
    rotation = time_rotating.rotate.id
  }
}
```

h3adex:
I also got a region error on the data-source.

Screenshot 2025-03-25 at 16 32 54

PatrickKoss:
> Creating the datasource still has some issues: when using the following terraform file:
> 
> ```hcl
> resource "time_rotating" "rotate" {
>   rotation_minutes = 60
> }
> 
> resource "stackit_modelserving_token" "example" {
>   project_id = var.project_id
>   region     = "eu01"
>   name       = "Example token"
> 
>   rotate_when_changed = {
>     rotation = time_rotating.rotate.id
>   }
> }
> 
> data "stackit_modelserving_token" "exampleds" {
>   project_id = var.project_id
>   token_id   = stackit_modelserving_token.example.token_id
> }
> 
> output "testoutput" {
>   value = [
>     "token id: ${data.stackit_modelserving_token.exampleds.token_id}",
>   ]
> }
> ```
> 
> the following error occurs
> 
> ```shell
> │ Error: Value Conversion Error
> │ 
> │   with data.stackit_modelserving_token.exampleds,
> │ An unexpected error was encountered trying to convert tftypes.Value into token.Model. This
> │ is always an error in the provider. Please report the following to the provider developer:
> │ 
> │ mismatch between struct and object: Struct defines fields not found in object: ttl_duration
> │ and rotate_when_changed.
> ╵
> ```
> 
> The datasource has a different structure as the resource, so we probably need a dedicated model for the datasource

fixed it

PatrickKoss:
> Ok, looks good now. Could you please provide a sample terraform file and some testing instructions, so we can do some acceptance testsing for this change. If it works, we can merge it asap. For future PR please try to avoid reformatting of existing code when not necessary to keep the diffs as small as possible

where do I need to add these instructions? I see there is some kind of acceptance test in each service.

bahkauv70:
The testing instructions are commonly added to the JIRA issues (for this one it would be STACKITTPR-147. Note that due to permission issues the TPR project is most suitable. In this case however the terraform config is sufficient.

stackit/internal/services/modelserving/token/const.go
bahkauv70, line 1: Informational note: We plan to move away from the string constants in a const.go file and intend to use embedded markdown files instead. No change required for this as of now

bahkauv70 (approved):
Looks good now, all tests from my side worked

bahkauv70 (dismissed):

bahkauv70 (approved):

PR #706, opened 2025-03-06
fix(network-interface): Response returns name "" if not set

### LINKED ISSUE #704: stackit_network_interface: Issue with updating resource when name property is not set
Hey,  

I encountered an issue when updating a `stackit_network_interface` that was initially created without specifying a `name`. While the `name` property is optional during creation and does not cause any issues, attempting to update the resource (e.g., by modifying `security_group_ids`) results in the following error:  

```bash
Error: Error updating network interface
│ 
│   with stackit_network_interface.test_nic,
│   on network.tf line 15, in resource "stackit_network_interface" "test_nic":
│   15: resource "stackit_network_interface" "test_nic" {
│ 
│ Calling API: 400 Bad Request, status code 400, Body: {"code":400,"msg":"request body has an error: doesn't match schema #/components/schemas/v1UpdateNIC:
│ Error at \"/name\": string doesn't match the regular expression \"^[A-Za-z0-9]+((-|_|\\s|\\.)[A-Za-z0-9]+)*$\""}
```

### Proposed Solutions  

I see two ways to fix this:

1. Include in the documentation that updating `stackit_network_interface` requires the name property to be set.
2. Make the name property a required field when creating the resource so it doesn’t break later.

I’d love to contribute a fix—do you have a preference on which approach makes the most sense?

Thanks! :)

Fyusel:
This has to be set to a nil value otherwise a name "" is sent via the next request which results in a failing validator

fixes #704

### DIFF
--- a/stackit/internal/services/iaas/networkinterface/resource.go
+++ b/stackit/internal/services/iaas/networkinterface/resource.go
@@ -532,8 +532,13 @@ func mapFields(ctx context.Context, networkInterfaceResp *iaas.NIC, model *Model
 		labels = types.MapNull(types.StringType)
 	}
 
+	networkInterfaceName := types.StringNull()
+	if networkInterfaceResp.Name != nil && *networkInterfaceResp.Name != "" {
+		networkInterfaceName = types.StringPointerValue(networkInterfaceResp.Name)
+	}
+
 	model.NetworkInterfaceId = types.StringValue(networkInterfaceId)
-	model.Name = types.StringPointerValue(networkInterfaceResp.Name)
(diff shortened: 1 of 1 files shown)


marceljk (approved):

jkroepke:
Hi @Fyusel @marceljk,  

As a previous contributor to [terraform-provider-azurerm](https://github.com/hashicorp/terraform-provider-azurerm), I’m wondering why this issue isn’t covered by an additional test case. Does StackIT not utilize end-to-end infrastructure tests or why isn't it a minimum requirement for contributions?  

Thanks in advance.

PR #707, opened 2025-03-07
feat: region adjustments SQLServerFlex

marceljk:


stackit/internal/services/sqlserverflex/instance/resource.go
```diff
@@ -322,7 +366,9 @@ func (r *instanceResource) Create(ctx context.Context, req resource.CreateReques
 		return
 	}
 	projectId := model.ProjectId.ValueString()
+	region := model.Region.ValueString()
```
Fyusel, line 369: I think we need the region check here as well, right?

stackit/internal/services/sqlserverflex/instance/resource.go
```diff
@@ -481,8 +533,11 @@ func (r *instanceResource) Update(ctx context.Context, req resource.UpdateReques
 	}
 	projectId := model.ProjectId.ValueString()
 	instanceId := model.InstanceId.ValueString()
+	region := model.Region.ValueString()
```
Fyusel, line 536: same here?

Fyusel (approved):

PR #712, opened 2025-03-12
Feat: implement sa tokens

h3adex:


h3adex:
Read for review :-)

marceljk (commented):
Thanks for your contribution! Looks overall good and tested it already. Works very nice and I really like rotate_when_change part. Only a few minor things reading documation need to be adjusted

marceljk (approved):

PR #720, opened 2025-03-18
feat/implement-sa-keys

h3adex:
As a user, I want to be able to generate keys for a service account with customizable validity periods and signing options, so that I can choose between using a STACKIT certificate or a private RSA key for signing. The JSON response used to authenticate against the Stackit service should be available as a resource attribute.

Implementing data sources does not seem feasible in this case, since STACKIT does not store the private key itself. Consequently, a data source would lack important content.


h3adex:
Ready for review

rubenhoenle:
- [ ] Check if it makes sense to not only implement a resource but also a data source for this

Also a general question / remark: Can you please link an GitHub issue or Jira issue which describes what you want to achieve within your PRs? You can also provide the information in the description of the PR for this time.

stackit/internal/services/serviceaccount/key/resource_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/stackitcloud/stackit-sdk-go/core/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/serviceaccount"
+)
+
+func approxTimeEqual(t1, t2 time.Time, tolerance time.Duration) bool {
```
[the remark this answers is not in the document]
h3adex: I don't think someone will use it again tbh. I removed the function completely.

stackit/internal/services/serviceaccount/key/resource.go
```diff
@@ -0,0 +1,379 @@
...
+			"id": schema.StringAttribute{
+				Description: descriptions["id"],
+				Computed:    true,
+			},
+			"project_id": schema.StringAttribute{
+				Description: descriptions["project_id"],
+				Required:    true,
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
```
malteehrlen, line 142: I think there is a redundancy here, as if its a valid UUID it would not contain a separator

stackit/internal/services/serviceaccount/key/resource.go
```diff
@@ -0,0 +1,379 @@
...
+// to 'project_id', 'service_account_email', 'ttl_days' or 'public_key' require the resource to be entirely replaced.
+// As a result, the Update function is redundant since any modifications will
+// automatically trigger a resource recreation through Terraform's built-in
+// lifecycle management.
+func (r *serviceAccountKeyResource) Update(ctx context.Context, _ resource.UpdateRequest, resp *resource.UpdateResponse) { // nolint:gocritic // function signature required by Terraform
+	// Service accounts cannot be updated, so we log an error.
+	core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating service account access key", "Service account key can't be updated")
+}
+
+// Delete deletes the service account and removes it from the Terraform state on success.
```
malteehrlen, line 291: the service account key (also same copypasta in other comments)

marceljk (commented):
Looks mostly good only a few small things

marceljk (approved):

PR #725, opened 2025-03-19
Update docs

rubenhoenle:


Fyusel (approved):

Makefile
```diff
@@ -23,6 +23,7 @@ lint: lint-golangci-lint lint-tf
 generate-docs:
 	@echo "Generating documentation with tfplugindocs"
 	@$(SCRIPTS_BASE)/tfplugindocs.sh
+	@git apply $(SCRIPTS_BASE)/docs.patch
```
Fyusel, line 26: I like this :+1: Maybe add a comment why this is added here for our future selves :-)

rubenhoenle:
- [ ] wait with merge until it's clarified if the `members` attribute can be removed. Then the patch wouldn't be even needed.

Makefile
rubenhoenle: Good catch! :smile:

marceljk (approved):

Makefile
rubenhoenle: done

marceljk (approved):

PR #752, opened 2025-03-31
fix(mongodb): store partial state

marceljk:
this fix avoids that an instance is created but doesn't get stored in terraform when an error occurred

### DIFF
--- a/stackit/internal/services/mongodbflex/instance/resource.go
+++ b/stackit/internal/services/mongodbflex/instance/resource.go
@@ -33,10 +33,6 @@ import (
 	"github.com/stackitcloud/stackit-sdk-go/services/mongodbflex/wait"
 )
 
-const (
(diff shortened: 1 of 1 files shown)


stackit/internal/services/mongodbflex/instance/resource.go
```diff
@@ -392,6 +388,16 @@ func (r *instanceResource) Create(ctx context.Context, req resource.CreateReques
 	}
 	instanceId := *createResp.Id
```
bahkauv70, line 389: Add a nil-pointer check just to be sure? I'm not sure, what would happen, if the create call returns without a HTTP error, but the instance could not be created for some internal reasons and is in state "error". In this case, the id might nil (a little paranoid perhaps).

bahkauv70 (approved):

PR #754, opened 2025-04-01
fix(rabbitmq): acceptance tests set enable_monitoring to false

marceljk:
relates to STACKITTPR-96
 - this requires a monitoring_instance_id to work


bahkauv70 (commented):
Should the change update the sdk version(s) in the go.mod?

bahkauv70 (commented):
> Should the change update the sdk version(s) in the go.mod? 

never mind, just found the instructions in the issue. However, we must remember to update the SDK version later

marceljk:
> Should the change update the sdk version(s) in the go.mod?

It isn't required to update the versions in this PR. This PR just fixes the rabbitmq acceptance tests. Because it had monitoring enabled, but without an monitoring instance id it doesn't work. And with the previous state the tests then always failed because the api requires also a monitoring id for monitoring

PR #811, opened 2025-05-08
feat(iaas): min/max acc tests

marceljk:
## Description

relates to STACKITTPR-190

- feat: add min/max acc tests
- fix: stackit_image when VirtioScsi is set to false, the provider returns an error
- fix: stackit_key_pair missing RequiresReplace for name
- fix: stackit_network_area validation for min_prefix_length adjusted
- fix: stackit_public_ip set network_interface_id to computed
- fix: stackit_server doesn't load user_data 

## Checklist

- [x] Issue was linked above


stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -22,10 +25,59 @@ import (
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/testutil"
 )
 
+//go:embed testdata/resource-security-group-min.tf
```
bahkauv70, line 28: personally I prefer to move multiple variable definitions in to a single block, like this:


```suggestion
var (
    //go:embed test1.txt
    test1 string

    //go:embed test2.txt
    test2 string
    ...

)
```

stackit/internal/services/iaas/image/resource.go
```diff
@@ -649,7 +649,7 @@ func mapFields(ctx context.Context, imageResp *iaas.Image, model *Model) error {
 		configModel.SecureBoot = types.BoolPointerValue(imageResp.Config.SecureBoot)
 		configModel.UEFI = types.BoolPointerValue(imageResp.Config.Uefi)
 		configModel.VideoModel = types.StringPointerValue(imageResp.Config.GetVideoModel())
-		configModel.VirtioScsi = types.BoolPointerValue(imageResp.Config.VirtioScsi)
+		configModel.VirtioScsi = types.BoolPointerValue(iaas.PtrBool(imageResp.Config.GetVirtioScsi()))
```
bahkauv70, line 652: I would have expected that the original code would be sufficient. Is this required, what am I missing?

stackit/internal/services/iaas/image/datasource.go
```diff
@@ -309,7 +309,7 @@ func mapDataSourceFields(ctx context.Context, imageResp *iaas.Image, model *Data
 		configModel.SecureBoot = types.BoolPointerValue(imageResp.Config.SecureBoot)
 		configModel.UEFI = types.BoolPointerValue(imageResp.Config.Uefi)
 		configModel.VideoModel = types.StringPointerValue(imageResp.Config.GetVideoModel())
-		configModel.VirtioScsi = types.BoolPointerValue(imageResp.Config.VirtioScsi)
+		configModel.VirtioScsi = types.BoolPointerValue(iaas.PtrBool(imageResp.Config.GetVirtioScsi()))
```
bahkauv70, line 312: I would have expected that the original code would be sufficient. Is this required, what am I missing?

stackit/internal/services/iaas/testdata/resource-server-max.tf
```diff
@@ -0,0 +1,91 @@
...
+  server_id             = stackit_server.server.server_id
+  service_account_email = var.service_account_mail
+}
+
+resource "stackit_server" "server" {
+  project_id         = var.project_id
+  name               = var.name
+  machine_type       = var.machine_type
+  affinity_group     = stackit_affinity_group.affinity_group.affinity_group_id
+  availability_zone  = var.availability_zone
```
bahkauv70, line 80: the 'desired_status' field is not covered here

stackit/internal/services/iaas/testdata/resource-server-max.tf
```diff
@@ -0,0 +1,91 @@
...
+  keypair_name       = stackit_key_pair.key_pair.name
```
bahkauv70, line 81: the image_id attribute is not covered here, but I'm not sure, if this even would make sense

stackit/internal/services/iaas/testdata/resource-server-min.tf
```diff
@@ -0,0 +1,17 @@
...
+variable "name" {}
+variable "machine_type" {}
+variable "image_id" {}
+
+
+resource "stackit_server" "server" {
+  project_id   = var.project_id
+  name         = var.name
+  machine_type = var.machine_type
+  boot_volume = {
```
bahkauv70, line 11: according to the tf-schema, this is an optional attribute. Interesting question here is, if it is even possible to create a server without a boot volume? If not, we actually should it mark as required in the resource...

stackit/internal/services/iaas/testdata/resource-volume-max.tf
```diff
@@ -0,0 +1,19 @@
...
+variable "label" {}
+
+resource "stackit_volume" "volume" {
+  project_id        = var.project_id
+  availability_zone = var.availability_zone
+  name              = var.name
+  size              = var.size
+  description       = var.description
+  performance_class = var.performance_class
+  labels = {
```
bahkauv70, line 16: the source attribute is not covered by the testcase. Problem is, that it is mutually exclusive with the size attribute, making the required structures and testcase even more complicated...

bahkauv70 (changes requested):

stackit/internal/services/iaas/image/resource.go
marceljk: In the original code, the value "VirtioScsi" was the "raw" value from the api and it's a pointer. In case VirtioScsi is set to false in the backend, the backend response without any value for VirtioScsi and terraform would save it as null. 
But when the user explicit set it to false, this would result in an state drift. So we need to set the value to false, in case the api response is null.
With the `GetVirtioScsi` function we get the bool without a pointer. When the pointer is nil, the function returns false. So we can use the `GetVirtioScsi` function, to get always a value and don't have to check if the pointer is nil

stackit/internal/services/iaas/image/datasource.go
marceljk: Same as in the resource above

stackit/internal/services/iaas/testdata/resource-server-max.tf
marceljk: I tried to use image_id, but non of the available images work, because the disk_space is then too low for the image. And the disk_space can not be changed, so I just used the boot_volume for the acc tests.

stackit/internal/services/iaas/testdata/resource-server-min.tf
marceljk: I saw it and had the same thoughts, but either the boot_volume or image_id are required. You can also a message when you try it without them, that you need to set one of them

stackit/internal/services/iaas/testdata/resource-server-max.tf
marceljk: Good catch. I will add it

stackit/internal/services/iaas/iaas_acc_test.go
marceljk: done

stackit/internal/services/iaas/testdata/resource-volume-max.tf
marceljk: i added now a separate resource, to cover also the `source` attribute in the max test

stackit/internal/services/iaas/image/resource.go
bahkauv70: Ah, I see. Ok

stackit/internal/services/iaas/testdata/resource-server-max.tf
bahkauv70: Ok, fair enough

stackit/internal/services/iaas/testdata/resource-server-min.tf
bahkauv70: Ok

bahkauv70 (approved):

PR #850, opened 2025-05-19
feat(objectstorage): Min/Max acceptance tests

Fyusel:
## Description

relates to #1234

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/objectstorage/objectstorage_acc_test.go
+++ b/stackit/internal/services/objectstorage/objectstorage_acc_test.go
@@ -2,14 +2,18 @@ package objectstorage_test
 
 import (
 	"context"
+	_ "embed"
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
-	"github.com/stackitcloud/stackit-sdk-go/core/config"
+
+	stackitSdkConfig "github.com/stackitcloud/stackit-sdk-go/core/config"
 	"github.com/stackitcloud/stackit-sdk-go/core/utils"
 	"github.com/stackitcloud/stackit-sdk-go/services/objectstorage"
 	"github.com/stackitcloud/stackit-sdk-go/services/objectstorage/wait"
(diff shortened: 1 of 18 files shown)


stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -2,86 +2,53 @@ package objectstorage_test
...
-		bucketResource["name"],
-		credentialsGroupResource["project_id"],
-		credentialsGroupResource["name"],
-		credentialResource["expiration_timestamp"],
-	)
+var testConfigVarsMin = config.Variables{
+	"project_id":                           config.StringVariable(testutil.ProjectId),
+	"objectstorage_bucket_name":            config.StringVariable(fmt.Sprintf("tf-acc-test-%s", acctest.RandStringFromCharSet(20, acctest.CharSetAlpha))),
+	"objectstorage_credentials_group_name": config.StringVariable(fmt.Sprintf("tf-acc-test-%s", acctest.RandStringFromCharSet(20, acctest.CharSetAlpha))),
+	"expiration_timestamp":                 config.StringVariable("2027-01-02T03:04:05Z"),
```
marceljk, line 30: I think it would be nice, if the date would be set dynamic or at least the year. So something like `fmt.Sprintf("%s-01-02T03:04:05Z", nextYear)`. Because if after the day "2027-01-02", I think the test will fail, because then the expiration_timestamp will be then in the past.

stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -95,37 +62,38 @@ func TestAccObjectStorageResource(t *testing.T) {
...
-					resourceConfig(),
+						data "stackit_objectstorage_bucket" "bucket" {
+							project_id  = stackit_objectstorage_bucket.bucket.project_id
+							name = stackit_objectstorage_bucket.bucket.name
+						}
+
+						data "stackit_objectstorage_credentials_group" "credentials_group" {
+							project_id  = stackit_objectstorage_credentials_group.credentials_group.project_id
+							credentials_group_id = stackit_objectstorage_credentials_group.credentials_group.credentials_group_id
+						}
```
marceljk, line 85: This doesn't relates directly to the tests, but I saw that in the [docs](https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/data-sources/objectstorage_credentials_group#optional) credentials_group_id is optional and also the name is optional.
I think the credentials_group_id should be required and the name readonly, because this isn't required to read the datasource. And setting the name, could lead to state drift. We could create a separate ticket for this, if you think this doesn't fit to this PR

stackit/internal/services/objectstorage/objectstorage_acc_test.go
```diff
@@ -197,7 +166,8 @@ func TestAccObjectStorageResource(t *testing.T) {
 				ImportStateVerifyIgnore: []string{"region"},
```
marceljk, line 166: I think we decided once, that for region specific resource, we add the region to the ID. Because otherwise it's not possible to import two resources from different regions. But I think we should create a separate ticket to fix this

stackit/internal/services/objectstorage/testfiles/resource-min.tf
```diff
@@ -0,0 +1,21 @@
...
+
+resource "stackit_objectstorage_credentials_group" "credentials_group" {
+  project_id = var.project_id
+  name       = var.objectstorage_credentials_group_name
+}
+
+resource "stackit_objectstorage_credential" "credential" {
+  project_id           = stackit_objectstorage_credentials_group.credentials_group.project_id
+  credentials_group_id = stackit_objectstorage_credentials_group.credentials_group.credentials_group_id
+  expiration_timestamp = var.expiration_timestamp
```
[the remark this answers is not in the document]
Fyusel: Yes, that is what I thought :-D I will add a second resource :+1:

stackit/internal/services/objectstorage/objectstorage_acc_test.go
Fyusel: Yes, that seems to be forgotten in the past. All import functions of the resources do not contain them. I will fix this on the fly.

marceljk (approved):

PR #888, opened 2025-06-20
feat: add pod_address_ranges to ske resource/datasource

h3adex:
## Description

This PR adds the `pod_address_ranges` attribute to the SKE resource/datasource. This attribute is helpful for implementing network peerings and firewall rules dedicated to the SKE cluster. I also implemented some helper functions to update SKE versions in place for E2E tests.

## Checklist

- [x] Issue was linked above


h3adex:
E2E Tests are failing due to https://github.com/stackitcloud/terraform-provider-stackit/issues/889

stackit/internal/services/ske/ske_acc_test.go
```diff
@@ -493,3 +504,121 @@ func testAccCheckSKEDestroy(s *terraform.State) error {
...
+
+type SkeProviderOptions struct {
+	options        *ske.ProviderOptions
+	nodePoolOsName string
+}
+
+// NewSkeProviderOptions fetches the latest available options from SKE.
+func NewSkeProviderOptions(nodePoolOs string) *SkeProviderOptions {
+	// skip if TF_ACC is not set
+	if os.Getenv("TF_ACC") == "" {
```
Fyusel: resource.EnvTfAcc can be used here instead

stackit/internal/services/ske/ske_acc_test.go
```diff
@@ -63,8 +66,8 @@ var testConfigVarsMax = config.Variables{
 	"nodepool_label_value":                             config.StringVariable("value"),
 	"nodepool_max_surge":                               config.StringVariable("1"),
 	"nodepool_max_unavailable":                         config.StringVariable("1"),
-	"nodepool_os_name":                                 config.StringVariable("flatcar"),
-	"nodepool_os_version_min":                          config.StringVariable("4152.2.1"),
+	"nodepool_os_name":                                 config.StringVariable(skeProviderOptions.nodePoolOsName),
```
Fyusel, line 68: I like this very much since those tests are failing often in the last time since the version was adjusted :+1:

stackit/internal/services/ske/ske_acc_test.go
```diff
@@ -493,3 +504,121 @@ func testAccCheckSKEDestroy(s *terraform.State) error {
...
+				}
+			}
+		}
+	}
+
+	return ""
+}
+
+// GetUpdateMachineVersion returns the next supported version after the create version (used for update).
+func (s *SkeProviderOptions) GetUpdateMachineVersion() string {
```
Fyusel, line 604: I would merge those two functions since the code is almost the same for both (GetCreate + GetUpdate).
E.g. you could add a parameter to the function in order to get the first, second, ... entry and use this instead. This would reduce code duplication.
In general I like the two functions so depending on you, you could still use both functions and call in GetUpdate() the GetCreateMachine() function with a parameter.

stackit/internal/services/ske/ske_acc_test.go
```diff
@@ -493,3 +504,121 @@ func testAccCheckSKEDestroy(s *terraform.State) error {
...
+	}
+
+	return &SkeProviderOptions{
+		options:        options,
+		nodePoolOsName: nodePoolOs,
+	}
+}
+
+// GetCreateK8sVersion returns the first supported Kubernetes version (used for create).
+func (s *SkeProviderOptions) GetCreateK8sVersion() string {
```
Fyusel: Same here

stackit/internal/services/ske/ske_acc_test.go
h3adex: If updated it in my review commit. Is that what you meant? Looks better to me

stackit/internal/services/ske/ske_acc_test.go
h3adex: Just my personal opinion: I feel it's cleaner to have four separate functions instead of two with a parameter that decides whether to get the first or second. It just feels easier to read and maintain the code this way.

Fyusel: Yes, I agree. That is what I meant with the last sentence.
So you could have one function with a parameter like for example:
GetMachineVersion(positionInList) and use this function in your functions:

```
GetCreateMachineVersion() {
   GetMachineVersion(0)
}

GetUpdateMachineVersion() {
   GetMachineVersion(1) // or MAX or whatever
}
```

The advantage is that you don't have the code duplicated (like it is right now) but still have the readability + maintainability.

h3adex: Good suggest. Implemented it.

Fyusel:
@h3adex There is an issue present where the var block in the ske_acc_test is initialized during the unit tests which fails.

Besides that this looks good

Fyusel (approved):

PR #891, opened 2025-06-24
feat(provider): Added acceptance tests

Fyusel:
## Description

relates to #1234

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/provider_acc_test.go
+++ b/stackit/provider_acc_test.go
@@ -1,0 +1,259 @@
+package stackit_test
+
+import (
+	_ "embed"
+	"fmt"
+	"os"
+	"path"
+	"regexp"
+	"runtime"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/testutil"
+)
+
+//go:embed testdata/provider-credentials.tf
+var providerCredentialConfig string
+
+//go:embed testdata/provider-invalid-attribute.tf
+var providerInvalidAttribute string
+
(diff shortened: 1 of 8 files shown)


stackit/testdata/provider-all-attributes.tf
```diff
@@ -0,0 +1,42 @@
...
+provider "stackit" {
+  default_region                     = "eu01"
+  credentials_path                   = "~/.stackit/credentials.json"
+  service_account_token              = ""
+  service_account_key_path           = ""
+  service_account_key                = ""
+  private_key_path                   = ""
+  private_key                        = ""
+  service_account_email              = "abc@abc.de"
+  argus_custom_endpoint              = "https://argus.api.eu01.stackit.cloud"
```
marceljk: It's not directly relevant to this PR, but we removed the argus resources a few months ago. I took a quick look into our code and it looks like that this isn't used anymore and could be removed. If you don't want to remove the argus_custom_endpoint option here, we could create a separate story for it

stackit/provider_acc_test.go
```diff
@@ -0,0 +1,250 @@
...
+		return "", fmt.Errorf("Token not set, please set TF_ACC_TEST_PROJECT_SERVICE_ACCOUNT_TOKEN to a valid token to perform tests")
+	}
+	return token, nil
+}
+
+func TestAccEnvVarTokenValid(t *testing.T) {
+	// Check if acceptance tests should be run
+	if v := os.Getenv("TF_ACC"); v == "" {
+		return
+	}
```
marceljk:
```suggestion
	tfAccEnv := "TF_ACC"
	if v := os.Getenv(tfAccEnv); v == "" {
		t.Skip(fmt.Sprintf(
			"Acceptance tests skipped unless env '%s' set",
			tfAccEnv))
		return
	}
```
I think it would be better to add a `t.Skip()`, so that when `go test -v provider_acc_test.go` runs, the tests are marked as skipped tests and not passed

marceljk (approved):

PR #914, opened 2025-07-15
feat(mongodbflex): region adjustment

rubenhoenle:
## Description

relates to STACKITTPR-282

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/mongodbflex/instance/resource.go
+++ b/stackit/internal/services/mongodbflex/instance/resource.go
@@ -53,6 +53,7 @@ type Model struct {
 	Storage        types.Object `tfsdk:"storage"`
 	Version        types.String `tfsdk:"version"`
 	Options        types.Object `tfsdk:"options"`
+	Region         types.String `tfsdk:"region"`
 }
 
 // Struct corresponding to Model.Flavor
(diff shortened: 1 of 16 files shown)


stackit/internal/services/mongodbflex/instance/resource.go
```diff
@@ -120,12 +122,13 @@ func (r *instanceResource) Metadata(_ context.Context, req resource.MetadataRequ
```
marceljk, line 123: The `ModifyPlan` method is missing:
https://github.com/stackitcloud/terraform-provider-stackit/blob/b313ef6a3958eb3dfb626fdb104f8da39146dbd2/stackit/internal/services/sqlserverflex/instance/resource.go#L135-L163

Without this an update will always result in a replace (e.g. if you update the name and you haven't set a explicit region)

stackit/internal/services/mongodbflex/user/resource.go
```diff
@@ -65,12 +67,13 @@ func (r *userResource) Metadata(_ context.Context, req resource.MetadataRequest,
```
marceljk, line 68: Here is also the `ModifyPlan` method missing

Issue #921, opened 2025-07-21
Enhancement: Add terraform import Examples to Resource Documentation


docbyte86:
## Problem description

I'm always frustrated when I need to import existing resources into my Terraform state using the stackitcloud provider because the necessary import commands and instructions are missing from the documentation for all resources. This makes it difficult and time-consuming to incorporate existing infrastructure into Terraform management, as users have to manually figure out the correct import syntax for each resource type. This deviates from the standard practice of other major Terraform providers, which consistently include import instructions in their resource documentation.

## Proposed solution

I propose that the documentation for every resource within the stackitcloud Terraform provider be updated to include clear and concise instructions on how to import existing resources. This should ideally include:

The specific terraform import command syntax for each resource.

An example of the required ID format for importing (e.g., a single ID, a combination of project ID and resource ID, etc.).

This would significantly improve the user experience, streamline the adoption of Terraform for existing STACKIT cloud resources, and align the provider's documentation with industry best practices.

```
# Example of how import instructions could be presented in the documentation for a hypothetical resource:
#
# To import an existing STACKIT Object Storage Bucket, use the following command:
#
# terraform import 'stackit_object_storage_bucket.example' 'project_id,bucket_name'
#
# Replace 'project_id' with your STACKIT Project ID and 'bucket_name' with the name of your bucket.
#
# For example:
# terraform import 'stackit_object_storage_bucket.my_bucket' '01234567-89ab-cdef-0123-456789abcdef,my-existing-bucket'
```

## Alternative solutions (optional)

I have considered relying on community forums or trial-and-error to determine import commands, but these methods are inefficient and error-prone. Providing official, documented import instructions is the most effective and user-friendly solution.

## Additional information

The absence of import documentation is a significant barrier for users trying to adopt Terraform for existing STACKIT infrastructure. Implementing this feature would greatly enhance the usability and completeness of the stackitcloud Terraform provider. Thank you for considering this request.

Fyusel:
Hi @docbyte86,
we created a new terraform release [v0.60.0](https://github.com/stackitcloud/terraform-provider-stackit/releases/tag/v0.60.0) where this is included.

PR #939, opened 2025-08-05
chore(examples): Add import examples for all resources

### LINKED ISSUE #893: Add "import" example section to all resources in the documentation
In the documentation of other providers, you can usually find an `import` section of the resource, e.g. https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/images_image_v2#import

Can we have that for the STACKIT terraform provider documentation, too?

If I want to import a STACKIT resource, I either look at the code, or let the error message show me what is required: `fmt.Sprintf("Expected import identifier with format: [project_id],[image_id]  Got: %q", req.ID),`

Fyusel:
## Description

fixes #893 and #921 

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/docs/resources/image.md
+++ b/docs/resources/image.md
@@ -23,6 +23,12 @@ resource "stackit_image" "example_image" {
 }
 
 # Only use the import statement, if you want to import an existing image
+# Must set a configuration value for the local_file_path attribute as the provider has marked it as required.
+# Since this attribute is not fetched in general from the API call, after adding it this would replace your image resource after an terraform apply.
+# In order to prevent this you need to add:
+#lifecycle {
+#    ignore_changes = [ local_file_path ]
+#  }
 import {
   to = stackit_image.import-example
   id = "${var.project_id},${var.image_id}"
--- a/docs/resources/server.md
+++ b/docs/resources/server.md
@@ -381,7 +381,7 @@ resource "stackit_server" "example" {
 # Only use the import statement, if you want to import an existing server
 # Note: There will be a conflict which needs to be resolved manually.
 # Must set a configuration value for the boot_volume.source_type and boot_volume.source_id attribute as the provider has marked it as required.
-# Since those attributes are not fetched in general form the API call, after adding them this would replace your server resource after an terraform apply.
+# Since those attributes are not fetched in general from the API call, after adding them this would replace your server resource after an terraform apply.
 # In order to prevent this you need to add:
 # lifecycle {
 #   ignore_changes = [ boot_volume ]
--- a/docs/resources/server_network_interface_attach.md
+++ b/docs/resources/server_network_interface_attach.md
@@ -22,7 +22,7 @@ resource "stackit_server_network_interface_attach" "attached_network_interface" 
 # Only use the import statement, if you want to import an existing server network interface attachment
 import {
   to = stackit_server_network_interface_attach.import-example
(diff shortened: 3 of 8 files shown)


docs/resources/affinity_group.md
```diff
@@ -87,6 +87,11 @@ resource "stackit_affinity_group" "example" {
   name       = "example-affinity-group-name"
   policy     = "hard-anti-affinity"
 }
+
+import {
+  to = stackit_affinity_group.import-example
+  id = "${var.project_id},${var.affinity_group_id}"
+}
```
marceljk: Maybe we could set the `to` attribute, to the value of the example resource. And add a small description to the import statement, so that users will know, they don't have to use the import statement when they just want to create the resource. E.g.:
```suggestion

# Only use the import statement, if you want to import an existing affinity group
import {
  to = stackit_affinity_group.example
  id = "${var.project_id},${var.affinity_group_id}"
}
```

Fyusel: :+1: I will add a small description.

I would not use the same name for the example creation and the import since this might confuse the user (might think both are needed for import which is not the case - the first time I used an import statement I also thought I need to add a resource with the corresponding name).

docs/resources/server.md
```diff
@@ -360,7 +360,37 @@ resource "stackit_server" "user-data-from-file" {
...
+  machine_type      = "g1.1"
+  network_interfaces = [
+    stackit_network_interface.example.network_interface_id
+  ]
+}
 
+# Only use the import statement, if you want to import an existing server
+# Note: There will be a conflict which needs to be resolved manually.
+# Must set a configuration value for the boot_volume.source_type and boot_volume.source_id attribute as the provider has marked it as required.
+# Since those attributes are not fetched in general form the API call, after adding them this would replace your server resource after an terraform apply.
```
marceljk: typo
```suggestion
# Since those attributes are not fetched in general from the API call, after adding them this would replace your server resource after an terraform apply.
```

docs/resources/server_network_interface_attach.md
```diff
@@ -18,6 +18,12 @@ resource "stackit_server_network_interface_attach" "attached_network_interface"
   server_id            = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
   network_interface_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
 }
+
+# Only use the import statement, if you want to import an existing server network interface attachment
+import {
+  to = stackit_server_network_interface_attach.import-example
+  id = "${var.project_id},${var.server_id},${var.network_interface_attach_id}"
```
marceljk: A different naming than the attribute of the resource could may confuse users 
```suggestion
  id = "${var.project_id},${var.server_id},${var.network_interface_id}"
```

docs/resources/image.md
```diff
@@ -21,6 +21,12 @@ resource "stackit_image" "example_image" {
   min_disk_size   = 10
   min_ram         = 5
 }
+
+# Only use the import statement, if you want to import an existing image
```
marceljk, line 25: For images there is also a lifecycle.ignore_changes required. 
```hcl
  lifecycle {
    ignore_changes = [ local_file_path ]
  }
```
Otherwise the will force an replacement

PR #940, opened 2025-08-05
fix(networkinterface): Recognize the removal of allowed_addresses and…

Fyusel:
… security_group_ids from terraform config

## Description

relates to #1234

## Checklist

- [ ] Issue was linked above


stackit/internal/services/iaas/networkinterface/resource.go
```diff
@@ -61,6 +62,44 @@ type networkInterfaceResource struct {
...
+	}
+	// If allowed_addresses were completly removed from the config this is not recognized by terraform
+	// since this field is optional and computed therefore this plan modifier is needed.
+	if configModel.AllowedAddresses.IsNull() && !planModel.AllowedAddresses.IsNull() {
+		p := path.Root("allowed_addresses")
+		resp.Diagnostics.Append(resp.Plan.SetAttribute(ctx, p, types.ListNull(types.StringType))...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
```
marceljk: I think we should move this in a utils function, so that we can reuse this logic in other places as well. E.g. `ipv4_namerservers` in `stackit_network` has the same issue.
I think this happens to most of the list attributes, which are optional and computed.
You can find a few of them with the following regex ListAttribute\{(([^\}]|\n)*)Optional:\s+true,

marceljk (approved):

PR #964, opened 2025-08-22
fix(observability): Some plans do not offer to configure alert config…

### LINKED ISSUE #930: Error using stackit_observability_instance as metrics endpoint
## Description

Error creating `stackit_observability_instance` as metrics endpoint related with missing components

## Steps to reproduce

```terraform
resource "stackit_observability_instance" "exporter" {
  project_id = var.stackit_project
  name       = "metrics-exporter"
  plan_name  = "Observability-Metrics-Endpoint-100k-EU01"
  acl                                    = []
  metrics_retention_days                 = 30
  metrics_retention_days_5m_downsampling = 10
  metrics_retention_days_1h_downsampling = 5
}
```

This also happens when you use the data with an instance of that type

```terrafrom
data "stackit_observability_instance" "exporter" {
  project_id  = var.stackit_project
  instance_id = var.observability_instance_id
}
```

## Actual behavior

I'm retrieving 403 messages related with missing permissions (probably because the instance in this mode doesn't deploy all the components)

```
 Getting metrics retention policy: 403 Forbidden
│ (%!s(*string=0xc0004b2eb0)), status code 403, Body: {"detail":"Instance
│ does not have the right components for the route"}
```

## Expected behavior

No errors

Fyusel:
## Description

fixes #930 

## Checklist

- [x] Issue was linked above

### DIFF
--- a/stackit/internal/services/observability/instance/resource.go
+++ b/stackit/internal/services/observability/instance/resource.go
@@ -778,6 +778,52 @@ func (r *instanceResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 	}
 }
 
+// ModifyPlan will be called in the Plan phase.
+// It will check if the plan contains a change that requires replacement. If yes, it will show an error to the user.
+// Since there are observabiltiy plans which do not support specific configurations the request needs to be aborted with an error.
+func (r *instanceResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) { // nolint:gocritic // function signature required by Terraform
+	// If the plan is empty we are deleting the resource
+	if req.Plan.Raw.IsNull() {
+		return
+	}
+
+	var configModel Model
+	// skip initial empty configuration to avoid follow-up errors
+	if req.Config.Raw.IsNull() {
+		return
+	}
+	resp.Diagnostics.Append(req.Config.Get(ctx, &configModel)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	plan, err := loadPlanId(ctx, *r.client, &configModel)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error validating plan", fmt.Sprintf("Loading service plan: %v", err))
+		return
+	}
+
+	// Plan does not support alert config
+	if plan.GetAlertMatchers() == 0 && plan.GetAlertReceivers() == 0 {
+		// If an alert config was set, return an error to the user
+		if !(utils.IsUndefined(configModel.AlertConfig)) {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Error validating plan", fmt.Sprintf("Plan (%s) does not support configuring an alert config. Remove this from your config or use a different plan.", *plan.Name))
+		}
+	}
+
+	// Plan does not support log storage and trace storage
+	if plan.GetLogsStorage() == 0 && plan.GetTracesStorage() == 0 {
+		metricsRetentionDays := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays)
+		metricsRetentionDays5mDownsampling := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays5mDownsampling)
+		metricsRetentionDays1hDownsampling := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays1hDownsampling)
+
+		// If any of the metrics retention days are set, return an error to the user
+		if metricsRetentionDays != nil || metricsRetentionDays5mDownsampling != nil || metricsRetentionDays1hDownsampling != nil {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Error validating plan", fmt.Sprintf("Plan (%s) does not support configuring metrics retention days. Remove this from your config or use a different plan.", *plan.Name))
+		}
+	}
+}
+
 // Create creates the resource and sets the initial Terraform state.
 func (r *instanceResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) { // nolint:gocritic // function signature required by Terraform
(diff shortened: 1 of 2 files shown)


stackit/internal/services/observability/instance/resource.go
```diff
@@ -1040,32 +974,47 @@ func (r *instanceResource) Read(ctx context.Context, req resource.ReadRequest, r
...
 
-	// Map response body to schema
-	err = mapAlertConfigField(ctx, alertConfigResp, &model)
-	if err != nil {
-		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating instance", fmt.Sprintf("Processing API response for the alert config: %v", err))
-		return
-	}
+	// There are plans where no alert matchers and receivers are present e.g. like Observability-Metrics-Endpoint-100k-EU01
+	if plan.GetAlertMatchers() != 0 && plan.GetAlertReceivers() != 0 {
+		alertConfigResp, err := r.client.GetAlertConfigs(ctx, instanceId, projectId).Execute()
```
marceljk, line 1053: I think we should add a validation with `ModifyPlan()`, to check if the alertConfig is set for an unsupported instances. At the moment it's possible to do this in an update and it results in an error. (For creation it's probably the same error, I didn't tested it). 
Beside that error, it could also confuse users, that it looks like it's possible to set this attribute, but it isn't sent to the API and it's only stored in the tf state. 

```hcl
resource "stackit_observability_instance" "frontend" {
  project_id = var.project_id
  name       = "frontend-test"
  plan_name  = "Observability-Frontend-Starter-EU01"
  acl                                    = []
  metrics_retention_days                 = 30
  metrics_retention_days_5m_downsampling = 10
  metrics_retention_days_1h_downsampling = 5
  alert_config = {
    receivers = [
      {
        name = "receiver-test"
        webhooks_configs = [
          {
            ms_teams = false,
            url      = "https://mywebhook.local"
          }
        ]
      },
    ]
    route = {
      receiver = "receiver-test"
    }
  }
}
```

```bash
stackit_observability_instance.frontend: Modifying... [id=66b03760-125b-4c63-9624-741e8a418c02,b512f967-424d-4078-8a85-f6a38df329f6]
╷
│ Error: Provider returned invalid result object after apply
│ 
│ After the apply operation, the provider still indicated an unknown value for stackit_observability_instance.frontend.alert_config.global. All values must be known after apply, so this is always
│ a bug in the provider and should be reported in the provider's own repository. Terraform will still save the other known object values in the state.
╵
```

stackit/internal/services/observability/instance/resource.go
```diff
@@ -1040,32 +974,47 @@ func (r *instanceResource) Read(ctx context.Context, req resource.ReadRequest, r
...
-	}
-
-	// Set state to fully populated data
-	diags = setMetricsRetentions(ctx, &resp.State, &model)
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
-		return
+	// There are some plans which does not offer storage e.g. like Observability-Metrics-Endpoint-100k-EU01
+	if plan.GetLogsStorage() != 0 && plan.GetTracesStorage() != 0 {
+		metricsRetentionResp, err := r.client.GetMetricsStorageRetention(ctx, instanceId, projectId).Execute()
```
marceljk, line 1032: Same for this

stackit/internal/services/observability/instance/resource.go
Fyusel: Hi @marceljk,
I think you were on holiday during this meeting. At this point it is not possible to use ModifyPlan() since in order to check if a specified plan offers alerting or storing of traces and logs we need to call the API for this. When ModifyPlan() is called the client is not yet configured therefore we don't get the information at this point to check this accordingly.

We agreed to add this to the documentation. Since you weren't present of course we can discuss this again.

Regarding the error I will check this, haven't seen this on my side but this should not happen.

Fyusel: Seems to be that I mixed the functions. My description is based on ValidateConfig() not ModifyPlan(). I will have a look at this.

stackit/internal/services/observability/instance/resource.go
```diff
@@ -778,6 +778,52 @@ func (r *instanceResource) Schema(_ context.Context, _ resource.SchemaRequest, r
...
+	plan, err := loadPlanId(ctx, *r.client, &configModel)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error validating plan", fmt.Sprintf("Loading service plan: %v", err))
+		return
+	}
+
+	// Plan does not support alert config
+	if plan.GetAlertMatchers() == 0 && plan.GetAlertReceivers() == 0 {
+		// If an alert config was set, return an error to the user
+		if !(configModel.AlertConfig.IsNull() || configModel.AlertConfig.IsUnknown()) {
```
marceljk: nitpick: could be replaced with the utils function IsUndefined()
```suggestion
		if !utils.IsUndefined(configModel.AlertConfig) {
```

stackit/internal/services/observability/instance/resource.go
```diff
@@ -2211,12 +2161,87 @@ func (r *instanceResource) loadPlanId(ctx context.Context, model *Model) error {
...
+	}
+	return observability.Plan{}, nil
+}
+
+func (r *instanceResource) getAlertConfigs(ctx context.Context, alertConfig *alertConfigModel, model *Model) error {
+	projectId := model.ProjectId.ValueString()
+	instanceId := model.InstanceId.ValueString()
+	var err error
+	// Alert Config
+	if model.AlertConfig.IsUnknown() || model.AlertConfig.IsNull() {
```
marceljk: same here
```suggestion
	if utils.IsUndefined(model.AlertConfig) {
```

marceljk (dismissed):

marceljk (approved):

PR #983, opened 2025-09-04
feat: add custom certificate

matheuspolitano:
## Description
https://jira.schwarz/browse/STACKITCDN-1000

Implement Certificate to CDN custom domain

relates to #1234

## Checklist

- [x] Issue was linked above

### DIFF
--- a/docs/resources/cdn_custom_domain.md
+++ b/docs/resources/cdn_custom_domain.md
@@ -22,7 +22,7 @@ resource "stackit_cdn_custom_domain" "example" {
   name            = "https://xxx.xxx"
   certificate = {
     certificate = "-----BEGIN CERTIFICATE-----\nY2VydGlmaWNhdGVfZGF0YQ==\n-----END CERTIFICATE---"
-    private_key = "-----BEGIN CERTIFICATE-----\nY2VydGlmaWNhdGVfZGF0YQ==\n-----END CERTIFICATE---"
+    private_key = "-----BEGIN RSA PRIVATE KEY-----\nY2VydGlmaWNhdGVfZGF0YQ==\n-----END RSA PRIVATE KEY---"
   }
 }
 


stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -260,7 +309,92 @@ func (r *customDomainResource) ImportState(ctx context.Context, req resource.Imp
...
+		certStr,
+		keyStr,
+		"custom",
+	)
+	certPayload := cdn.PutCustomDomainCustomCertificateAsPutCustomDomainPayloadCertificate(customCert)
+
+	return &certPayload, diags
+}
+
+func mapCustomDomainFields(customDomain *cdn.CustomDomain, model *CustomDomainModel, certificateInput interface{}) error {
```
marceljk: It would make the code cleaner, if here the whole response of `GetCustomDomain()` will be passed, like show in the suggestion. Then only the response itself and the model will be passed to this map function and probably the `normalizeCertificate()` wouldn't be needed anymore and could be removed.
With this change, the function itself needs to be adjusted as well.
```suggestion
func mapCustomDomainFields(customDomain *cdn.GetCustomDomainResponse, model *CustomDomainModel) error {
```

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -160,7 +208,7 @@ func (r *customDomainResource) Create(ctx context.Context, req resource.CreateRe
 		return
 	}
 
-	err = mapCustomDomainFields(waitResp, &model)
+	err = mapCustomDomainFields(waitResp, &model, respPutCustomDomain.Certificate)
```
marceljk: Based on the suggestion of the comment before, you would need to make here an GET call and pass the response to the map function.

```suggestion
	respCustomDomain, err := r.client.GetCustomDomainExecute(ctx, projectId, distributionId, name)
	if err != nil {
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating CDN custom domain", fmt.Sprintf("Calling API: %v", err))
		return
	}
	err = mapCustomDomainFields(respCustomDomain, &model)
```

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -202,7 +251,7 @@ func (r *customDomainResource) Read(ctx context.Context, req resource.ReadReques
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading CDN custom domain", fmt.Sprintf("Calling API: %v", err))
 		return
 	}
-	err = mapCustomDomainFields(customDomainResp.CustomDomain, &model)
+	err = mapCustomDomainFields(customDomainResp.CustomDomain, &model, customDomainResp.Certificate)
```
marceljk: based on the previous comments, this change will be required
```suggestion
	err = mapCustomDomainFields(customDomainResp, &model)
```

stackit/internal/services/cdn/customdomain/datasource.go
```diff
@@ -121,7 +141,7 @@ func (r *customDomainDataSource) Read(ctx context.Context, req datasource.ReadRe
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading CDN custom domain", fmt.Sprintf("Calling API: %v", err))
 		return
 	}
-	err = mapCustomDomainFields(customDomainResp.CustomDomain, &model)
+	err = mapCustomDomainFields(customDomainResp.CustomDomain, &model, customDomainResp.Certificate)
```
marceljk: and this one
```suggestion
	err = mapCustomDomainFields(customDomainResp, &model)
```

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: Actually, I use CustomDomain instead of GetCustomDomainResponse because Create doesn't have GetCustomDomainResponse. This was already our previous solution, so I believe keeping it this way is better :)

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -37,6 +40,18 @@ var (
 	_ resource.ResourceWithConfigure   = &customDomainResource{}
 	_ resource.ResourceWithImportState = &customDomainResource{}
 )
+var certificateSchemaDescriptions = map[string]string{
+	"main":        "The TLS certificate for the custom domain. If omitted, a managed certificate will be used. If the block is specified, a custom certificate is used.",
+	"certificate": "The PEM-encoded TLS certificate. Required for custom certificates.",
+	"private_key": "The PEM-encoded private key for the certificate. Required for custom certificates.",
+	"version":     "A version identifier for the certificate. The certificate will be updated if this field is changed.",
```
marceljk: The description of certificate.version says "_... The certificate will be updated if this field is changed._". At the moment this isn't possible, because Update isn't implemented and it always results in a replace:
https://github.com/stackitcloud/terraform-provider-stackit/blob/c4b81a4bef28cdfa3e9153f40bd5b896d96b49a8/stackit/internal/services/cdn/customdomain/resource.go#L273-L276

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -120,6 +141,26 @@ func (r *customDomainResource) Schema(_ context.Context, _ resource.SchemaReques
...
+					"certificate": schema.StringAttribute{
+						Description: certificateSchemaDescriptions["certificate"],
+						Optional:    true,
+						Sensitive:   true,
+					},
+					"private_key": schema.StringAttribute{
+						Description: certificateSchemaDescriptions["private_key"],
+						Optional:    true,
+						Sensitive:   true,
+					},
```
marceljk, line 162: If I saw it correct then these values can't be read from the API. We recently researched internal in our Team about uses cases for `WriteOnly: true`, which is specially for secrets like the private_key and maybe also the certificate helpful, because the values of them are never stored in the state which increases the security.

We haven't use them yet, but I think this attribute fits perfect for it. If these values should be updatable, then you probably need to an additional attribute, to detect changes and perform an update.
You can read more about it in the official docs [here](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only) or write me a message :)

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -260,30 +314,142 @@ func (r *customDomainResource) ImportState(ctx context.Context, req resource.Imp
...
+			Type: *managedCert.Type,
+		}, nil
+	}
+
+	return Certificate{}, errors.New("certificate structure is empty, neither custom nor managed is set")
+}
+
+// buildCertificatePayload constructs the certificate part of the payload for the API request.
+// It defaults to a managed certificate if the certificate block is omitted, otherwise it creates a custom certificate.
+func buildCertificatePayload(ctx context.Context, model *CustomDomainModel) (*cdn.PutCustomDomainPayloadCertificate, diag.Diagnostics) {
```
marceljk: Replace the return type with error.
```suggestion
func buildCertificatePayload(ctx context.Context, model *CustomDomainModel) (*cdn.PutCustomDomainPayloadCertificate, error) {
```

If you have within the buildCertificatePayload an error from diags, return it like this:
```go
	if diags.HasError() {
		return fmt.Errorf("invalid certificate or private key: %w", core.DiagsToError(diags))
	}
```

And in the place where you call the function buildCertificatePayload(), e.g. in `Create()` you can handle the error like this:
```go
_, err := buildCertificatePayload(ctx, model)
if err != nil {
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating CDN custom domain", fmt.Sprintf("Creating API payload: %v", err))
		return
}
```

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: I added the update method

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -216,9 +270,59 @@ func (r *customDomainResource) Read(ctx context.Context, req resource.ReadReques
 	tflog.Info(ctx, "CDN custom domain read")
 }
 
-func (r *customDomainResource) Update(ctx context.Context, _ resource.UpdateRequest, resp *resource.UpdateResponse) { // nolint:gocritic // function signature required by Terraform
-	// Update shouldn't be called; custom domains have only computed fields and fields that require replacement when changed
-	core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating CDN custom domain", "Custom domain cannot be updated")
+func (r *customDomainResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) { // nolint:gocritic // function signature required by Terraform
+	var plan CustomDomainModel
```
marceljk: Only a nitpick, but could rename `plan` to `model`.
```suggestion
	var model CustomDomainModel
```

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -51,7 +57,7 @@ func configResources(regions string) string {
 				resource "stackit_dns_zone" "dns_zone" {
 					project_id    = "%s"
 					name          = "cdn_acc_test_zone"
-					dns_name      = "cdntestzone.stackit.gg"
+					dns_name      = "cdntest-zone.stackit.gg"
```
marceljk: When I run the acc tests, I get an error that this dns_name is already used. To avoid collisions, this should be (partly) a random name. You can create a var, which stores the dns name for the acc test run and reuse it everywhere. e.g.:

```go
var dnsName = fmt.Sprintf("tf-acc-%s.stackit.gg", acctest.RandStringFromCharSet(5, acctest.CharSetAlpha))
```

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -206,20 +263,28 @@ func TestAccCDNDistributionResource(t *testing.T) {
...
 					resource.TestCheckResourceAttrPair("stackit_cdn_distribution.distribution", "distribution_id", "stackit_cdn_custom_domain.custom_domain", "distribution_id"),
 				),
 			},
+			// Check if the credentials not update
+			{
+				Config: configDatasources(instanceResource["config_regions"], string(cert), string(key)),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("data.stackit_cdn_custom_domain.custom_domain", "certificate.version", "1"),
+				),
+			},
```
marceljk: This step isn't needed and can be removed, because you already checked this a few lines above. Also the datasource can't perform an update, only the resource can do it. You can change the step to use `configCustomDomainResources` if you want, but this step is not necessary

stackit/internal/services/cdn/customdomain/datasource.go
```diff
@@ -89,6 +89,26 @@ func (r *customDomainDataSource) Schema(_ context.Context, _ datasource.SchemaRe
...
+				Optional:    true,
+				Attributes: map[string]schema.Attribute{
+					"certificate": schema.StringAttribute{
+						Description: certificateSchemaDescriptions["certificate"],
+						Optional:    true,
+						Sensitive:   true,
+					},
+					"private_key": schema.StringAttribute{
+						Description: certificateSchemaDescriptions["private_key"],
+						Optional:    true,
```
marceljk: Both values should be changed to computed, since they don't need to be set for a read and it doesn't change everything. 
```suggestion
						Computed:    true,
						Sensitive:   true,
					},
					"private_key": schema.StringAttribute{
						Description: certificateSchemaDescriptions["private_key"],
						Computed:    true,
```

marceljk: Actually, I think they both attributes can be removed, because they can't be read from the API and will always stay empty. This will probably require some adjustments, because then the datasource and resource can't share a model and the map function

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -260,30 +364,138 @@ func (r *customDomainResource) ImportState(ctx context.Context, req resource.Imp
...
+		// Set the computed version from the API response
+		if normalizedCert.Version != nil {
+			certAttributes["version"] = types.Int64Value(*normalizedCert.Version)
+		}
+
+		certificateObj, conversionDiags := types.ObjectValue(certificateTypes, certAttributes)
+		diags.Append(conversionDiags...)
+		if diags.HasError() {
+			return core.DiagsToError(diags)
+		}
```
marceljk: nitpick
```suggestion
		certificateObj, diags := types.ObjectValue(certificateTypes, certAttributes)
		if diags.HasError() {
			return fmt.Errorf("failed to map certificate: %w", core.DiagsToError(diags))
		}
```

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -260,30 +364,138 @@ func (r *customDomainResource) ImportState(ctx context.Context, req resource.Imp
...
+	}
+
+	if customDomainResponse.CustomDomain.Status == nil {
+		return fmt.Errorf("status missing in response")
 	}
+	normalizedCert, err := normalizeCertificate(customDomainResponse.Certificate)
+	if err != nil {
+		return fmt.Errorf("Certificate error in normalizer: %w", err)
+	}
+	var diags diag.Diagnostics
```
marceljk: Can be removed when applying the previous change
```suggestion
```

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -260,30 +364,138 @@ func (r *customDomainResource) ImportState(ctx context.Context, req resource.Imp
...
+		certPayload := cdn.PutCustomDomainManagedCertificateAsPutCustomDomainPayloadCertificate(managedCert)
+		return &certPayload, nil
+	}
+
+	// If the certificate block is specified, it must be a custom certificate.
+	var certModel struct {
+		Certificate types.String `tfsdk:"certificate"`
+		PrivateKey  types.String `tfsdk:"private_key"`
+		Version     types.Int64  `tfsdk:"version"`
+	}
```
marceljk: Can be stored in a separate type and reuse it here.

Add this at the top of this file, next to CustomDomainModel
```go
type CertificateModel struct {
	Certificate types.String `tfsdk:"certificate"`
	PrivateKey  types.String `tfsdk:"private_key"`
	Version     types.Int64  `tfsdk:"version"`
}
```

and then you can use it here
```suggestion
		var certModel CertificateModel
```

stackit/internal/services/cdn/customdomain/resource_test.go
```diff
@@ -80,10 +167,135 @@ func TestMapFields(t *testing.T) {
...
+						"version":     types.Int64Null(),
+						"certificate": types.StringValue(""), // Empty certificate
+						"private_key": types.StringValue(keyPEM),
+					},
+				),
+			},
+			expectErr:      true,
+			expectedErrMsg: "invalid certificate or private key. Please check if the string of the public certificate and private key in PEM format",
+		},
+	}
```
marceljk, line 284: Add a test case, where all attributes are nil. Suggestion:
```suggestion

		"success_managed_when_certificate_attributes_are_nil": {
			model: &CustomDomainModel{
				Certificate: basetypes.NewObjectValueMust(
					certificateTypes,
					map[string]attr.Value{
						"version":     types.Int64Null(),
						"certificate": types.StringNull(),
						"private_key": types.StringNull(),
					},
				),
			},
			expectedPayload: &cdn.PutCustomDomainPayloadCertificate{
				PutCustomDomainManagedCertificate: cdn.NewPutCustomDomainManagedCertificate("managed"),
			},
			expectErr: false,
		},
	}
```

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/datasource.go
matheuspolitano: Done

stackit/internal/services/cdn/cdn_acc_test.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: Done

stackit/internal/services/cdn/cdn_acc_test.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -260,30 +369,130 @@ func (r *customDomainResource) ImportState(ctx context.Context, req resource.Imp
 	tflog.Info(ctx, "CDN custom domain state imported")
 }
 
-func mapCustomDomainFields(customDomain *cdn.CustomDomain, model *CustomDomainModel) error {
-	if customDomain == nil {
+func NormalizeCertificate(certInput cdn.GetCustomDomainResponseGetCertificateAttributeType) (Certificate, error) {
```
marceljk: The change for making this function public can be reverted, because it's only required within the customdomain package
```suggestion
func normalizeCertificate(certInput cdn.GetCustomDomainResponseGetCertificateAttributeType) (Certificate, error) {
```

stackit/internal/services/cdn/customdomain/datasource.go
```diff
@@ -134,3 +157,71 @@ func (r *customDomainDataSource) Read(ctx context.Context, req datasource.ReadRe
 	}
 	tflog.Info(ctx, "CDN custom domain read")
 }
+
+var certificateDataSourceTypes = map[string]attr.Type{
+	"version": types.Int64Type,
+}
```
marceljk: Please move it to the top of this file, next to the type declaration

stackit/internal/services/cdn/customdomain/datasource_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/stackitcloud/stackit-sdk-go/services/cdn"
+)
+
+func TestMapDataSourceFields(t *testing.T) {
+	// Define certificateTypes for the data source schema
+	certificateDataSourceTypes := map[string]attr.Type{
+		"version": types.Int64Type,
+	}
```
marceljk: Not needed, can be reused from the datasource.go file

stackit/internal/services/cdn/customdomain/datasource.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/datasource_test.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/resource_test.go
marceljk: This test is still missing. To make it work, you need to add a check in the `buildCertificatePayload()`. 
Something like this:
```go
	if utils.IsUndefined(certModel.Certificate) || utils.IsUndefined(certModel.PrivateKey) {
		managedCert := cdn.NewPutCustomDomainManagedCertificate("managed")
		certPayload := cdn.PutCustomDomainManagedCertificateAsPutCustomDomainPayloadCertificate(managedCert)
		return &certPayload, nil
	}
```
Probably it would be even better, to return an error with a message that "certificate" and "private_key" must be set, when the Certificate model is not empty.

examples/resources/stackit_cdn_custom_domain/resource.tf
```diff
@@ -2,10 +2,14 @@ resource "stackit_cdn_custom_domain" "example" {
   project_id      = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
   distribution_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
   name            = "https://xxx.xxx"
+  certificate = {
+    certificate = "-----BEGIN CERTIFICATE-----\nY2VydGlmaWNhdGVfZGF0YQ==\n-----END CERTIFICATE---"
+    private_key = "-----BEGIN CERTIFICATE-----\nY2VydGlmaWNhdGVfZGF0YQ==\n-----END CERTIFICATE---"
```
marceljk:
```suggestion
    private_key = "-----BEGIN RSA PRIVATE KEY-----\nY2VydGlmaWNhdGVfZGF0YQ==\n-----END RSA PRIVATE KEY---"
```

stackit/internal/services/cdn/customdomain/resource_test.go
matheuspolitano: we found another solution

stackit/internal/services/cdn/customdomain/datasource.go
```diff
@@ -134,3 +161,67 @@ func (r *customDomainDataSource) Read(ctx context.Context, req datasource.ReadRe
...
 }
+
+func mapCustomDomainDataSourceFields(customDomainResponse *cdn.GetCustomDomainResponse, model *customDomainDataSourceModel, projectId, distributionId string) error {
+	if customDomainResponse == nil {
+		return fmt.Errorf("response input is nil")
+	}
+	if model == nil {
+		return fmt.Errorf("model input is nil")
+	}
+	if customDomainResponse.CustomDomain.Name == nil {
```
marceljk, line 177: add here an nil check for `customDomainResponse.CustomDomain` to prevent a nil pointer

stackit/internal/services/cdn/customdomain/datasource.go
```diff
@@ -134,3 +161,67 @@ func (r *customDomainDataSource) Read(ctx context.Context, req datasource.ReadRe
...
+		return fmt.Errorf("name is missing in response")
+	}
+	if customDomainResponse.CustomDomain.Status == nil {
+		return fmt.Errorf("status missing in response")
+	}
+
```
marceljk, line 183: and here also one nil check for `customDomainResponse.Certificate`

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -260,30 +369,135 @@ func (r *customDomainResource) ImportState(ctx context.Context, req resource.Imp
...
-	if customDomain.Name == nil {
-		return fmt.Errorf("Name is missing in response")
+	if customDomainResponse.CustomDomain.Name == nil {
+		return fmt.Errorf("name is missing in response")
+	}
+
+	if customDomainResponse.CustomDomain.Status == nil {
+		return fmt.Errorf("status missing in response")
 	}
+	normalizedCert, err := normalizeCertificate(customDomainResponse.Certificate)
```
marceljk: add the same nil checks for `customDomainResponse.CustomDomain` and `customDomainResponse.Certificate`

stackit/internal/services/cdn/customdomain/resource.go
```diff
@@ -260,30 +369,135 @@ func (r *customDomainResource) ImportState(ctx context.Context, req resource.Imp
...
+			Type: *managedCert.Type,
+		}, nil
+	}
+
+	return Certificate{}, errors.New("certificate structure is empty, neither custom nor managed is set")
+}
+
+// buildCertificatePayload constructs the certificate part of the payload for the API request.
+// It defaults to a managed certificate if the certificate block is omitted, otherwise it creates a custom certificate.
+func buildCertificatePayload(ctx context.Context, model *CustomDomainModel) (*cdn.PutCustomDomainPayloadCertificate, error) {
```
marceljk: rename `buildCertifcatePayload` to `toCertificatePayload` to stay consistent with our other resources

stackit/internal/services/cdn/customdomain/resource_test.go
```diff
@@ -1,90 +1,314 @@
...
 	"github.com/stackitcloud/stackit-sdk-go/services/cdn"
 )
 
 func TestMapFields(t *testing.T) {
+	// Redefine certificateTypes locally for testing, matching the updated schema
+	certificateTypes := map[string]attr.Type{
+		"version":     types.Int64Type,
+		"certificate": types.StringType,
+		"private_key": types.StringType,
+	}
```
marceljk, line 30: Is not needed

stackit/internal/services/cdn/customdomain/datasource.go
matheuspolitano: i've already check it in the normalizeCertificate

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: Done :) we dont check need extra check because we check further for purpose, and also custom domain will be never nil because our sdk always return it, but i added just not discuss more about it

stackit/internal/services/cdn/customdomain/resource.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/resource_test.go
matheuspolitano: Done

stackit/internal/services/cdn/customdomain/resource.go
marceljk: It's just because we had at some places nil pointer issues in the past, where usually the pointer shouldn't be nil. So it's just to ensure, that it doesn't break in the future :)

rubenhoenle (approved):

PR #994, opened 2025-09-16
CDN - Add Geofencing 

matheuspolitano:
## Description

https://jira.schwarz/browse/STACKITCDN-875

## Checklist

- [x] Issue was linked above


docs/data-sources/cdn_custom_domain.md
```diff
@@ -32,8 +32,19 @@ data "stackit_cdn_custom_domain" "example" {
 - `name` (String)
 - `project_id` (String) STACKIT project ID associated with the distribution
 
+### Optional
+
+- `certificate` (Attributes) The TLS certificate for the custom domain. If omitted, a managed certificate will be used. If the block is specified, a custom certificate is used. (see [below for nested schema](#nestedatt--certificate))
```
rubenhoenle, line 37: Isn't the certificate stuff done within https://github.com/stackitcloud/terraform-provider-stackit/pull/983? Please don't duplicate stuff across PRs :smile:

rubenhoenle:
And I would kindly ask you to squash your commits into some meaningful ones. Having these 26 commits makes not much sense to me.

docs/data-sources/cdn_custom_domain.md
matheuspolitano: Actually i'm using as base that mr as base, i am waiting for be deployed the certificate MR. So is not a duplicate

rubenhoenle: Got it, but that means this PR will be on hold until https://github.com/stackitcloud/terraform-provider-stackit/pull/983 is merged.

rubenhoenle:
@matheuspolitano Could you please resolve the conflicts? :)

docs/data-sources/cdn_custom_domain.md
matheuspolitano: yep

stackit/internal/services/cdn/distribution/datasource.go
```diff
@@ -142,6 +142,13 @@ func (r *distributionDataSource) Schema(_ context.Context, _ datasource.SchemaRe
 								Description: schemaDescriptions["config_backend_origin_request_headers"],
 								ElementType: types.StringType,
 							},
+							"geofencing": schema.MapAttribute{
+								Description: "A map of URLs to a list of countries where content is allowed.",
+								Optional:    true,
```
marceljk: must be computed for the datasource
```suggestion
								Computed:    true,
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -722,6 +760,22 @@ func convertConfig(ctx context.Context, model *Model) (*cdn.Config, error) {
 		}
 	}
 
+	// geofencing
+	geofencing := map[string][]string{}
+	if configModel.Backend.Geofencing != nil {
+		for endpoint, countryCodes := range *configModel.Backend.Geofencing {
+			geofencingContry := make([]string, len(countryCodes))
```
marceljk: typo
```suggestion
			geofencingCountry := make([]string, len(countryCodes))
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -256,6 +262,13 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
 								Description: schemaDescriptions["config_backend_origin_request_headers"],
 								ElementType: types.StringType,
 							},
+							"geofencing": schema.MapAttribute{
+								Description: "A map of URLs to a list of countries where content is allowed.",
+								Optional:    true,
+								ElementType: types.ListType{
+									ElemType: types.StringType,
+								},
```
marceljk, line 270: add here a validator, otherwise it's possible to configure an empty map for geofencing which results in an error
```suggestion
								Description: "A map of URLs to a list of countries where content is allowed.",
								Optional:    true,
								ElementType: types.ListType{
									ElemType: types.StringType,
								},
								Validators: []validator.Map{
									mapvalidator.SizeAtLeast(1),
								},
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -88,9 +88,10 @@ type optimizerConfig struct {
 }
 
 type backend struct {
-	Type                 string             `tfsdk:"type"`                   // The type of the backend. Currently, only "http" backend is supported
-	OriginURL            string             `tfsdk:"origin_url"`             // The origin URL of the backend
-	OriginRequestHeaders *map[string]string `tfsdk:"origin_request_headers"` // Request headers that should be added by the CDN distribution to incoming requests
+	Type                 string               `tfsdk:"type"`                   // The type of the backend. Currently, only "http" backend is supported
+	OriginURL            string               `tfsdk:"origin_url"`             // The origin URL of the backend
+	OriginRequestHeaders *map[string]string   `tfsdk:"origin_request_headers"` // Request headers that should be added by the CDN distribution to incoming requests
+	Geofencing           *map[string][]string `tfsdk:"geofencing"`             // The geofencing is an object mapping multiple alternative origins to country codes.
```
marceljk: When using the resource like that, it results in an error.

```hcl
resource "stackit_cdn_distribution" "distribution" {
  project_id = var.project_id
  config = {
    backend = {
      type       = "http"
      origin_url = "https://test-backend-1.cdn-dev.runs.onstackit.cloud"
      geofencing = {
        "https://test-backend-1.cdn-dev.runs.onstackit.cloud" = [null] # instead a string with a country code, null was set
      }
    }
    regions           = ["EU", "US"]
    blocked_countries = ["CU", "AQ"]
    optimizer = {
      enabled = true
    }
  }
}
```

```bash
│ Error: Error creating CDN distribution
│ 
│   with stackit_cdn_distribution.distribution,
│   on main.tf line 20, in resource "stackit_cdn_distribution" "distribution":
│   20: resource "stackit_cdn_distribution" "distribution" {
│ 
│ Creating API payload: failed as parsing (Value Conversion Error) An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the
│ provider developer:
│ 
│ Received null value, however the target type cannot handle null values. Use the corresponding `types` package type, a pointer type or a custom type that handles null values.
│ 
│ Path: backend.geofencing["https://test-backend-1.cdn-dev.runs.onstackit.cloud"][0]
│ Target Type: string
│ Suggested `types` Type: basetypes.StringValue
│ Suggested Pointer Type: *string
```

From a first look, you probably need to change the type. Then you need to check within `convertConfig` if the countryCode is nil and if that's the case, return an error
```suggestion
	Geofencing           *map[string][]*string `tfsdk:"geofencing"`             // The geofencing is an object mapping multiple alternative origins to country codes.
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -584,11 +598,34 @@ func mapFields(distribution *cdn.Distribution, model *Model) error {
...
+	// geofencing
+	geofencingVal := types.MapNull(geofencingTypes.ElemType)
+	if geofencingAPI := distribution.Config.Backend.HttpBackend.Geofencing; geofencingAPI != nil && len(*geofencingAPI) > 0 {
+		geofencingMap := make(map[string]attr.Value)
+		for url, countries := range *geofencingAPI {
+			countryVals := []attr.Value{}
+			for _, country := range countries {
+				countryVals = append(countryVals, types.StringValue(country))
+			}
+			listVal, diags := types.ListValue(types.StringType, countryVals)
```
marceljk: Not sure if the order of the countries list will always remain in the same order like in the create / update request. But to prevent false recognized state drifts in terraform, I would suggest that you use here our `utils.ReconcileStringSlices` function, in order to ensure that the order of the countries will always match the order from the state.
You can take a look here, how we use it in a other resource
https://github.com/stackitcloud/terraform-provider-stackit/blob/f16b83170a819511dc6f899f42684936a83e42e0/stackit/internal/services/iaas/networkinterface/resource.go#L497-L503

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -256,6 +262,13 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
 								Description: schemaDescriptions["config_backend_origin_request_headers"],
 								ElementType: types.StringType,
 							},
+							"geofencing": schema.MapAttribute{
```
rubenhoenle, line 265: Please adjust the acc tests and add your new field there for the resource and datasource.

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: I've implemented the ReconcileStringSlices

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -696,6 +738,19 @@ func mapFields(distribution *cdn.Distribution, model *Model) error {
...
+		return nil // Or []string{} depending on how you want to handle it
+	}
+	stringSlice := make([]string, 0, len(pointerSlice))
+	for _, strPtr := range pointerSlice {
+		if strPtr != nil { // Safely skip any nil pointers in the list
+			stringSlice = append(stringSlice, *strPtr)
+		}
+	}
+	return stringSlice
+}
```
marceljk: Can you move this to the utils package and add a few tests for this function? This is a function we probably need in some other places as well.

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -598,28 +616,52 @@ func mapFields(distribution *cdn.Distribution, model *Model) error {
...
-			countryVals := []attr.Value{}
-			for _, country := range countries {
-				countryVals = append(countryVals, types.StringValue(country))
-			}
-			listVal, diags := types.ListValue(types.StringType, countryVals)
+		newGeofencingMap := *geofencingAPI
+		for url, newCountries := range newGeofencingMap {
+			oldCountriesPtrs := oldGeofencingMap[url]
+
+			oldCountries := convertPointerSliceToStringSlice(oldCountriesPtrs)
```
marceljk: it could be the case, that the old map might doesn't have the url, e.g. if someone extended the geofencing values via the api directly. So you need to check here, if the oldMap has the url key and if not, it must be handled correctly

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -764,15 +819,18 @@ func convertConfig(ctx context.Context, model *Model) (*cdn.Config, error) {
 	geofencing := map[string][]string{}
 	if configModel.Backend.Geofencing != nil {
 		for endpoint, countryCodes := range *configModel.Backend.Geofencing {
-			geofencingContry := make([]string, len(countryCodes))
-			for i, countryCode := range countryCodes {
-				validatedBlockedCountry, err := validateCountryCode(countryCode)
+			geofencingCountry := make([]string, len(countryCodes))
+			for i, countryCodePtr := range countryCodes {
+				if countryCodePtr == nil {
+					continue
```
marceljk: If an url has only one country code and this one is set to null, it will result in an api error, because the URL would have an empty array. So I would return here an error instead of continue

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -421,13 +424,28 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
...
+	var geofencingPatch *map[string][]string
+	if configModel.Backend.Geofencing != nil {
+		gf := make(map[string][]string)
+		for url, countries := range *configModel.Backend.Geofencing {
+			countryStrings := make([]string, len(countries))
+			for i, countryPtr := range countries {
+				if countryPtr != nil {
+					countryStrings[i] = *countryPtr
+				}
+			}
```
marceljk, line 475: You should return here an error, if coutryPtr is nil to prevent api issues. This could happen if someone sets null within the country list and then an empty array will be send to the api

stackit/internal/services/cdn/distribution/resource.go

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: Done

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: Actually we dont need it beacuse the convertPointerSliceToStringSlice already handle it

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: Dine

matheuspolitano: i'm return the list of string

stackit/internal/utils/utils.go
```diff
@@ -95,6 +95,19 @@ func SimplifyBackupSchedule(schedule string) string {
 	return simplifiedSchedule
 }
 
+// ConvertPointerSliceToStringSlice safely converts a slice of string pointers to a slice of strings.
+func ConvertPointerSliceToStringSlice(pointerSlice []*string) []string {
```
marceljk, line 99: Can you add a few testcases for this function?

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -149,6 +158,7 @@ func TestAccCDNDistributionResource(t *testing.T) {
 	fullDomainName := fmt.Sprintf("%s.%s", instanceResource["custom_domain_prefix"], instanceResource["dns_name"])
 	organization := fmt.Sprintf("organization-%s", uuid.NewString())
 	cert, key := makeCertAndKey(t, organization)
+	gefencing := []string{"DE", "ES"}
```
marceljk: typo
```suggestion
	geofencing := []string{"DE", "ES"}
```

stackit/internal/utils/utils.go
matheuspolitano: Done

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -408,12 +424,30 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
...
 	}
 
+	var geofencingPatch *map[string][]string
+	if configModel.Backend.Geofencing != nil {
+		gf := make(map[string][]string)
+		for url, countries := range *configModel.Backend.Geofencing {
+			countryStrings := make([]string, len(countries))
+			for i, countryPtr := range countries {
+				if countryPtr == nil {
+					core.LogAndAddError(ctx, &resp.Diagnostics, "Update CDN distribution", fmt.Sprintf("Geodencing url %q has a null value", url))
```
marceljk: typo
```suggestion
					core.LogAndAddError(ctx, &resp.Diagnostics, "Update CDN distribution", fmt.Sprintf("Geofencing url %q has a null value", url))
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -722,6 +804,25 @@ func convertConfig(ctx context.Context, model *Model) (*cdn.Config, error) {
...
 	}
 
+	// geofencing
+	geofencing := map[string][]string{}
+	if configModel.Backend.Geofencing != nil {
+		for endpoint, countryCodes := range *configModel.Backend.Geofencing {
+			geofencingCountry := make([]string, len(countryCodes))
+			for i, countryCodePtr := range countryCodes {
+				if countryCodePtr == nil {
+					return nil, errors.New("country can not be nil")
```
marceljk: nitpick
```suggestion
					return nil, fmt.Errorf("geofencing url %q has a null value", endpoint)
```

stackit/internal/utils/utils_test.go
```diff
@@ -128,6 +128,60 @@ func TestListValuetoStrSlice(t *testing.T) {
 	}
 }
 
+// stringp is a helper function to return a pointer to a string.
+func stringp(s string) *string {
+	return &s
+}
```
marceljk: this function isn't necessary. you can use instead the `utils.Ptr("apple")` function from our go sdk, when you add this import `"github.com/stackitcloud/stackit-sdk-go/core/utils"`

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -256,6 +262,16 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
 								Description: schemaDescriptions["config_backend_origin_request_headers"],
 								ElementType: types.StringType,
 							},
+							"geofencing": schema.MapAttribute{
+								Description: "A map of URLs to a list of countries where content is allowed.",
+								Optional:    true,
+								ElementType: types.ListType{
+									ElemType: types.StringType,
+								},
```
marceljk, line 270: The validation if the geofencing values are correct works when terraform run but not during the plan, what can be annoying for our customers. To prevent this, you can add the `ValidateConfig` function, where you can validate it already during the plan. E.g.:

```go
func (r *distributionResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
    var model Model
    resp.Diagnostics.Append(req.Config.Get(ctx, &model)...)
        return
    }

    if !utils.IsUndefined(model.Config) {
        var config distributionConfig
        if !model.Config.IsNull() {
            diags := model.Config.As(ctx, &config, basetypes.ObjectAsOptions{})
            if diags.HasError() {
                return
            }
            if geofencing := config.Backend.Geofencing; geofencing != nil {
                for url, region := range *geofencing {
                    if region == nil {
                        core.LogAndAddError(ctx, &resp.Diagnostics, "Error configuring cdn distribution", fmt.Sprintf("Region for url %q is not set", url))
                        continue
                    }
                    if len(region) == 0 {
                        core.LogAndAddError(ctx, &resp.Diagnostics, "Error configuring cdn distribution", fmt.Sprintf("Region for url %q is empty", url))
                        continue
                    }
                }
            }
        }
    }
}
```

stackit/internal/utils/utils_test.go
matheuspolitano: Done

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -408,12 +424,30 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
...
+		geofencingPatch = &gf
+	}
+
 	configPatch := &cdn.ConfigPatch{
 		Backend: &cdn.ConfigPatchBackend{
 			HttpBackendPatch: &cdn.HttpBackendPatch{
 				OriginRequestHeaders: configModel.Backend.OriginRequestHeaders,
 				OriginUrl:            &configModel.Backend.OriginURL,
 				Type:                 &configModel.Backend.Type,
+				Geofencing:           geofencingPatch, // Use the converted variable
```
marceljk: When the geofencing map is completely removed within an update, than it results in an error state, because in the api the geofencing map is not set to an empty map and terraform detects then a state drift. To avoid this, you just need to declare the initial value of `geofencingPatch` with an empty map and then it will be sent correctly on the patch request.

```suggestion
	geofencingPatch := map[string][]string{}
	if configModel.Backend.Geofencing != nil {
		gf := make(map[string][]string)
		for url, countries := range *configModel.Backend.Geofencing {
			countryStrings := make([]string, len(countries))
			for i, countryPtr := range countries {
				if countryPtr == nil {
					core.LogAndAddError(ctx, &resp.Diagnostics, "Update CDN distribution", fmt.Sprintf("Geofencing url %q has a null value", url))
					return
				}
				countryStrings[i] = *countryPtr
			}
			gf[url] = countryStrings
		}
		geofencingPatch = gf
	}

	configPatch := &cdn.ConfigPatch{
		Backend: &cdn.ConfigPatchBackend{
			HttpBackendPatch: &cdn.HttpBackendPatch{
				OriginRequestHeaders: configModel.Backend.OriginRequestHeaders,
				OriginUrl:            &configModel.Backend.OriginURL,
				Type:                 &configModel.Backend.Type,
				Geofencing:           &geofencingPatch, // Use the converted variable
```

You can also simplify the code a bit more by removing the `gf` map and use directly `geofencingPatch`.

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: done and i create a extra validation

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -500,7 +534,44 @@ func (r *distributionResource) ImportState(ctx context.Context, req resource.Imp
...
+					}
+					if len(region) == 0 {
+						core.LogAndAddError(ctx, &resp.Diagnostics, "Invalid geofencing config", fmt.Sprintf("The list of countries for URL %q must not be empty.", url))
+						continue
+					}
+
+					for i, countryPtr := range region {
+						if countryPtr == nil {
+							core.LogAndAddError(ctx, &resp.Diagnostics, "Invalid geofencing config", fmt.Sprintf("Found a null value in the country list for URL %q at index %d.", url, i))
+							break
```
marceljk: Change the `break` to `continue`, so that if there are multiple null values, the user will get all at once
```suggestion
							continue
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -500,7 +534,44 @@ func (r *distributionResource) ImportState(ctx context.Context, req resource.Imp
 	tflog.Info(ctx, "CDN distribution state imported")
 }
 
-func mapFields(distribution *cdn.Distribution, model *Model) error {
+func (r *distributionResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
```
marceljk: Can you move this function above the Schema function. We usually have it at the place and it's more an unspoken rule

PR #986, opened 2025-09-08
Added Security_group_id to tests and added non deprecated machine_type

bm-stackit:
## Description

https://github.com/stackitcloud/terraform-provider-stackit/pull/923#discussion_r2298151919
https://github.com/stackitcloud/terraform-provider-stackit/pull/923#issuecomment-3237210871

It was requested that we add the security group id to the tests and change machine_type to non deprecated ones.

## Checklist

- [x] Issue was linked above
- [x] Code format was applied: `make fmt`
- [x] Examples were added / adjusted (see `examples/` directory)
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI)
- [x] Unit tests got implemented or updated
- [x] Acceptance tests got implemented or updated (see e.g. [here](https://github.com/stackitcloud/terraform-provider-stackit/blob/f5f99d170996b208672ae684b6da53420e369563/stackit/internal/services/dns/dns_acc_test.go))
- [x] Unit tests are passing: `make test` (will be checked by CI)
- [x] No linter issues: `make lint` (will be checked by CI)

PR #1009, opened 2025-09-25
added new attribite load_balancer_security_group

bm-stackit:
## Description

The attribute of the actual loadbalancer security group wasnt in the provider yet and is now added.

## Checklist

- [x] Issue was linked above

### DIFF
--- a/docs/resources/loadbalancer.md
+++ b/docs/resources/loadbalancer.md
@@ -242,8 +242,9 @@ import {
 ### Read-Only
 
 - `id` (String) Terraform's internal resource ID. It is structured as "`project_id`","region","`name`".
+- `load_balancer_security_group_id` (String) The ID of the egress security group assigned to the Load Balancer's internal machines. This ID is essential for allowing traffic from the Load Balancer to targets in different networks or STACKIT network areas (SNA). To enable this, create a security group rule for your target VMs and set the `remote_security_group_id` of that rule to this value. This is typically used when `disable_security_group_assignment` is set to `true`.
 - `private_address` (String) Transient private Load Balancer IP address. It can change any time.
-- `security_group_id` (String) The ID of the egress security group assigned to the Load Balancer's internal machines. This ID is essential for allowing traffic from the Load Balancer to targets in different networks or STACKIT network areas (SNA). To enable this, create a security group rule for your target VMs and set the `remote_security_group_id` of that rule to this value. This is typically used when `disable_security_group_assignment` is set to `true`.
+- `security_group_id` (String) The ID of the backend security group
 
 <a id="nestedatt--listeners"></a>
 ### Nested Schema for `listeners`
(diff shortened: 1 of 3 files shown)


bm-stackit:
Gonna add the acceptance tests now

stackit/internal/services/loadbalancer/loadbalancer/datasource.go
```diff
@@ -349,6 +350,10 @@ func (r *loadBalancerDataSource) Schema(_ context.Context, _ datasource.SchemaRe
 				Description: descriptions["security_group_id"],
 				Computed:    true,
 			},
+			"load_balancer_security_group_id": schema.StringAttribute{
```
rubenhoenle, line 353: So you came here 2 weeks ago and introduced the `security_group` field to this resource (https://github.com/stackitcloud/terraform-provider-stackit/pull/986). Now you want to introduce a second one. 

I don't quite get it yet. What's the other security group id for now?

bm-stackit: Since I took this one over from a colleague that is no longer in our team we realized a bit too late, that he missed adding the other crucial security group attribute that is on the loadbalancer vm itself which is needed to do the actual routing. The Load_balancer_security_group_id is the security group of the LB VM and the other security_group_id is the one that we create but do not assign. Load_balancer_security_group_id is being put into the remote_security_group_id of the backend security group which in return allows communication of the LB and the Backend target.

rubenhoenle: You were so speeding with your last PR, now we have the mess. Anyways, lets break it down:

---

We have the new `load_balancer_security_group_id` field now which relates to this field in the API docs right?

https://docs.api.stackit.cloud/documentation/load-balancer/version/v2#tag/Load-Balancer/operation/APIService_GetLoadBalancer

image

---

And then there is the "old" `security_group_id` attribute which relates to this field in the API docs, right?

image

bm-stackit: Your observations are correct. The Load_balancer_security_group_id(loadBalanerSecurityGroup) is important for loadbalancers across different Projects in 1 SNA. And the security_group_id (targetSecurityGroup) is useful for load balancers with targets in the same project but different networks.

rubenhoenle: So I would say the `security_group_id` field be named `target_security_group_id` field instead. Which isn't possible now that easily because we have a deprecation time of 6 months...

bm-stackit: That is not a problem, leave it as is.

rubenhoenle: > That is not a problem, leave it as is.

Maybe for you but I care about our users. IMO the `security_group_id` field must be deprecated and a new field `target_security_group_id` should be added.

Issue #1014, opened 2025-10-07
dns_record_set: CNAME to stackit_postgresflex_user.host not possible


thomas-schn:
## Description

I want to create a dns_record_set of type CNAME the points to my stackit_postgresflex_user.host. 
Unfortunately the name of the record is postfixed with the hosted_zone fro that Project.

The terraform plan looks like this 

```
  # stackit_dns_record_set.db_dns_record_set will be created
  + resource "stackit_dns_record_set" "db_dns_record_set" {
      + active        = true
      + comment       = "DNS alias auf die Datenbank von app"
      + error         = (known after apply)
      + fqdn          = (known after apply)
      + id            = (known after apply)
      + name          = "my.hosted.zone.de" 
      + project_id    = "---"
      + record_set_id = (known after apply)
      + records       = [
          + "xxx.postgresql.eu01.onstackit.cloud",
        ]
      + state         = (known after apply)
      + ttl           = 300
      + type          = "CNAME"
      + zone_id       = "---"
    }
Plan: 1 to add, 0 to change, 0 to destroy.
Changes to Outputs:
  + db_dns_alias = (known after apply)
```
But after applying my terraform module, the output looks like this:

```
stackit_dns_record_set.db_dns_record_set: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to stackit_dns_record_set.db_dns_record_set, provider
│ "provider[\"registry.opentofu.org/stackitcloud/stackit\"]" produced an
│ unexpected new value: .records[0]: was
│ cty.StringVal("xxx.postgresql.eu01.onstackit.cloud"),
│ but now
│ cty.StringVal("xxx.postgresql.eu01.onstackit.cloud.my.hosted.zone.de.").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
```

## Steps to reproduce

```terraform
resource "stackit_postgresflex_instance" "db_instance" {
  project_id = local.project_id
  name       = "${var.environment_name}-db-instance"
  region     = "eu01"
  acl = [
    "${stackit_public_ip.my_ip.ip}/32"
  ]

  backup_schedule = "00 00 * * *"
  flavor = { # is 1/1 in AWS
    cpu = 2
    ram = 4
  }
  replicas = 1
  storage = {
    class = "premium-perf2-stackit"
    size  = 5
  }
  version = "16"
}

resource "stackit_postgresflex_user" "db_user" {
[...]

PR #1019, opened 2025-10-09
feat(dns): add validation for records in CNAME record-sets

marceljk:
## Description

feat(dns): add validation for records in CNAME record-sets, to prevent state drifts if the record is not as FQDN specified

relates to #1014

## Checklist

- [x] Issue was linked above

### DIFF
--- a/stackit/internal/validate/validate.go
+++ b/stackit/internal/validate/validate.go
@@ -67,6 +67,23 @@ func UUID() *Validator {
 	}
 }
 
+func NoUUID() *Validator {
+	description := "value must not be an UUID"
+
+	return &Validator{
+		description: description,
+		validate: func(_ context.Context, req validator.StringRequest, resp *validator.StringResponse) {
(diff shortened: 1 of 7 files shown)


stackit/internal/validate/validate.go
```diff
@@ -115,6 +115,14 @@ func RecordSet() *Validator {
 					))
 				}
 			case "CNAME":
+				name := req.ConfigValue.ValueString()
+				if name == "" || name[len(name)-1] != '.' {
+					resp.Diagnostics.Append(validatordiag.InvalidAttributeValueDiagnostic(
+						req.Path,
+						"value must be an Fully Qualified Domain Name (FQDN)",
```
xtavras: I would extend the message to be more precise "value must be an Fully Qualified Domain Name (FQDN) and end with dot (.)" or similar.

marceljk: Thanks for the hint. I'll extend it :)

Benjosh95 (approved):

PR #1022, opened 2025-10-09
fix(dns): store IDs immediately after provisioning

rubenhoenle:
## Description

relates to STACKITTPR-373

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/dns/recordset/resource.go
+++ b/stackit/internal/services/dns/recordset/resource.go
@@ -8,7 +8,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
-	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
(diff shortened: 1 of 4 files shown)


stackit/internal/services/dns/recordset/resource.go
```diff
@@ -219,7 +218,16 @@ func (r *recordSetResource) Create(ctx context.Context, req resource.CreateReque
...
 		return
 	}
-	ctx = tflog.SetField(ctx, "record_set_id", *recordSetResp.Rrset.Id)
+
+	// Write id attributes to state before polling via the wait handler - just in case anything goes wrong during the wait handler
+	utils.SetAndLogStateFields(ctx, &resp.Diagnostics, &resp.State, map[string]any{
+		"project_id":    projectId,
+		"zone_id":       zoneId,
+		"record_set_id": *recordSetResp.Rrset.Id,
+	})
```
marceljk, line 226: I like the utils function you added, which makes it more convenient to set multiple attributes at the same time. The only thing which could  result in an issue in the future is, if the required fields changes and we forget to update it either here or in the ImportState function. 
Maybe we could add an additional function, which creates the map and enforces via the parameters that always all fields are set. What do you thing about this?

marceljk: Could be something like this:
```go
func setReadFields(projectId string, zoneId string, recordSetId string) map[string]interface{} {
	return map[string]interface{}{
		"project_id": projectId,
		"zone_id":    zoneId,
		"record_set_id": recordSetId,
	}
}
```

rubenhoenle: Well then I would say we need to add some interface which enforces the implementation of this func for a resource.

I also thought about that, but didn't do it because it doesn't really solve the problem. With this additional func when you touch the id attributes (add a new one) you have to think of adjusting this func. Also there's the problem that when the order of the id attributes changes or someone messes up the order of the id attributes in this func call, you will have a very hard debugging session :smile:

marceljk: I understand, the function could create more issues than solving them. Especially the point with changing the order or extending the function convinced me. I would say then it is fine to just use the map, because it's easier to see which variable will be used for which attribute in the schema

Issue #792, opened 2025-04-25
Feature Request: stackit_observability_instance Log and Traces Retention Configuration


mitterle-sit:
## Problem description

I would like to be able to configure the Log and Traces Retention Days via the Terraform Ressource stackit_observability_instance. Currently this is only possible for the metrics with the attribute [metrics_retention_days](https://registry.terraform.io/providers/stackitcloud/stackit/latest#metrics_retention_days-1)

## Proposed solution

Add two attributes for the tracing and log retention in days, like log_retention_days and tracing_retention_days.

rubenhoenle:
Hey @mitterle-sit,

thanks for your feature request. We already had this in our backlog, I bumped the priority for it so we can implement this soon for you. 😄

mitterle-sit:
Hey @rubenhoenle ,
thank you for the priority bump.

mitterle-sit:
Closed by #1032

PR #1032, opened 2025-10-16
feat(observability): add logs and traces retentions days

mitterle-sit:
## Description

relates to #792

## Checklist

- [x] Issue was linked above

### DIFF
--- a/examples/resources/stackit_observability_instance/resource.tf
+++ b/examples/resources/stackit_observability_instance/resource.tf
@@ -1,7 +1,7 @@
 resource "stackit_observability_instance" "example" {
   project_id                             = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
   name                                   = "example-instance"
-  plan_name                              = "Observability-Monitoring-Medium-EU01"
+  plan_name                              = "Observability-Starter-EU01"
   acl                                    = ["1.1.1.1/32", "2.2.2.2/32"]
   logs_retention_days                    = 30
   traces_retention_days                  = 30
--- a/stackit/internal/services/observability/instance/resource.go
+++ b/stackit/internal/services/observability/instance/resource.go
@@ -872,9 +872,16 @@ func (r *instanceResource) ModifyPlan(ctx context.Context, req resource.ModifyPl
 		metricsRetentionDays := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays)
 		metricsRetentionDays5mDownsampling := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays5mDownsampling)
 		metricsRetentionDays1hDownsampling := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays1hDownsampling)
-
+		// If logs retention days are set, return an error to the user
+		if logsRetentionDays != nil {
+			resp.Diagnostics.AddAttributeError(path.Root("logs_retention_days"), "Error validating plan", fmt.Sprintf("Plan (%s) does not support configuring logs retention days. Remove this from your config or use a different plan.", *plan.Name))
+		}
+		// If traces retention days are set, return an error to the user
+		if tracesRetentionDays != nil {
+			resp.Diagnostics.AddAttributeError(path.Root("traces_retention_days"), "Error validating plan", fmt.Sprintf("Plan (%s) does not support configuring trace retention days. Remove this from your config or use a different plan.", *plan.Name))
+		}
 		// If any of the metrics retention days are set, return an error to the user
-		if logsRetentionDays != nil || tracesRetentionDays != nil || metricsRetentionDays != nil || metricsRetentionDays5mDownsampling != nil || metricsRetentionDays1hDownsampling != nil {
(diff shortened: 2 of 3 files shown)


stackit/internal/services/observability/instance/resource.go
```diff
@@ -974,13 +986,48 @@ func (r *instanceResource) Create(ctx context.Context, req resource.CreateReques
 		if resp.Diagnostics.HasError() {
 			return
 		}
+
```
Fyusel, line 998: Very good PR!

There is another logic in the **ModifyPlan()** function which checks if the current plan does support what was configured, if not a creation is not started (terraform plan will show an error in this case). Can you adjust this function as well in order to check **traces_retention_days** and **logs_retention_days** please

stackit/internal/services/observability/observability_acc_test.go
```diff
@@ -69,6 +69,8 @@ var testConfigVarsMax = config.Variables{
 	"alert_annotation": config.StringVariable("annotation1"),
 	"alert_interval":   config.StringVariable("5h"),
 	// max instance
+	"logs_retention_days":                    config.StringVariable("30"),
+	"traces_retention_days":                  config.StringVariable("30"),
```
Fyusel, line 73: This fails at the moment. The issue here is that you need to adjust the **resource-max.tf** file as well (located under testdata in the same directory as the acc test). 

(The values are passed via **ConfigVariables: testConfigVarsMax**.)

Fyusel: If not already known: You can run the acceptance tests for a single service like this:
TF_ACC=1 TF_ACC_PROJECT_ID=<project_id> TF_ACC_TEST_PROJECT_SERVICE_ACCOUNT_EMAIL=<email> TF_ACC_REGION="eu01" go test -timeout=20m -v stackit/internal/services/observability/observability_acc_test.go

mitterle-sit: Thank you for the hint I'll check it

Fyusel (dismissed):

Fyusel:
Hi @mitterle-sit, looks good from my side :+1:  There are only some linter issues which needs to be fixed

Fyusel (dismissed):

stackit/internal/services/observability/instance/resource.go
```diff
@@ -855,12 +867,14 @@ func (r *instanceResource) ModifyPlan(ctx context.Context, req resource.ModifyPl
...
+		tracesRetentionDays := conversion.Int64ValueToPointer(configModel.TracesRetentionDays)
 		metricsRetentionDays := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays)
 		metricsRetentionDays5mDownsampling := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays5mDownsampling)
 		metricsRetentionDays1hDownsampling := conversion.Int64ValueToPointer(configModel.MetricsRetentionDays1hDownsampling)
 
 		// If any of the metrics retention days are set, return an error to the user
-		if metricsRetentionDays != nil || metricsRetentionDays5mDownsampling != nil || metricsRetentionDays1hDownsampling != nil {
+		if logsRetentionDays != nil || tracesRetentionDays != nil || metricsRetentionDays != nil || metricsRetentionDays5mDownsampling != nil || metricsRetentionDays1hDownsampling != nil {
 			core.LogAndAddError(ctx, &resp.Diagnostics, "Error validating plan", fmt.Sprintf("Plan (%s) does not support configuring metrics retention days. Remove this from your config or use a different plan.", *plan.Name))
 		}
```
marceljk: Can you add a separate error message for logsRetentionDays and tracesRetentionDays? Because this could confuse users, if they have set one of them and then they get a error message, that metrics retention days is not supported.

E.g.:
```suggestion
if logsRetentionDays != nil {
			resp.Diagnostics.AddAttributeError(path.Root("logs_retention_days"), "Error validating plan", fmt.Sprintf("Plan (%s) does not support configuring logs retention days. Remove this from your config or use a different plan.", *plan.Name))
		}
		if tracesRetentionDays != nil {
			resp.Diagnostics.AddAttributeError(path.Root("traces_retention_days"), "Error validating plan", fmt.Sprintf("Plan (%s) does not support configuring trace retention days. Remove this from your config or use a different plan.", *plan.Name))
		}
		if metricsRetentionDays != nil || metricsRetentionDays5mDownsampling != nil || metricsRetentionDays1hDownsampling != nil {
			core.LogAndAddError(ctx, &resp.Diagnostics, "Error validating plan", fmt.Sprintf("Plan (%s) does not support configuring metrics retention days. Remove this from your config or use a different plan.", *plan.Name))
		}
```

examples/resources/stackit_observability_instance/resource.tf
```diff
@@ -3,6 +3,8 @@ resource "stackit_observability_instance" "example" {
   name                                   = "example-instance"
   plan_name                              = "Observability-Monitoring-Medium-EU01"
```
marceljk: The plan in this example doesn't support the logs_retention_days, traces, and metrics. So I would suggest to change the plan. For me it works with "Observability-Starter-EU01"

marceljk (commented):
Thanks for your contribution. Overall the PR looks good, I left only two nitpick comments

PR #1052, opened 2025-11-10
Feature/dns skip wait and partial state

PatrickKoss:
## Description

I want to get the initial buy in to skip the wait handlers (needed for some client libraries) and set the state in the Create implementation of the interface to the model null values + ids. The current implementation throws errors that the model has attributes that are "unknown" for some clients.

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/dns/recordset/resource.go
+++ b/stackit/internal/services/dns/recordset/resource.go
@@ -203,6 +203,13 @@ func (r *recordSetResource) Create(ctx context.Context, req resource.CreateReque
 		return
 	}
 
+	// Get a fresh copy from plan for minimal state
+	var minimalModel Model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &minimalModel)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
 	projectId := model.ProjectId.ValueString()
 	zoneId := model.ZoneId.ValueString()
 	ctx = tflog.SetField(ctx, "project_id", projectId)
@@ -223,16 +230,16 @@ func (r *recordSetResource) Create(ctx context.Context, req resource.CreateReque
 
 	// Write id attributes to state before polling via the wait handler - just in case anything goes wrong during the wait handler
 	recordSetId := *recordSetResp.Rrset.Id
-	model.RecordSetId = types.StringValue(recordSetId)
-	model.Id = utils.BuildInternalTerraformId(projectId, zoneId, recordSetId)
+	minimalModel.RecordSetId = types.StringValue(recordSetId)
+	minimalModel.Id = utils.BuildInternalTerraformId(projectId, zoneId, recordSetId)
 
 	// Set all unknown/null fields to null before saving state
-	if err := utils.SetModelFieldsToNull(ctx, &model); err != nil {
+	if err := utils.SetModelFieldsToNull(ctx, &minimalModel); err != nil {
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating record set", fmt.Sprintf("Setting model fields to null: %v", err))
 		return
 	}
 
-	diags = resp.State.Set(ctx, model)
+	diags = resp.State.Set(ctx, minimalModel)
 	resp.Diagnostics.Append(diags...)
 	if resp.Diagnostics.HasError() {
 		return
@@ -245,7 +252,13 @@ func (r *recordSetResource) Create(ctx context.Context, req resource.CreateReque
 
 	waitResp, err := wait.CreateRecordSetWaitHandler(ctx, r.client, projectId, zoneId, *recordSetResp.Rrset.Id).WaitWithContext(ctx)
 	if err != nil {
-		tflog.Warn(ctx, fmt.Sprintf("Record set creation waiting failed: %v. The record set creation was triggered but waiting for completion was interrupted. The record set may still be creating.", err))
+		if utils.ShouldIgnoreWaitError(err) {
+			tflog.Warn(ctx, fmt.Sprintf("Record set creation waiting failed: %v. The record set creation was triggered but waiting for completion was interrupted. The record set may still be creating.", err))
+			return
+		}
+
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating record set", fmt.Sprintf("Record set creation waiting: %v", err))
+
 		return
 	}
 
@@ -342,7 +355,13 @@ func (r *recordSetResource) Update(ctx context.Context, req resource.UpdateReque
 	}
 
 	if !utils.ShouldWait() {
-		tflog.Info(ctx, "Skipping wait; async mode for Crossplane/Upjet")
+		if utils.ShouldIgnoreWaitError(err) {
+			tflog.Warn(ctx, fmt.Sprintf("Record set update waiting failed: %v. The record set update was triggered but waiting for completion was interrupted. The record set may still be updating.", err))
+			return
+		}
+
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating record set", fmt.Sprintf("Record set update waiting: %v", err))
+
 		return
 	}
 
@@ -403,7 +422,13 @@ func (r *recordSetResource) Delete(ctx context.Context, req resource.DeleteReque
 
 	_, err = wait.DeleteRecordSetWaitHandler(ctx, r.client, projectId, zoneId, recordSetId).WaitWithContext(ctx)
 	if err != nil {
-		tflog.Warn(ctx, fmt.Sprintf("Record set deletion waiting failed: %v. The record set deletion was triggered but waiting for completion was interrupted. The record set may still be deleting.", err))
+		if utils.ShouldIgnoreWaitError(err) {
+			tflog.Warn(ctx, fmt.Sprintf("Record set deletion waiting failed: %v. The record set deletion was triggered but waiting for completion was interrupted. The record set may still be deleting.", err))
+			return
+		}
+
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error deleting record set", fmt.Sprintf("Record set deletion waiting: %v", err))
+
 		return
 	}
 	tflog.Info(ctx, "DNS record set deleted")
--- a/stackit/internal/utils/utils.go
+++ b/stackit/internal/utils/utils.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
+	"net/http"
 	"os"
 	"reflect"
 	"regexp"
@@ -750,4 +752,34 @@ func createNullValue(_ context.Context, val attr.Value, attrType attr.Type) attr
 func ShouldWait() bool {
 	v := os.Getenv("STACKIT_TF_WAIT_FOR_READY")
 	return v == "" || strings.EqualFold(v, "true")
+}
+
+// ShouldIgnoreWaitError determines if a wait error should be ignored.
+// Returns true for transient errors like context timeouts, network errors, and API 5xx errors.
+// These errors are considered recoverable and shouldn't fail the operation.
+func ShouldIgnoreWaitError(err error) bool {
+	// Context errors (deadline exceeded, canceled)
+	if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+		return true
+	}
+
+	// Network errors (timeout, connection refused, etc.)
+	var netErr net.Error
+	if errors.As(err, &netErr) {
+		return true
+	}
+
+	// API 5xx errors
+	var oapiErr *oapierror.GenericOpenAPIError
+	if errors.As(err, &oapiErr) {
+		if oapiErr.StatusCode >= http.StatusInternalServerError {
+			return true
+		}
+
+		if oapiErr.StatusCode == http.StatusNotFound || oapiErr.StatusCode == http.StatusGone {
+			return true
+		}
+	}
+
+	return false
 }
--- a/stackit/internal/services/dns/zone/resource.go
+++ b/stackit/internal/services/dns/zone/resource.go
@@ -287,6 +287,13 @@ func (r *zoneResource) Create(ctx context.Context, req resource.CreateRequest, r
 		return
 	}
 
+	// Get a fresh copy from plan for minimal state
+	var minimalModel Model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &minimalModel)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
 	projectId := model.ProjectId.ValueString()
 	ctx = tflog.SetField(ctx, "project_id", projectId)
 
@@ -305,16 +312,16 @@ func (r *zoneResource) Create(ctx context.Context, req resource.CreateRequest, r
 
 	// Save minimal state immediately after API call succeeds to ensure idempotency
 	zoneId := *createResp.Zone.Id
-	model.ZoneId = types.StringValue(zoneId)
-	model.Id = utils.BuildInternalTerraformId(projectId, zoneId)
+	minimalModel.ZoneId = types.StringValue(zoneId)
+	minimalModel.Id = utils.BuildInternalTerraformId(projectId, zoneId)
 
 	// Set all unknown/null fields to null before saving state
-	if err := utils.SetModelFieldsToNull(ctx, &model); err != nil {
+	if err := utils.SetModelFieldsToNull(ctx, &minimalModel); err != nil {
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating zone", fmt.Sprintf("Setting model fields to null: %v", err))
 		return
 	}
 
-	diags := resp.State.Set(ctx, model)
+	diags := resp.State.Set(ctx, minimalModel)
 	resp.Diagnostics.Append(diags...)
 	if diags.HasError() {
 		return
@@ -327,7 +334,19 @@ func (r *zoneResource) Create(ctx context.Context, req resource.CreateRequest, r
 
 	waitResp, err := wait.CreateZoneWaitHandler(ctx, r.client, projectId, zoneId).WaitWithContext(ctx)
 	if err != nil {
-		tflog.Warn(ctx, fmt.Sprintf("Zone creation waiting failed: %v. The zone creation was triggered but waiting for completion was interrupted. The zone may still be creating.", err))
+		if utils.ShouldIgnoreWaitError(err) {
+			tflog.Warn(
+				ctx,
+				fmt.Sprintf(
+					"Zone creation waiting failed: %v. The zone creation was triggered but waiting for completion was interrupted. The zone may still be creating.",
+					err,
+				),
+			)
+			return
+		}
+
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating zone", fmt.Sprintf("Zone creation waiting: %v", err))
+
 		return
 	}
 
@@ -424,7 +443,19 @@ func (r *zoneResource) Update(ctx context.Context, req resource.UpdateRequest, r
 
 	waitResp, err := wait.PartialUpdateZoneWaitHandler(ctx, r.client, projectId, zoneId).WaitWithContext(ctx)
 	if err != nil {
-		tflog.Warn(ctx, fmt.Sprintf("Zone update waiting failed: %v. The zone update was triggered but waiting for completion was interrupted. The zone may still be updating.", err))
+		if utils.ShouldIgnoreWaitError(err) {
+			tflog.Warn(
+				ctx,
+				fmt.Sprintf(
+					"Zone update waiting failed: %v. The zone update was triggered but waiting for completion was interrupted. The zone may still be updating.",
+					err,
+				),
+			)
+			return
+		}
+
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating zone", fmt.Sprintf("Zone update waiting: %v", err))
+
 		return
 	}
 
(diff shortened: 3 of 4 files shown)


stackit/internal/services/dns/recordset/resource.go
```diff
@@ -59,12 +61,20 @@ type recordSetResource struct {
 }
 
 // Metadata returns the resource type name.
-func (r *recordSetResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+func (r *recordSetResource) Metadata(
+	_ context.Context,
+	req resource.MetadataRequest,
+	resp *resource.MetadataResponse,
+) {
```
marceljk: There are a lot of formatting changes. Please revert them, so we can easier see what has changed

PatrickKoss: had some auto reformatter in the ide :D Reverted hopefully all formatting stuff

Makefile
```diff
@@ -1,17 +1,26 @@
...
 # SETUP AND TOOL INITIALIZATION TASKS
 project-help:
 	@$(SCRIPTS_BASE)/project.sh help
 
 project-tools:
 	@$(SCRIPTS_BASE)/project.sh tools
 
+# GOLANGCI-LINT INSTALLATION
+$(GOLANGCI_LINT):
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | bash -s -- -b bin v$(GOLANGCI_VERSION)
```
rubenhoenle: Running bash scripts blindly from a master branch of another repository is a no-go for me, sorry

Overall, what's the point of this? This whole thing feels wrong to me. For managing development dependencies there are things like dev containers, devenvs, nix flakes, ...

I'm aware we're not providing any of these currently, but this download process inside the Makefile seems pretty hacky to me :smile:

rubenhoenle (changes requested):

stackit/internal/services/dns/recordset/resource.go
```diff
@@ -264,8 +278,19 @@ func (r *recordSetResource) Read(ctx context.Context, req resource.ReadRequest,
 	ctx = tflog.SetField(ctx, "zone_id", zoneId)
 	ctx = tflog.SetField(ctx, "record_set_id", recordSetId)
 
+	if recordSetId == "" || zoneId == "" || projectId == "" {
```
rubenhoenle: How should this case even happen? The project id is a required input of the resource schema

stackit/internal/services/dns/recordset/resource.go
```diff
@@ -264,8 +278,19 @@ func (r *recordSetResource) Read(ctx context.Context, req resource.ReadRequest,
...
+		tflog.Info(ctx, "Record set ID, zone ID, or project ID is empty, removing resource")
+		resp.State.RemoveResource(ctx)
+		return
+	}
+
 	recordSetResp, err := r.client.GetRecordSet(ctx, projectId, zoneId, recordSetId).Execute()
 	if err != nil {
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
```
rubenhoenle:
```suggestion
		var oapiErr *oapierror.GenericOpenAPIError
	    ok := errors.As(err, &oapiErr)
```

saves you the nolint comment ^^

stackit/internal/services/dns/recordset/resource.go
```diff
@@ -219,15 +221,27 @@ func (r *recordSetResource) Create(ctx context.Context, req resource.CreateReque
...
-		"project_id":    projectId,
-		"zone_id":       zoneId,
-		"record_set_id": *recordSetResp.Rrset.Id,
-	})
+	recordSetId := *recordSetResp.Rrset.Id
+	model.RecordSetId = types.StringValue(recordSetId)
+	model.Id = utils.BuildInternalTerraformId(projectId, zoneId, recordSetId)
+
+	// Set all unknown/null fields to null before saving state
+	if err := utils.SetModelFieldsToNull(ctx, &model); err != nil {
```
rubenhoenle: Sorry, but I just don't get why one would want to have this. What's the point of this?

stackit/internal/utils/utils.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"reflect"
```
rubenhoenle, line 10: https://principles.schwarz/principles/engineering-principles/go/avoid-using-reflection

Makefile
PatrickKoss: yeah that was a bit too ambitious. You know once you copy it from somewhere you always copy it :P
I replaced it with downloading from the releases which should be secure.

It is actually quite typically to download binaries that are needed to interact with the application (like linting, kubectl, kind, helm, etc) via scripts/make. In many stackit projects that is already the case. And there are also many opensource projects that do similar things like:
- https://github.com/cert-manager/cert-manager/blob/master/Makefile
- https://github.com/crossplane-contrib/provider-upjet-aws/blob/main/Makefile

I guess many ways solve the same problem. Currently my biggest problem is that I cannot lint locally since there are version diffs between my installed golangci lint and the one in the pipeline. Therefore I want to have a make command that runs the same version in the pipeline as in our local env. Some might say that is the shift left approach.

stackit/internal/services/dns/recordset/resource.go
PatrickKoss: you are right :D I had some weird client implementation and probably weird configuration where the ids were empty for some reason. In the current client they are always set so I removed the code blog entirely

stackit/internal/services/dns/recordset/resource.go
PatrickKoss: Thanks for the hint. Used the errors.As now

stackit/internal/services/dns/recordset/resource.go
PatrickKoss: It might be because of weird clients. Currently we only set project_id and zone_id in the state before waiting. This lead to the following error if the waiting is skipped which some clients want:

```
"error": "cannot get a terraform workspace for resource: cannot ensure tfstate file: cannot check whether the state is empty: cannot work with a non-string id: <nil>", "errorVerbose": "cannot work with a non-string id: <nil>
```

So I have set the id as well in the helper function. I think I observed an error in the past that the client got an error because some field in the state were "unknown". But I can no longer find this error message anymore. So currently I get:

```
apply failed: Provider produced inconsistent result after apply: When applying changes to stackit_dns_zone.example-zone, provider "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new value: .description: was cty.StringVal("Example DNS zone for demonstration"), but now null.

      This is a bug in the provider, which should be reported in the provider's own issue tracker.
      Provider produced inconsistent result after apply: When applying changes to stackit_dns_zone.example-zone, provider "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new value: .dns_name: was cty.StringVal("patrick.test.patrick.patrick"), but now null.

      This is a bug in the provider, which should be reported in the provider's own issue tracker.
      Provider produced inconsistent result after apply: When applying changes to stackit_dns_zone.example-zone, provider "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new value: .name: was cty.StringVal("example-zone"), but now null.

      This is a bug in the provider, which should be reported in the provider's own issue tracker.
      Provider produced inconsistent result after apply: When applying changes to stackit_dns_zone.example-zone, provider "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new value: .is_reverse_zone: was cty.False, but now null.

      This is a bug in the provider, which should be reported in the provider's own issue tracker.
      Provider produced inconsistent result after apply: When applying changes to stackit_dns_zone.example-zone, provider "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new value: .type: was cty.StringVal("primary"), but now null.

      This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

Because of it the client wants to destroy the resource:

```
"error": "cannot run plan: plan failed: Instance cannot be destroyed: Resource stackit_dns_zone.example-zone has lifecycle.prevent_destroy set, but the plan calls for this resource to be destroyed. To avoid this error and continue with the plan, either disable lifecycle.prevent_destroy or reduce the scope of the plan using the -target flag.", "errorVerbose": "plan failed: Instance cannot be destroyed: Resource stackit_dns_zone.example-zone has lifecycle.prevent_destroy set, but the plan calls for this resource to be destroyed. To avoid this error and continue with the plan, either disable lifecycle.prevent_destroy or reduce the scope of the plan using the -target flag.
```

For some reason if you set the fields to null instead of unknown the client accepts it and proceeds correctly. Maybe we need to take a look together into the topic. If you have some better ways to handle this case feel free to suggest :)

stackit/internal/utils/utils.go
PatrickKoss: its an avoid and not strictly dont :P 
As mentioned above we may need to set all fields in the model to null instead of unknown. I don´t write a function in every single resource that does that since it may be quite error prone as it is repetetive and every single a new api field is introduced we should not forget that we need to set the field to null as well.
Therefore I attempted to create one function that sets all fields of a model no null if they are unknown. So we can reuse it in all resources.
If you have a better idea how to achieve the goal feel free to suggest. Depending on the outcome of the discussion above it may not be needed as well.

stackit/internal/services/dns/recordset/resource.go
rubenhoenle: Just to make sure I don't mess things up here, what do you mean with `client`?

PatrickKoss: crossplane+upjet that then executes terraform cli commands

rubenhoenle: image

Messages like this are there for a reason by Terraform. You would break this behavior with this change.

PatrickKoss: I suspect that is a problem with complex objects, list of lists and list of complex objects in the utils function SetModelFieldsToNull.
I also tried adding the same logic as in zone to iaas network and added alot of unit tests to provoke the error and couldn´t reproduce. You can check it [here](https://github.com/PatrickKoss/terraform-provider-stackit/blob/feature/cp-enhancements/stackit/internal/services/iaas/network/resource_create_test.go) if you want.

Can you provide the input parameters so I can add unit tests for this case to verify if it happens in the implementation or not?
Additionally you can check with in your setup as well if the added functionality resolves the issue.

PatrickKoss: or do you imply that it is perfectly fine to have errors? because if we want to use upjet to generate a crossplane provider we cannot accept such error since it simply does not work :D

rubenhoenle: > or do you imply that it is perfectly fine to have errors?

Clear no.

> because if we want to use upjet to generate a crossplane provider we cannot accept such error since it simply does not work :D

Well, I guess it doesn't work because you modified the code of the terraform provider and didn't understand the impacts of your changes. 

I have to start from scratch here: Unknown values are a core concept of Terraform (see https://developer.hashicorp.com/terraform/plugin/framework/handling-data/terraform-concepts#unknown-values). Unknown values are important for Terraform to apply resources in the correct order, ...

But what does this mean for us? After a `terraform apply` run which creates a new resource, all fields of the resource must be set by the Terraform provider to a value or to null explicitly. If this isn't done for a field of the resource, you will get a message like this:

image

Whenever you get a message like this it's clear that this is a bug in the Terraform provider. And I'm going to lean myself out of the window here and say this doesn't happen for the `stackit_dns_record_set` resource on the main branch of our STACKIT Terraform provider repository. :smile: 

---

## Let me explain why

We create the resource on API side and then use the wait handler.

https://github.com/stackitcloud/terraform-provider-stackit/blob/b5f82e7de9056a237587a5d12b7ca7d6915155c0/stackit/internal/services/dns/recordset/resource.go#L215-L235

After the wait handler we use the `mapFields` function to map the API response to the Terraform state model.

https://github.com/stackitcloud/terraform-provider-stackit/blob/b5f82e7de9056a237587a5d12b7ca7d6915155c0/stackit/internal/services/dns/recordset/resource.go#L237-L248

Now comes the important part: Here is the section in the `mapFields` function, **which makes sure all fields of the resource get set to a value or null.** [1]



https://github.com/stackitcloud/terraform-provider-stackit/blob/b5f82e7de9056a237587a5d12b7ca7d6915155c0/stackit/internal/services/dns/recordset/resource.go#L432-L445

Well, and after that the model struct must be persisted in the Terraform state (this doesn't happen automatically):

https://github.com/stackitcloud/terraform-provider-stackit/blob/b5f82e7de9056a237587a5d12b7ca7d6915155c0/stackit/internal/services/dns/recordset/resource.go#L243-L248

To sum it up, here's what happens in the main branch implementation of this resource:

1. Create request for the API resource
2. (Write id fields to the state in case anything goes wrong during the wait handler)
3. Wait handler to wait for creation of the API resource to complete
4. Map API response to Terraform resource model struct (`mapFields`)
5. Persist the Terraform model struct of the resource in the Terraform state

---

## Now to your changes 

Now to your changes and why it's not working (without setting all fields to null using your new reflection-powered util func):

In your `func (r *recordSetResource) Create(...) ...` implementation...
- You also do the Create request for the API resource (see no. 1 above)
- You write the id fields to the state (see no 2. above)
- And then you jump out of the `Create` implementation of the Terraform resource prematurely with the code below.

```golang
	if !utils.ShouldWait() {
		tflog.Info(ctx, "Skipping wait; async mode for Crossplane/Upjet")
		return
	}
```

The problem is: This doesn't only skip the wait handler (no. 3 above), but also the `mapFields` func call (no. 4 above) which (as said) sets explicitly all values to a value or null. 

Again, you just skip this. This is a core part of the resource implementation. You don't call it. That's why Terraform complains about unknown values. Terraform says this is a bug in the provider implementation, and it's correct. 

But it's sadly not a bug in **our** implementation on the main branch, but in **your** implementation.

You circumvent this problem by setting all fields of the Terraform resource state model explicitly to null by using your new util func. This circumvents the problem (Terraform doesn't complain anymore about unknown values), but it doesn't really fix the problem (at least not in a clean way). 

In fact setting all fields of the Terraform resource model struct to null circumvents existing checks of Terraform which we want to take advantage of during our resource implementations (at least for pure Terraform usage, without thinking of crossplane here).

---

[1] Btw, if you forget to set one field of the Terraform resource model struct to a value of null here during the implementation of the Terraform resource you will also get exactly the error `After the apply operation, the provider still indicated an unknown value...` from above. This is what I consider a terraform feature. As said, unknown values are a concept of Terraform

PatrickKoss: Thanks for the detailed explanation. It covers well my observations. I think we are actually on two sides of the same coin. 

Let´s take a step and start with the requirements for the create again, then I share my observations during testing and then check different alternatives.

## Requirements
- Have idempotency. If I apply a resource and somehow fail right after the api call (for example due to timeouts, context cancels, random api errors in the wait handler) I want the resource to be in the state and use a Read to fill the model in the next apply. There should be no state drifts or replacements of the resource created in the first apply.
- Have a way to return right after the creation of the resource without waiting. This comes from upjet/crossplane (therefore the skip method with the log that it is intended to only use by this tool). Upjet needs the ids of the resources quite fast to persist them in kubernetes custom resources (the database so to say). Because the resource is only known once stored in the custom resource. It does hold the terraform state in a file only temporary. During applies the state is constructed with the custom resource. The problem with the waiting here is that the controller executing terraform can restart in any point in time. Since the wait handler can take quite a bit of time we risk creating the same resource twice. Therefore the early return. And it is completely fine for the tool since it executes a Read directly after the return. Every 10 min it queries the state of the cloud resource as well. So eventually it will reach the point where the cloud resource and the custom resource have the same data. That´s the standard kubernetes reconciling mechanism. 
- (optional) have a common way to achieve idempotency in every single resource. We should have a rock solid way without much custom implementation as it is error prone to do it for each resource.

## Code Walkthrough, Testing and Observations

We already recognized that we need to set partial states in the terraform state. That´s why the following code already exists in the main branch:

```
	utils.SetAndLogStateFields(ctx, &resp.Diagnostics, &resp.State, map[string]any{
		"project_id":    projectId,
		"zone_id":       zoneId,
		"record_set_id": *recordSetResp.Rrset.Id,
	})
		return
	}
```
https://github.com/PatrickKoss/terraform-provider-stackit/blob/main/stackit/internal/services/dns/recordset/resource.go#L222-L229

In my tests I have setup a terraform resource (in this case mariadb with the same function as mariadb takes way longer to create and dns is super fast. So please don´t be confused about the resource we are still talking about the same code)

```
resource "stackit_mariadb_instance" "example_maria_db" {
  name       = "example-mariadb"
  plan_name  = "stackit-mariadb-1.4.10-single"
  project_id = "xxx"
  version    = "10.6"
}
```

Then I applied and once the wait handler started and I saw mariadb in creating state in the portal I canceled the apply to simulate random failures as mentioned above. Then I reapplied and got the error: `stackit_mariadb_instance.example_maria_db is tainted, so must be replaced`.

That´s when I recognized that setting ids is not enough and we need to include the fields in the resource as well (name, plane_name,version). So I changed the code to:

```
	utils.SetAndLogStateFields(ctx, &resp.Diagnostics, &resp.State, map[string]interface{}{
		"project_id":  projectId,
		"instance_id": model.InstanceId.ValueString(),
		"id":          model.Id.ValueString(),
		"name":        model.Name.ValueString(),
		"plan_name":   model.PlanName.ValueString(),
		"version":     model.Version.ValueString(),
		"plan_id":     model.PlanId.ValueString(),
	})
		return
	}
```

and that almost worked. We also should not log and error in the wait handler as it messes up terraform and result in non idempotent behaviour:

```
	waitResp, err := wait.CreateInstanceWaitHandler(ctx, r.client, projectId, instanceId).WaitWithContext(ctx)
	if err != nil {
		tflog.Warn(ctx, fmt.Sprintf("Instance creation waiting failed: %v. The instance was created but waiting for ready state was interrupted. State will be refreshed on next apply.", err))
		return
	}
```

And that works perfectly fine in the case of create/cancel/reapply. Now there are no state drift and the resource stays as it is. 

Now I went a step further and wrote unit tests for the behaviour. So we can really verify that it works how we think it works. 

```
	// Verify that Read successfully populated all fields from the API
	var stateAfterRead Model
	diags = readResp.State.Get(tc.Ctx, &stateAfterRead)
	require.False(t, diags.HasError(), "Expected no errors reading state after Read")

	// Verify all fields are now complete after successful Read (prevents state drift)
	require.Equal(t, instanceId, stateAfterRead.InstanceId.ValueString())
	require.Equal(t, fmt.Sprintf("%s,%s", projectId, instanceId), stateAfterRead.Id.ValueString())
	require.Equal(t, projectId, stateAfterRead.ProjectId.ValueString())
	require.Equal(t, instanceName, stateAfterRead.Name.ValueString())
	require.Equal(t, planId, stateAfterRead.PlanId.ValueString())
	require.Equal(t, planName, stateAfterRead.PlanName.ValueString())
	require.Equal(t, version, stateAfterRead.Version.ValueString())

	// CRITICAL: Verify fields that were NULL after Create are now populated
	// This prevents Terraform state drift on the next apply
	require.False(t, stateAfterRead.DashboardUrl.IsNull(), "DashboardUrl must be populated by Read to prevent state drift")
	require.Equal(t, dashboardUrl, stateAfterRead.DashboardUrl.ValueString())
	require.False(t, stateAfterRead.CfGuid.IsNull(), "CfGuid must be populated by Read to prevent state drift")
	require.False(t, stateAfterRead.ImageUrl.IsNull(), "ImageUrl must be populated by Read to prevent state drift")
```
https://github.com/PatrickKoss/terraform-provider-stackit/blob/feature/cp-enhancements/stackit/internal/services/mariadb/instance/resource_create_test.go#L16-L157

The unit test covers the manual test create/cancel/read. Note that setting the partial state actually leads to null fields while reading the state again. Then I inserted the `utils.SetModelFieldsToNull` instead of  `utils.SetAndLogStateFields` and the test(s) were equally successful. This lead me to the assumption we are actually on two different sides of the same coin (different code but same outcome). Not setting fields in the state leads to null values while setting them to null explicitly also result in reading out null values. So we probably found out multiple ways to solve the idempotency problem. More in the alternatives.

Second the early exit is this code:

```
	if !utils.ShouldWait() {
		tflog.Info(ctx, "Skipping wait; async mode for Crossplane/Upjet")
		return
	}
```

Note this function is only executed if an environment variable is set to "true". If the variable is not set or to any other value than "true" we would continue with the wait handler. Not pretty but we somehow need to cover the requirement since the tool works as it works.

## Alternatives/Conclusion

I think there is no real discussion about the early return but if there is feel free to suggest something.

The more interesting point is the idempotency part.

- As we already saw in the tests we need to set the ids and the fields specified in the resource to avoid state drift and resource recreation. One approach could be like the current one in the main branch but make it a bit more abstract. We can construct the map based on the map. Similiar to the proposed implementation `utils.SetModelFieldsToNull` we can iterate over the models attributes with reflection magic check for non null/unknown fields and use the tags (tfsdk) of the model as keys for the map and the value of the attribute of the model. This should result in the map we want to store as partial state in the terraform state.
- Similiar to the first approach we can go the reverse approach and have the model already set and then set all fields to null that are unknown. That´s also the proposed approach. You highlighted correctly that it might not be the best idea to use the same model that is used after the wait handler as we also want to verify the behaviour of the map function after the wait handler. Means we should do a deepcopy and set the model fields on this deepcopy to null and also save the deepcopy in the struct.
- One last approach that I could came up with is the construction of the map with a lot of if-conditions in the resource without any reflection magic. That is the least preferred option as it requires implementing it in every resource and is error prone since we may miss fields. (That´s what I mean in the third requirement)

So, what do you think? Do you have other testing experiences? Which direction should we go?

PatrickKoss: updated the code. Catched wait errors. Using now another model to save the minimal state which preserve the upstream logic and still fixes the idempotency bugs

PR #1068, opened 2025-11-21
feat(access-token): add ephemeral access-token resource

h3adex:
## Description

Internal Issue: STACKITTPR-353

## Checklist

- [x] Issue was linked above

### DIFF
--- a/docs/ephemeral-resources/access_token.md
+++ b/docs/ephemeral-resources/access_token.md
@@ -38,30 +38,6 @@ locals {
   }
 }
 
-# Docs: https://registry.terraform.io/providers/magodo/restful/latest
-provider "restful" {
-  base_url = local.stackit_api_base_url
-
-  security = {
-    http = {
-      token = {
-        token = ephemeral.stackit_access_token.example.access_token
-      }
-    }
-  }
-}
-
-resource "restful_resource" "public_ip_restful" {
-  path = local.public_ip_path
-  body = local.public_ip_payload
-
-  read_path     = "$(path)/$(body.id)"
-  update_path   = "$(path)/$(body.id)"
-  update_method = "PATCH"
-  delete_path   = "$(path)/$(body.id)"
-  delete_method = "DELETE"
-}
-
 # Docs: https://registry.terraform.io/providers/Mastercard/restapi/latest
 provider "restapi" {
   uri                  = local.stackit_api_base_url
--- a/examples/ephemeral-resources/stackit_access_token/ephemeral-resource.tf
+++ b/examples/ephemeral-resources/stackit_access_token/ephemeral-resource.tf
@@ -17,30 +17,6 @@ locals {
   }
 }
 
-# Docs: https://registry.terraform.io/providers/magodo/restful/latest
-provider "restful" {
-  base_url = local.stackit_api_base_url
-
-  security = {
-    http = {
-      token = {
-        token = ephemeral.stackit_access_token.example.access_token
-      }
-    }
-  }
-}
-
-resource "restful_resource" "public_ip_restful" {
-  path = local.public_ip_path
-  body = local.public_ip_payload
-
-  read_path     = "$(path)/$(body.id)"
-  update_path   = "$(path)/$(body.id)"
-  update_method = "PATCH"
-  delete_path   = "$(path)/$(body.id)"
-  delete_method = "DELETE"
-}
-
 # Docs: https://registry.terraform.io/providers/Mastercard/restapi/latest
 provider "restapi" {
   uri                  = local.stackit_api_base_url


stackit/internal/core/core.go
```diff
@@ -28,6 +28,12 @@ const (
 type ProviderData struct {
 	RoundTripper        http.RoundTripper
 	ServiceAccountEmail string // Deprecated: ServiceAccountEmail is not required and will be removed after 12th June 2025.
+
+	PrivateKey            string
```
rubenhoenle, line 33: Could we have a seperate struct for that? E.g. by using https://gobyexample.com/struct-embedding

```golang
type EphermalProviderData struct {
    ProviderData
    
    PrivateKey            string
    PrivateKeyPath        string
	ServiceAccountKey     string
	ServiceAccountKeyPath string
}
```

The fields like `PrivateKey`, ... don't belong into regular resources and datasources. You don't even set them there (what is good of course :sweat_smile:). So some abstraction won't hurt here to keep things clean.

stackit/internal/services/access_token/ephemeral_access_token.go
```diff
@@ -0,0 +1,200 @@
...
+		return
+	}
+
+	accessToken, err := client.GetAccessToken()
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Access token generation failed", fmt.Sprintf("Error generating access token: %v", err))
+		return
+	}
+
+	ctx = tflog.SetField(ctx, "access_token", accessToken)
```
rubenhoenle: **This is a HUGE no no, please don't ever (!!) log credentials**

stackit/internal/services/access_token/ephemeral_resource.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	return &key, diags
+}
+
+// resolvePrivateKey determines the private key value using env, conf, fallbacks.
+func resolvePrivateKey(ctx context.Context, cfgValue, cfgPath string, key *clients.ServiceAccountKeyResponse) (string, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	env := os.Getenv("STACKIT_PRIVATE_KEY")
```
rubenhoenle: shouldn't this be done already by the provider logic? I'm not willing to read env variables here...

stackit/internal/services/access_token/ephemeral_resource.go
```diff
@@ -0,0 +1,200 @@
...
+	ctx = tflog.SetField(ctx, "access_token", accessToken)
+	model.AccessToken = types.StringValue(accessToken)
+	resp.Diagnostics.Append(resp.Result.Set(ctx, model)...)
+}
+
+// loadServiceAccountKey loads the service account key based on env vars, or fallback to provider config.
+func loadServiceAccountKey(ctx context.Context, cfgValue, cfgPath string) (*clients.ServiceAccountKeyResponse, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	env := os.Getenv("STACKIT_SERVICE_ACCOUNT_KEY")
```
rubenhoenle: shouldn't this be done already by the provider logic? I'm not willing to read env variables here...

stackit/internal/services/access_token/ephemeral_resource.go
rubenhoenle: If it's not done there, it belongs there. This doesn't belong into some resource implementation for sure.

stackit/internal/services/access_token/ephemeral_access_token.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/ephemeral/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/stackitcloud/stackit-sdk-go/core/clients"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/conversion"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/core"
+)
+
+// #nosec G101 tokenUrl is a public endpoint, not a hardcoded credential
+const tokenUrl = "https://service-account.api.stackit.cloud/token"
```
rubenhoenle: No need to duplicate, this is already the default

https://github.com/stackitcloud/stackit-sdk-go/blob/ca0bcc279e1be97bcb1a4b4377091f26c737483e/core/clients/key_flow.go#L134-L136

rubenhoenle: Other question, what's with custom endpoints?

stackit/internal/services/access_token/ephemeral_resource.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	rt, err := auth.KeyAuth(&cfg)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Access token generation failed", fmt.Sprintf("Failed to initialize authentication: %v", err))
+		return
+	}
+
+	// Type assert to access token functionality
+	client, ok := rt.(*clients.KeyFlow)
```
h3adex: Implementing all the service account key and environment variable logic might not have been the best idea in hindsight. What do you think about leveraging the existing `auth.KeyAuth` method and type asserting to `*client.KeyFlow` instead?  

Right now, the Go SDK only returns an `http.RoundTripper`, but since the underlying type is `KeyFlow`, we could safely assert it and access the full client. This would significantly simplify and clean up the entire resource implementation.

stackit/internal/services/access_token/ephemeral_resource.go
```diff
@@ -0,0 +1,111 @@
...
+func NewAccessTokenEphemeralResource() ephemeral.EphemeralResource {
+	return &accessTokenEphemeralResource{}
+}
+
+type accessTokenEphemeralResource struct {
+	serviceAccountKeyPath string
+	serviceAccountKey     string
+	privateKeyPath        string
+	privateKey            string
+	tokenCustomEndpoint   string
```
rubenhoenle: Why is this defined on resource level? All other custom endpoints are defined on provider configuration level. Pretty confusing from a user perspective.

stackit/internal/services/access_token/ephemeral_resource.go
```diff
@@ -0,0 +1,111 @@
...
+	AccessToken types.String `tfsdk:"access_token"`
+}
+
+func (e *accessTokenEphemeralResource) Metadata(_ context.Context, req ephemeral.MetadataRequest, resp *ephemeral.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_access_token"
+}
+
+func (e *accessTokenEphemeralResource) Schema(_ context.Context, _ ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "STACKIT Access Token ephemeral resource schema.",
```
rubenhoenle: Missing information for the docs:

- How long is the token valid?
- What happens when it expires?

stackit/internal/conversion/conversion.go
```diff
@@ -183,3 +183,18 @@ func ParseProviderData(ctx context.Context, providerData any, diags *diag.Diagno
 	}
 	return stackitProviderData, true
 }
+
+// TODO: write tests
```
rubenhoenle: resolve Todo

examples/ephemeral-resources/stackit_access_token/ephemeral-resource.tf
```diff
@@ -0,0 +1,16 @@
...
+  uri                  = "https://iaas.api.eu01.stackit.cloud"
+  write_returns_object = true
+
+  headers = {
+    "Authorization" = "Bearer ${ephemeral.stackit_access_token.example.access_token}"
+  }
+
+  create_method  = "GET"
+  update_method  = "GET"
+  destroy_method = "GET"
```
rubenhoenle:
```suggestion
  destroy_method = "DELETE"
```

Or is it really `GET`? :joy:

rubenhoenle: Same for update btw, `GET` doesn't seem right here

stackit/internal/services/access_token/ephemeral_resource.go
h3adex: We could clean it up like this:

```go
type accessTokenEphemeralResource struct {
	keyAuthConfig config.Configuration
}

func (e *accessTokenEphemeralResource) Configure(ctx context.Context, req ephemeral.ConfigureRequest, resp *ephemeral.ConfigureResponse) {
	providerData, ok := conversion.ParseEphemeralProviderData(ctx, req.ProviderData, &resp.Diagnostics)
	if !ok {
		return
	}

	e.keyAuthConfig = config.Configuration{
		ServiceAccountKey:     providerData.ServiceAccountKey,
		ServiceAccountKeyPath: providerData.ServiceAccountKeyPath,
		PrivateKeyPath:        providerData.PrivateKey,
		PrivateKey:            providerData.PrivateKeyPath,
		TokenCustomUrl:        providerData.TokenCustomEndpoint,
	}
}
```

stackit/provider.go
```diff
@@ -465,6 +467,15 @@ func (p *Provider) Configure(ctx context.Context, req provider.ConfigureRequest,
 	resp.DataSourceData = providerData
 	resp.ResourceData = providerData
 
+	// Copy service account, private key credentials and custom-token endpoint to support ephemeral access token generation
+	var ephemeralProviderData core.EphemeralProviderData
+	setStringField(providerConfig.ServiceAccountKey, func(v string) { ephemeralProviderData.ServiceAccountKey = v })
+	setStringField(providerConfig.ServiceAccountKeyPath, func(v string) { ephemeralProviderData.ServiceAccountKeyPath = v })
+	setStringField(providerConfig.PrivateKey, func(v string) { ephemeralProviderData.PrivateKey = v })
+	setStringField(providerConfig.PrivateKeyPath, func(v string) { ephemeralProviderData.PrivateKeyPath = v })
+	setStringField(providerConfig.TokenCustomEndpoint, func(v string) { ephemeralProviderData.TokenCustomEndpoint = v })
```
rubenhoenle, line 484: either you remove the `ProviderData` type embedding from the `EphemeralProviderData` because you don't seem to need it or you set the values here. Currently this is pretty useless.

docs/ephemeral-resources/access_token.md
```diff
@@ -0,0 +1,57 @@
...
+Ephemeral resource that generates a short-lived STACKIT access token (JWT) using a service account key. A new token is generated each time the resource is evaluated, and it remains consistent for the duration of a Terraform operation. If a private key is not explicitly provided, the provider attempts to extract it from the service account key instead. Token generation logic prioritizes environment variables first, followed by provider configuration. Access tokens generated from service account keys expire after 60 minutes.
+
+## Example Usage
+
+```terraform
+ephemeral "stackit_access_token" "example" {}
+
+// https://registry.terraform.io/providers/Mastercard/restapi/latest/docs
+provider "restapi" {
+  uri                  = "https://iaas.api.eu01.stackit.cloud/"
```
rubenhoenle:
```suggestion
  uri                  = "https://iaas.api.stackit.cloud/"
```

For IaaS v2 it's without the `.eu01` part, see https://docs.api.stackit.cloud/documentation/iaas/version/v2 for reference

docs/ephemeral-resources/access_token.md
```diff
@@ -0,0 +1,57 @@
...
+---
+
+# stackit_access_token (Ephemeral Resource)
+
+Ephemeral resource that generates a short-lived STACKIT access token (JWT) using a service account key. A new token is generated each time the resource is evaluated, and it remains consistent for the duration of a Terraform operation. If a private key is not explicitly provided, the provider attempts to extract it from the service account key instead. Token generation logic prioritizes environment variables first, followed by provider configuration. Access tokens generated from service account keys expire after 60 minutes.
+
+## Example Usage
+
+```terraform
+ephemeral "stackit_access_token" "example" {}
```
rubenhoenle, line 28:
```hcl
terraform {
  required_providers {
    stackit = {
      source = "registry.terraform.io/stackitcloud/stackit"
    }
  }
}

provider "stackit" {
  default_region        = "eu01"
  service_account_key_path = "/path/to/sa_key.json"
}

ephemeral "stackit_access_token" "example" {}

...
```

Maybe add a simple provider block for completeness :sweat_smile:

stackit/internal/services/access_token/ephemeral_resource.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (e *accessTokenEphemeralResource) Open(ctx context.Context, req ephemeral.OpenRequest, resp *ephemeral.OpenResponse) {
+	var model ephemeralTokenModel
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &model)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	rt, err := auth.KeyAuth(&e.keyAuthConfig)
```
rubenhoenle:
```suggestion
	roundTripper, err := auth.KeyAuth(&e.keyAuthConfig)
```

Please use speaking variable names

stackit/internal/services/access_token/ephemeral_resource.go
rubenhoenle: After taking a deeper look at `auth.KeyAuth` from the SDK: lgtm

rubenhoenle: But: Could we please have a unit tests which mocks the API call(s) and tests the whole process? To make sure the type assertion doesn't break on future updates of the SDK core module?

It might help to extract the logic for the access token into a seperate func beforehand which might look like this:

```golang
func (e *accessTokenEphemeralResource) Open(ctx context.Context, req ephemeral.OpenRequest, resp *ephemeral.OpenResponse) {
	var model ephemeralTokenModel

	resp.Diagnostics.Append(req.Config.Get(ctx, &model)...)
		return
	}

	accessToken, err := getAccessToken(ctx, &e.keyAuthConfig)
	if err != nil {
		core.LogAndAddError(ctx, &resp.Diagnostics, "Access token generation failed", err.Error())
		return
	}

	model.AccessToken = types.StringValue(accessToken)
	resp.Diagnostics.Append(resp.Result.Set(ctx, model)...)
}

func getAccessToken(ctx context.Context, keyAuthConfig *config.Configuration) (string, error) {
	rt, err := auth.KeyAuth(keyAuthConfig)
	if err != nil {
		return "", fmt.Errorf("failed to initialize authentication: %v", err)
	}

	// Type assert to access token functionality
	client, ok := rt.(*clients.KeyFlow)
	if !ok {
		return "", fmt.Errorf("internal error: expected *clients.KeyFlow, but received a different implementation of http.RoundTripper")
	}

	// Retrieve the access token
	accessToken, err := client.GetAccessToken()
	if err != nil {
		return "", fmt.Errorf("error obtaining access token: %v", err)
	}

	return accessToken, nil
}
```

stackit/internal/services/access_token/ephemeral_resource.go
```diff
@@ -0,0 +1,100 @@
...
+type accessTokenEphemeralResource struct {
+	keyAuthConfig config.Configuration
+}
+
+func (e *accessTokenEphemeralResource) Configure(ctx context.Context, req ephemeral.ConfigureRequest, resp *ephemeral.ConfigureResponse) {
+	providerData, ok := conversion.ParseEphemeralProviderData(ctx, req.ProviderData, &resp.Diagnostics)
+	if !ok {
+		return
+	}
+
```
marceljk, line 36: I think it's better to start with this resource as beta or even set it as experimental, to have the option to easier change things, if the resource has some unexpected issued

stackit/internal/services/access_token/ephemeral_resource.go
h3adex: Good idea. Just pushed something which goes into that direction.

stackit/internal/services/access_token/ephemeral_resource.go
h3adex: Unfortunately, the `features.CheckBetaResourcesEnabled` function expects a `core.ProviderData` object. I've pushed code that handles this by creating a new `core.ProviderData` instance with the required field, like so:

```go
features.CheckBetaResourcesEnabled(
    ctx,
    &core.ProviderData{EnableBetaResources: ephemeralProviderData.EnableBetaResources},
    &resp.Diagnostics,
    "stackit_access_token", "ephemeral_resource",
)
    return
}
```

Is that fine?

h3adex:
> One general question I still have: What happens when the user authenticates via token flow?
> 
> ```hcl
> provider "stackit" {
>   default_region        = "eu01"
>   service_account_token = var.service_account_token
> }
> ```

Will look into that tomorrow.

stackit/internal/services/access_token/ephemeral_resource.go
rubenhoenle: As said, I guess the type embedding of `core.ProviderData` into `core.EphemeralProviderData` is the way to go for me...

With your solution you're duplicating fields across these structs...

stackit/internal/services/access_token/ephemeral_resource_test.go
```diff
@@ -0,0 +1,253 @@
...
+	tests := []struct {
+		description string
+		setupEnv    func()
+		cleanupEnv  func()
+		cfgFactory  func() *config.Configuration
+		expectError bool
+		expected    string
+	}{
+		{
+			description: "should_return_token_when_service_account_key_passed_by_value",
```
rubenhoenle:
```suggestion
			description: "should return token when service account key passed by value",
```

Why the underscores? Makes it only hard to read

stackit/internal/services/access_token/ephemeral_resource.go
h3adex: I’ve done that now, but this still doesn’t solve the issue that `features.CheckBetaResourcesEnabled` expects a value of type `core.ProviderData`. I don’t see a clean way to fix this without changing the function to accept only a boolean, which would then require updates to all resources and data sources. Given that, it might be easier to stick with the current approach for now. Lmk if you want me to change the func to accept a boolean and update all resources/data sources.

rubenhoenle: Try this one :wink: 

```golang
features.CheckBetaResourcesEnabled(
		ctx,
		&ephemeralProviderData.ProviderData,
		&resp.Diagnostics,
		"stackit_access_token", "ephemeral_resource",
)
```

rubenhoenle:
> Error message when you try to generate a access token without a valid config
> Screenshot 2025-11-28 at 09 38 16

For me it must be at least stated in the docs which requirements exist to use this resource: User must use keyflow for authentication. Or am I wrong here?

stackit/provider.go
rubenhoenle: @h3adex **this is still not resolved**. The following line is missing here:

```golang
ephemeralProviderData.ProviderData = providerData
```

stackit/internal/services/access_token/ephemeral_resource.go
h3adex: 🤯

stackit/provider.go
h3adex: Thank you! Fixed.

stackit/internal/conversion/conversion.go
```diff
@@ -183,3 +183,17 @@ func ParseProviderData(ctx context.Context, providerData any, diags *diag.Diagno
...
+
+func ParseEphemeralProviderData(ctx context.Context, providerData any, diags *diag.Diagnostics) (core.EphemeralProviderData, bool) {
+	// Prevent panic if the provider has not been configured.
+	if providerData == nil {
+		return core.EphemeralProviderData{}, false
+	}
+
+	stackitProviderData, ok := providerData.(core.EphemeralProviderData)
+	if !ok {
+		core.LogAndAddError(ctx, diags, "Error configuring API client", fmt.Sprintf("Expected configure type stackit.ProviderData, got %T", providerData))
```
rubenhoenle: Again, you're potentially leaking credentials here. Stop it. `providerData` could contain tokens and keys.

rubenhoenle: and...

```suggestion
		core.LogAndAddError(ctx, diags, "Error configuring API client", fmt.Sprintf("Expected configure type core.EphemeralProviderData, got %T", providerData))
```

PR #1113, opened 2026-01-08
STACKITRCO-187 - Add option iaas API param agent

aeter:
Adds a terraform `stackit_server` option for the iaas ( _create server_ ) API param: `"agent": {"provisioned": true}`

ref STACKITRCO-187

ref: https://github.com/stackitcloud/stackit-cli/pull/1213

---

Tests:
    * ran `make fmt`, `make generate-docs`, `make lint`

    * ran unit tests
    ```
    [~/terraform-provider-stackit] go test stackit/internal/services/iaas/server/*
    ok      command-line-arguments  15.005s
    [~/terraform-provider-stackit] make test
    ...
    ok      github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/iaas/server       15.006s  coverage: 33.0% of statements
    ...
    [~/terraform-provider-stackit]
    ```

    * Tested: with a locally-configured terraform, by adding, changing, deleting `agent`-related parts of a tf config.
## Description

relates to  STACKITRCO-187

## Checklist

- [x] Issue was linked above

### DIFF
--- a/docs/resources/server.md
+++ b/docs/resources/server.md
@@ -404,6 +404,7 @@ import {
 ### Optional
 
 - `affinity_group` (String) The affinity group the server is assigned to.
+- `agent` (Attributes) The STACKIT Server Agent configured for the server (see [below for nested schema](#nestedatt--agent))
 - `availability_zone` (String) The availability zone of the server.
 - `boot_volume` (Attributes) The boot volume for the server (see [below for nested schema](#nestedatt--boot_volume))
 - `desired_status` (String) The desired status of the server resource. Possible values are: `active`, `inactive`, `deallocated`.
@@ -421,6 +422,18 @@ import {
 - `launched_at` (String) Date-time when the server was launched
 - `server_id` (String) The server ID.
 - `updated_at` (String) Date-time when the server was updated
+
+<a id="nestedatt--agent"></a>
+### Nested Schema for `agent`
+
+Optional:
+
+- `provisioning_policy` (String) Agent provisioning policy: `ALWAYS`, `NEVER`, or `INHERIT`. `INHERIT` follows the image default value.
+
+Read-Only:
+
+- `provisioned` (Boolean) Whether a STACKIT Server Agent is provisioned at the server
+
 
 <a id="nestedatt--boot_volume"></a>
 ### Nested Schema for `boot_volume`
--- a/stackit/internal/services/iaas/iaas_acc_test.go
+++ b/stackit/internal/services/iaas/iaas_acc_test.go
@@ -149,6 +149,7 @@ var testConfigServerVarsMax = config.Variables{
 	"service_account_mail": config.StringVariable(testutil.TestProjectServiceAccountEmail),
 	"public_key":           config.StringVariable(keypairPublicKey),
 	"desired_status":       config.StringVariable("active"),
+	"agent_policy":         config.StringVariable("ALWAYS"),
 }
 
 var testConfigServerVarsMaxUpdated = func() config.Variables {
@@ -158,6 +159,7 @@ var testConfigServerVarsMaxUpdated = func() config.Variables {
 	updatedConfig["machine_type"] = config.StringVariable("t1.2")
 	updatedConfig["label"] = config.StringVariable("updated")
 	updatedConfig["desired_status"] = config.StringVariable("inactive")
+	updatedConfig["agent_policy"] = config.StringVariable("NEVER")
 	return updatedConfig
 }()
 
@@ -2223,6 +2225,8 @@ func TestAccServerMin(t *testing.T) {
 					resource.TestCheckResourceAttrSet("stackit_server.server", "created_at"),
 					resource.TestCheckResourceAttrSet("stackit_server.server", "launched_at"),
 					resource.TestCheckResourceAttrSet("stackit_server.server", "updated_at"),
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioning_policy", "INHERIT"),
+					resource.TestCheckNoResourceAttr("stackit_server.server", "agent.provisioned"),
 				),
 			},
 			// Data source
@@ -2275,6 +2279,7 @@ func TestAccServerMin(t *testing.T) {
 					resource.TestCheckResourceAttrSet("data.stackit_server.server", "created_at"),
 					resource.TestCheckResourceAttrSet("data.stackit_server.server", "launched_at"),
 					resource.TestCheckResourceAttrSet("data.stackit_server.server", "updated_at"),
+					resource.TestCheckNoResourceAttr("data.stackit_server.server", "agent.provisioned"),
 				),
 			},
 			// Import
@@ -2328,6 +2333,8 @@ func TestAccServerMin(t *testing.T) {
 					resource.TestCheckResourceAttrSet("stackit_server.server", "created_at"),
 					resource.TestCheckResourceAttrSet("stackit_server.server", "launched_at"),
 					resource.TestCheckResourceAttrSet("stackit_server.server", "updated_at"),
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioning_policy", "INHERIT"),
+					resource.TestCheckNoResourceAttr("stackit_server.server", "agent.provisioned"),
 				),
 			},
 			// Deletion is done by the framework implicitly
@@ -2351,6 +2358,10 @@ func TestAccServerMax(t *testing.T) {
 					resource.TestCheckResourceAttr("stackit_affinity_group.affinity_group", "name", testutil.ConvertConfigVariable(testConfigServerVarsMax["name_not_updated"])),
 					resource.TestCheckResourceAttr("stackit_affinity_group.affinity_group", "policy", testutil.ConvertConfigVariable(testConfigServerVarsMax["policy"])),
 					resource.TestCheckResourceAttrSet("stackit_affinity_group.affinity_group", "affinity_group_id"),
+
+					// Agent
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioning_policy", testutil.ConvertConfigVariable(testConfigServerVarsMax["agent_policy"])),
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioned", "true"),
 
 					// Volume base
 					resource.TestCheckResourceAttr("stackit_volume.base_volume", "project_id", testutil.ConvertConfigVariable(testConfigServerVarsMax["project_id"])),
@@ -2500,6 +2511,7 @@ func TestAccServerMax(t *testing.T) {
 						"stackit_key_pair.key_pair", "name",
 						"data.stackit_server.server", "keypair_name",
 					),
+					resource.TestCheckResourceAttr("data.stackit_server.server", "agent.provisioned", "true"),
 					// All network interface which was are attached appear here
 					resource.TestCheckResourceAttr("data.stackit_server.server", "network_interfaces.#", "2"),
 					resource.TestCheckTypeSetElemAttrPair(
@@ -2725,7 +2737,7 @@ func TestAccServerMax(t *testing.T) {
 				},
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"boot_volume", "desired_status", "network_interfaces"}, // Field is not mapped as it is only relevant on creation
+				ImportStateVerifyIgnore: []string{"boot_volume", "desired_status", "network_interfaces", "agent"}, // Field is not mapped as it is only relevant on creation
 			},
 			// Update
 			{
@@ -2737,6 +2749,10 @@ func TestAccServerMax(t *testing.T) {
 					resource.TestCheckResourceAttr("stackit_affinity_group.affinity_group", "name", testutil.ConvertConfigVariable(testConfigServerVarsMaxUpdated["name_not_updated"])),
 					resource.TestCheckResourceAttr("stackit_affinity_group.affinity_group", "policy", testutil.ConvertConfigVariable(testConfigServerVarsMaxUpdated["policy"])),
 					resource.TestCheckResourceAttrSet("stackit_affinity_group.affinity_group", "affinity_group_id"),
+
+					// Agent
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioning_policy", testutil.ConvertConfigVariable(testConfigServerVarsMaxUpdated["agent_policy"])),
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioned", "false"),
 
 					// Volume base
 					resource.TestCheckResourceAttr("stackit_volume.base_volume", "project_id", testutil.ConvertConfigVariable(testConfigServerVarsMaxUpdated["project_id"])),
@@ -2842,6 +2858,10 @@ func TestAccServerMax(t *testing.T) {
 					resource.TestCheckResourceAttr("stackit_affinity_group.affinity_group", "name", testutil.ConvertConfigVariable(testConfigServerVarsMaxUpdatedDesiredStatus["name_not_updated"])),
 					resource.TestCheckResourceAttr("stackit_affinity_group.affinity_group", "policy", testutil.ConvertConfigVariable(testConfigServerVarsMaxUpdatedDesiredStatus["policy"])),
 					resource.TestCheckResourceAttrSet("stackit_affinity_group.affinity_group", "affinity_group_id"),
+
+					// Agent
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioning_policy", testutil.ConvertConfigVariable(testConfigServerVarsMaxUpdatedDesiredStatus["agent_policy"])),
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioned", "false"),
 
 					// Volume base
 					resource.TestCheckResourceAttr("stackit_volume.base_volume", "project_id", testutil.ConvertConfigVariable(testConfigServerVarsMaxUpdatedDesiredStatus["project_id"])),
--- a/stackit/internal/services/iaas/server/datasource.go
+++ b/stackit/internal/services/iaas/server/datasource.go
@@ -35,6 +35,7 @@ type DataSourceModel struct {
 	ServerId          types.String `tfsdk:"server_id"`
 	MachineType       types.String `tfsdk:"machine_type"`
 	Name              types.String `tfsdk:"name"`
+	Agent             types.Object `tfsdk:"agent"`
 	AvailabilityZone  types.String `tfsdk:"availability_zone"`
 	BootVolume        types.Object `tfsdk:"boot_volume"`
 	ImageId           types.String `tfsdk:"image_id"`
@@ -51,6 +52,10 @@ type DataSourceModel struct {
 var bootVolumeDataTypes = map[string]attr.Type{
 	"id":                    basetypes.StringType{},
 	"delete_on_termination": basetypes.BoolType{},
+}
+
+var agentDataTypes = map[string]attr.Type{
+	"provisioned": basetypes.BoolType{},
 }
 
 // NewServerDataSource is a helper function to simplify the provider implementation.
@@ -123,6 +128,16 @@ func (d *serverDataSource) Schema(_ context.Context, _ datasource.SchemaRequest,
 			"machine_type": schema.StringAttribute{
 				MarkdownDescription: "Name of the type of the machine for the server. Possible values are documented in [Virtual machine flavors](https://docs.stackit.cloud/products/compute-engine/server/basics/machine-types/)",
 				Computed:            true,
+			},
+			"agent": schema.SingleNestedAttribute{
+				Description: "STACKIT Server Agent as setup on the server",
+				Computed:    true,
+				Attributes: map[string]schema.Attribute{
+					"provisioned": schema.BoolAttribute{
+						Description: "Whether a STACKIT Server Agent is provisioned at the server",
+						Computed:    true,
+					},
+				},
 			},
 			"availability_zone": schema.StringAttribute{
 				Description: "The availability zone of the server.",
@@ -304,6 +319,18 @@ func mapDataSourceFields(ctx context.Context, serverResp *iaas.Server, model *Da
 	} else {
 		model.BootVolume = types.ObjectNull(bootVolumeDataTypes)
 	}
+
+	agentProvisioned := types.BoolNull()
+	if serverResp.Agent != nil && serverResp.Agent.Provisioned != nil {
+		agentProvisioned = types.BoolPointerValue(serverResp.Agent.Provisioned)
+	}
+	agent, diags := types.ObjectValue(agentDataTypes, map[string]attr.Value{
+		"provisioned": agentProvisioned,
+	})
+	if diags.HasError() {
+		return fmt.Errorf("failed to map agent: %w", core.DiagsToError(diags))
+	}
+	model.Agent = agent
 
 	if serverResp.UserData != nil && len(*serverResp.UserData) > 0 {
 		model.UserData = types.StringValue(string(*serverResp.UserData))
--- a/stackit/internal/services/iaas/server/datasource_test.go
+++ b/stackit/internal/services/iaas/server/datasource_test.go
@@ -10,6 +10,10 @@ import (
 	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
 )
 
+var expectedNullAgentData = types.ObjectValueMust(agentDataTypes, map[string]attr.Value{
+	"provisioned": types.BoolNull(),
+})
+
 func TestMapDataSourceFields(t *testing.T) {
 	type args struct {
 		state  DataSourceModel
@@ -40,6 +44,7 @@ func TestMapDataSourceFields(t *testing.T) {
 				ServerId:          types.StringValue("sid"),
 				Name:              types.StringNull(),
 				AvailabilityZone:  types.StringNull(),
+				Agent:             expectedNullAgentData,
 				Labels:            types.MapNull(types.StringType),
 				ImageId:           types.StringNull(),
 				NetworkInterfaces: types.ListNull(types.StringType),
@@ -77,7 +82,10 @@ func TestMapDataSourceFields(t *testing.T) {
 							NicId: new("nic2"),
 						},
 					},
-					KeypairName:   new("keypair_name"),
+					KeypairName: new("keypair_name"),
+					Agent: &iaas.ServerAgent{
+						Provisioned: new(true),
+					},
 					AffinityGroup: new("group_id"),
 					CreatedAt:     new(testTimestamp()),
 					UpdatedAt:     new(testTimestamp()),
@@ -100,7 +108,10 @@ func TestMapDataSourceFields(t *testing.T) {
 					types.StringValue("nic1"),
 					types.StringValue("nic2"),
 				}),
-				KeypairName:   types.StringValue("keypair_name"),
+				KeypairName: types.StringValue("keypair_name"),
+				Agent: types.ObjectValueMust(agentDataTypes, map[string]attr.Value{
+					"provisioned": types.BoolValue(true),
+				}),
 				AffinityGroup: types.StringValue("group_id"),
 				CreatedAt:     types.StringValue(testTimestampValue),
 				UpdatedAt:     types.StringValue(testTimestampValue),
@@ -131,6 +142,7 @@ func TestMapDataSourceFields(t *testing.T) {
 				Labels:            types.MapValueMust(types.StringType, map[string]attr.Value{}),
 				ImageId:           types.StringNull(),
 				NetworkInterfaces: types.ListNull(types.StringType),
+				Agent:             expectedNullAgentData,
 				KeypairName:       types.StringNull(),
 				AffinityGroup:     types.StringNull(),
 				UserData:          types.StringNull(),
--- a/stackit/internal/services/iaas/server/resource.go
+++ b/stackit/internal/services/iaas/server/resource.go
@@ -26,6 +26,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/objectplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
@@ -65,6 +66,7 @@ type Model struct {
 	ServerId          types.String `tfsdk:"server_id"`
 	MachineType       types.String `tfsdk:"machine_type"`
 	Name              types.String `tfsdk:"name"`
+	Agent             types.Object `tfsdk:"agent"`
 	AvailabilityZone  types.String `tfsdk:"availability_zone"`
 	BootVolume        types.Object `tfsdk:"boot_volume"`
 	ImageId           types.String `tfsdk:"image_id"`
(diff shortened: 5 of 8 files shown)


docs/resources/server.md
```diff
@@ -6,46 +6,43 @@ description: |-
   Server resource schema. Must have a region specified in the provider configuration.
   Example Usage
   With key pair
-  
-  resource "stackit_key_pair" "keypair" {
+  ```terraform
+  resource "stackitkeypair" "keypair" {
```
cgoetz-inovex: these changes look odd, sure these belong into this PR? Probably also explain the check-docs-failure on CI.

cgoetz-inovex (dismissed):
lgtm besides the docs

docs/resources/server.md
rubenhoenle: Apart from that the docs are auto-generated and this looks like manual edits to these documents.

See the result of `make generate-docs` :wink:

stackit/internal/services/iaas/server/datasource.go
```diff
@@ -123,6 +128,16 @@ func (d *serverDataSource) Schema(_ context.Context, _ datasource.SchemaRequest,
 				MarkdownDescription: "Name of the type of the machine for the server. Possible values are documented in [Virtual machine flavors](https://docs.stackit.cloud/products/compute-engine/server/basics/machine-types/)",
 				Computed:            true,
 			},
+			"agent": schema.SingleNestedAttribute{
```
rubenhoenle, line 132: The acceptance tests don't include your new field yet:

https://github.com/stackitcloud/terraform-provider-stackit/blob/44312a8ebc5b9da9af0608d845b2b733ea349851/stackit/internal/services/iaas/testdata/resource-volume-max.tf

https://github.com/stackitcloud/terraform-provider-stackit/blob/44312a8ebc5b9da9af0608d845b2b733ea349851/stackit/internal/services/iaas/iaas_acc_test.go

Could you also adjust them please?

stackit/internal/services/iaas/server/datasource.go
```diff
@@ -123,6 +128,16 @@ func (d *serverDataSource) Schema(_ context.Context, _ datasource.SchemaRequest,
...
+				Description: "STACKIT Server Agent as setup on the server",
+				Computed:    true,
+				Attributes: map[string]schema.Attribute{
+					"provisioned": schema.BoolAttribute{
+						Description: "Whether a STACKIT Server Agent is provisioned at the server",
+						Computed:    true,
```
rubenhoenle, line 138: Here's what the API docs say about the "provisioned" server agent field: https://docs.api.stackit.cloud/documentation/iaas/version/v2#tag/Servers/operation/v2CreateServer

> When false the agent IS NOT installed. When true the agent IS installed. When its not set the result depend on the used image and its default provisioning setting.

Since this is not trivial to solve, we as the STACKIT Developer Tools team decided that the default value for the "provisioned" field should be `false`. Could you please adjust your implementation to respect that decision? :sweat_smile:

docs/resources/server.md
aeter: I'll update my `tfplugindocs` from v0.16.0 to the newest version 0.21.0, per `scripts/project.sh`, and update the PR's docs. 

The docs were not edited manually. My dev branch was cut after 74ed4bd38be285d1ac24a110ae66d841c6184ff3. I just trusted the auto-doc-generator of `make generate-docs` - without seeing I have an old version of `tfplugindocs` installed locally some years ago for contributing to this project. I did skim over the generated docs, but (wrongly) thought anything not mine were leftover docs from previous commits in this repo that people had forgotten to commit.

```
[~/terraform-provider-stackit] go version -m $(which tfplugindocs)
/home/debian/go/bin/tfplugindocs: go1.22.1
        path    github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs
        mod     github.com/hashicorp/terraform-plugin-docs      v0.16.0 h1:UmxFr3AScl6Wged84jndJIfFccGyBZn52KtMNsS12dI=
```

stackit/internal/services/iaas/server/datasource.go
aeter: edited: I did not want to add code to files I cannot test.
If this is a strong requirement for this PR, how can I do this?

stackit/internal/services/iaas/server/datasource.go
aeter: To be sure I understand it right:

* By the previous code, so far, the param always has defaulted to null. (I think) I keep it as it has always been. Like - don't send any agent-related param to the API.
* Currently (by my code) the param has 3 states: `null`, `false`, `true`. Default remains null.

It seems like you want to be even more explicit - and to always default the param to `false`, without a `null` state (thus - to not use any image default agent provisioning) . 
OK. I'll update the code (cannot do it today, I also need to re-test it all).

docs/resources/server.md
aeter: I've updated my `tfplugindocs` to v0.21.0, ran `make generate-docs` again and force-pushed an update to the PR (docs only).

stackit/internal/services/iaas/server/datasource.go
btotev-schwarz: As a product owner of STACKIT Server Agent, I disagree with this to make the agent provisioned false by default.

Please note that agent is also a prerequisite to use the server update management service. If we put it by default false, I am sure that we will need to answer on a lot of customer support tickets, why the updates are not working. (other features like basic server monitoring are also planned to work only via the agent). 
From other point of view there are some concerns to make the agent provisioning always true as on custom images this might not be expected. 

There was a long discussion last year about the default value for agent installation on different levels (even up to the patrons). The final decision which was made is that it is the best is to have the default property configured on the image. This means that on publicly provided by STACKIT images it will be true and for custom images the customers can decide what to do. 

I agree on this that image resource should return also the agent properties to make it more transparent, but we shouldn't have a default value about agent provisioning on server level.

The behavior of this property should be consistent through all available options to create a server - API, Portal, Terraform provider, CLI. That is why please ensure that it follows the already agreed behavior:
 - no default value for the agent on server level
 - if there is no value provided, use the value from the API (UseStateForUnknown)
 - agent provisioning property is read only, so if you change it, the server need to be recreated (RequiresReplace)

rubenhoenle: @btotev-schwarz We decided to use a default value (`false`) as the Developer Tools team and maintainers of this Terraform provider for the following reason: The API in it's current state is **not declarative-friendly**. 

If one decides to not send a "provisioned" flag for a [server POST request](https://docs.api.stackit.cloud/documentation/iaas/version/v2#tag/Servers/operation/v2CreateServer), the "provisioned" flag in the response can either be `true` or `false` in the response, but you can't know beforehand. This is not at all compatible with declarative tools like Terraform IMO. Making this work in Terraform will always require some kind of workarounds and I'm not willing to accept this. 
 
> - agent provisioning property is read only, so if you change it, the server need to be recreated (RequiresReplace)

Don't mix up read-only fields and fields marked as `RequiresReplace`. `agent.provisioned` can't be read-only because the user has to be able to set a value for it in his Terraform configuration. But it has to be marked as `RequiresReplace` because the [server update endpoint](https://docs.api.stackit.cloud/documentation/iaas/version/v2#tag/Servers/operation/v2UpdateServer) doesn't allow updating the "provisioned" flag of an existing server.

stackit/internal/services/iaas/server/datasource.go
rubenhoenle: Yes it is, we won't accept this without properly adjusted Acc tests.

rubenhoenle: Since you're working at STACKIT it shouldn't be a problem for you I hope? :sweat_smile:

stackit/internal/services/iaas/server/datasource.go
aeter: I'm sorry, you'll think I'm stubborn to continue the discussion and propose some technical solution. I know my code in the PR needs adjustments. I know the IaaS API itself probably could use a better design for this parameter (to use strings, like "yes", "no", "image-default" instead of a bool). I'll follow whatever you guys decide (product owners, terraform team).

A small question: is the below solution (keeping the current API model) harder to maintain for the team and harder/unintuitive for the users? Sorry from taking your time with it.
  
```
 // If the user omits "agent" and the API returns nothing, Terraform sees null in the config and null in the state. No changes are detected.
// If the API returns {"provisioned": true}, UseStateForUnknown() will copy that object into the plan. Terraform will respect the API's default.
    return schema.Schema{
        Attributes: map[string]schema.Attribute{
            "agent": schema.SingleNestedAttribute{
                // allow user to omit it, or API to set it.
                Optional: true,
                Computed: true,
                
                // If user omits it, keep the old value
                PlanModifiers: []planmodifier.Object{
                    objectplanmodifier.UseStateForUnknown(),
                },

                Attributes: map[string]schema.Attribute{
                    "provisioned": schema.BoolAttribute{
                        Optional: true, // Can be set 
                        Computed: true, // But the API might fill it too.
                        
                        // If user changes this value, force recreation (Immutable behavior)
                        PlanModifiers: []planmodifier.Bool{
                            boolplanmodifier.RequiresReplace(),
                        },
                    },
                },
            },
        },
    } 
 ```
 
It seems the "provisioned" flag is not returned from the API at all when the `agent: provisioned: ...` was not set during image creation. The above will work whether nothing is returned or `true/false` is returned.
```
[~/stackit-cli] stackit curl -X GET -H "Content-Type: application/json" https://iaas.api.stackit.cloud/v
2/projects/c904f41c-2f8c-4edb-b966-e87d65f10b64/regions/eu01/servers/5c013c9e-edc0-49a0-ba11-3c971f113bd
5
{
  "availabilityZone": "eu01-m",
  "bootVolume": {
    "deleteOnTermination": true,
    "id": "8af61d89-41cc-49cb-946c-bc75e0cb74c6"
  },
  "createdAt": "2026-01-22T10:30:10Z",
  "id": "5c013c9e-edc0-49a0-ba11-3c971f113bd5",
  "labels": {},
  "launchedAt": "2026-01-22T10:31:37Z",
  "machineType": "t1.1",
  "metadata": {},
  "name": "test2",
  "powerStatus": "RUNNING",
  "status": "ACTIVE",
  "updatedAt": "2026-01-22T10:31:37Z",
  "volumes": [
    "8af61d89-41cc-49cb-946c-bc75e0cb74c6"
  ]
}
```

If you're interested, I can provide further testing with creating a resource with the current code and modifying it (true/false/missing). Sorry for the interruption.

Comment (rubenhoenle) [reply stackit/internal/services/iaas/server/datasource.go:138]: > Is UseStateForUnknown() a workaround or officially supported plan modifier for such cases (https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown)?

Of course `UseStateForUnknown()`is a officially supported plan modifier, that's not up for discussion. But in this case I consider using it a workaround, because it wouldn't be needed at all with a declarative-friendly API design. We have decided that we as the Developer Tools team don't have the capacity at the moment to handle such API design failures somehow in our products. 

Only because an API design works for the portal, that doesn't mean we have to stick with the same behavior in the Developer Tools (Terraform provider, CLI, ...). The portal isn't a declarative IaC tool which keeps a state, the Terraform provider is. This is why I also don't count in the argument that every touchpoint should integrate this feature the same way - that's only possible if the API matches the requirements of these touchpoints. Which isn't the case here.

The question for me is not if we technically can integrate the API behavior 1:1 in the Terraform provider - of course it's possible. **The question for me is if it makes sense. And that's still not the case here from my perspective.**

Furthermore, I think that this would be the correct way to go in a Terraform configuration. See the example below.

```terraform
data "stackit_image" "example" {
  project_id = var.project_id
  image_id   = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
}

resource "stackit_server" "boot-from-volume" {
  project_id = var.project_id
  name       = "example-server"
  boot_volume = {
    size        = 64
    source_type = "image"
    source_id   = stackit_image.example.image_id
  }
  agent = {
    # This way you still have the same behavior. In a declarative manner. 
    # This can be included in the examples and users will be happy.

    # option 1
    provisioned = stackit_image.example.agent.provision_by_default
    # option 2
    # provisioned = stackit_image.example.agent.supported
  }
  availability_zone = "eu01-1"
  machine_type      = "g2i.1"
  keypair_name      = "example-keypair"
}
```

stackit/internal/services/iaas/server/datasource.go
aeter: as agreed, I will adjust the acceptance tests with my next PR code update.

stackit/internal/services/iaas/server/datasource.go
aeter: Adding technical context. As agreed, one possible implementation is: in terraform to use the fields 'provisionPolicy (always|never|inherit)' + 'provisioned (true|false)'. For this, the IaaS API GET/PATCH endpoints need to start always to return the 'agent' key with 'provisioned' state (true|false). Status of this PR (and the related stackit-cli PR, I guess): waiting for the API changes. To terraform provider maintainers: sorry; if you strongly prefer, please feel free to close the PRs, we can reopen/rework later when the API conditions are met.

aeter:
please review.

I've updated the code as requested - added `provisioning_policy`. Retested. I ran the acc tests, they failed with unrelated to my code failures. 

---

#### acceptance tests
```
[~/terraform-provider-stackit] TF_ACC=1 TF_ACC_PROJECT_ID=c904f41c-2f8c-4edb-b966-e87d65f10b64 STACKIT_SERVICE_ACCOUNT_KEY_PATH=~/terraform_dev/.terraform_key.json TF_ACC_REGION=eu01 go test -timeout=60m -v ./stackit/internal/services/iaas/... -run "TestAccServer(Min|Max)"
```

#### A server created without sending 'provisioning' flag via the CLI - and then imported into terraform:
```
[~/terraform_dev] STACKIT_TF_ENABLE_BETA_RESOURCES=true TF_LOG_PROVIDER=DEBUG terraform import stackit_server.imported_server c904f41c-2f8c-4edb-b966-e87d65f10b64,eu01,7359173a-b476-4c3c-9cd4-35c7984114e2
[~/terraform_dev] terraform show
Outputs:

imported_server = {
   agent             = {
       provisioning_policy = "INHERIT"
   }
   id                = "c904f41c-2f8c-4edb-b966-e87d65f10b64,eu01,7359173a-b476-4c3c-9cd4-35c7984114e2"
}
```

#### A server created in terraform with defaults (no 'agent' block):
```
server_info_from_data = {
 "agent" = {
   "provisioned" = tobool(null)
   "provisioning_policy" = "INHERIT"
 }
 "id" = "c904f41c-2f8c-4edb-b966-e87d65f10b64,eu01,b4e69af2-a8a6-407b-81e5-b32df3041719"
```

#### A server created in terraform like this. Note the ALWAYS->INHERIT in datasource, which can/should be fixed when the API is fixed one day:
```
//------------ main.tf
resource "stackit_network_interface" "server_nic" {
 project_id = "c904f41c-2f8c-4edb-b966-e87d65f10b64"
 network_id = "97c5dde4-cb9d-49b8-be55-9cdf0c3795e1"
}

resource "stackit_server" "myserver1" {
   project_id = "c904f41c-2f8c-4edb-b966-e87d65f10b64"
   name = "nackovterraform1"
   boot_volume = {
       size = 64
       source_type = "image"
       source_id = "21466190-b904-4267-8bf3-1be4323f4ffb"
       delete_on_termination = true
   }
   availability_zone = "eu01-1"
   agent = {
       provisioning_policy = "ALWAYS"
   }
   machine_type = "t1.1"
   network_interfaces = [ stackit_network_interface.server_nic.network_interface_id ]
}

data "stackit_server" "myserver1_data" {
  project_id = "c904f41c-2f8c-4edb-b966-e87d65f10b64"
  server_id  = stackit_server.myserver1.server_id
}

output "server_info_from_data" {
  value = data.stackit_server.myserver1_data
}

output "server_info_from_resource" {
  value = stackit_server.myserver1.agent
}
//------------

####### output from make apply
server_info_from_data = {
 "affinity_group" = tostring(null)
 "agent" = {
   "provisioned" = true
   "provisioning_policy" = "INHERIT"
 }
}
server_info_from_resource = {
  "provisioned" = true
  "provisioning_policy" = "ALWAYS"
}

```

stackit/internal/services/iaas/server/datasource.go
```diff
@@ -124,6 +130,20 @@ func (d *serverDataSource) Schema(_ context.Context, _ datasource.SchemaRequest,
...
+					},
+					"provisioning_policy": schema.StringAttribute{
+						Description: "The provisioning policy. In a data source, this defaults to \"INHERIT\" as the API does not provide this value.",
```
rubenhoenle: This `provisioning_policy` field is pretty useless in the datasource, just leave it out 😅

stackit/internal/services/iaas/server/resource_test.go
```diff
@@ -92,12 +93,15 @@ func TestMapFields(t *testing.T) {
...
+					Agent: &iaas.ServerAgent{
+						Provisioned: utils.Ptr(true),
+					},
 					KeypairName:   new("keypair_name"),
-					AffinityGroup: new("group_id"),
-					CreatedAt:     new(testTimestamp()),
-					UpdatedAt:     new(testTimestamp()),
-					LaunchedAt:    new(testTimestamp()),
-					Status:        new("active"),
+					AffinityGroup: utils.Ptr("group_id"),
```
rubenhoenle:
```suggestion
					AffinityGroup: new("group_id"),
```

Please leave this like it was before :)

stackit/internal/services/iaas/server/resource_test.go
```diff
@@ -240,6 +253,9 @@ func TestToCreatePayload(t *testing.T) {
 						NicIds: &[]string{"nic1", "nic2"},
 					},
 				},
+				Agent: &iaas.ServerAgent{
+					Provisioned: utils.Ptr(true),
```
rubenhoenle:
```suggestion
					Provisioned: new(true),
```

stackit/internal/services/iaas/server/resource_test.go
```diff
@@ -292,6 +312,9 @@ func TestToCreatePayload(t *testing.T) {
 						NicIds: &[]string{"nic1", "nic2"},
 					},
 				},
+				Agent: &iaas.ServerAgent{
+					Provisioned: utils.Ptr(false),
```
rubenhoenle:
```suggestion
					Provisioned: new(false),
```

stackit/internal/services/iaas/server/datasource.go
```diff
@@ -305,6 +325,27 @@ func mapDataSourceFields(ctx context.Context, serverResp *iaas.Server, model *Da
 		model.BootVolume = types.ObjectNull(bootVolumeDataTypes)
 	}
 
+	if serverResp.Agent != nil && serverResp.Agent.Provisioned != nil {
+		agent, diags := types.ObjectValue(agentDataTypes, map[string]attr.Value{
```
rubenhoenle:
```golang
provisioned := types.BoolNull()
if serverResp.Agent != nil && serverResp.Agent.Provisioned != nil {
    provisioned = types.BoolPointerValue(serverResp.Agent.Provisioned)
}

agent, diags := types.ObjectValue(agentDataTypes, map[string]attr.Value{
	"provisioned": provisioned
	"provisioning_policy": types.StringValue("INHERIT")
})
if diags.HasError() {
	return fmt.Errorf("failed to map agent: %w", core.DiagsToError(diags))
}
model.Agent = agent
```

I suggest to rewrite this part so it looks e.g. like above. You currently duplicate a lot of code in the if/else blocks. I think there are better ways.

stackit/internal/services/iaas/server/resource.go
```diff
@@ -166,6 +180,14 @@ func (r *serverResource) ValidateConfig(ctx context.Context, req resource.Valida
 		}
 	}
 
+	var agent = &agentModel{}
+	if !(model.Agent.IsNull() || model.Agent.IsUnknown()) {
+		diags := model.Agent.As(ctx, agent, basetypes.ObjectAsOptions{})
```
rubenhoenle: What's the point of this? I don't think you have to adjust the `ValidateConfig` implementation 😅

rubenhoenle: See https://developer.hashicorp.com/terraform/plugin/framework/resources/validate-configuration for reference

stackit/internal/services/iaas/server/resource.go
```diff
@@ -982,6 +1033,47 @@ func mapFields(ctx context.Context, serverResp *iaas.Server, model *Model, regio
 		model.NetworkInterfaces = types.ListNull(types.StringType)
 	}
 
+	if serverResp.Agent != nil && serverResp.Agent.Provisioned != nil {
+		// convert agent model
+		// provisioning_policy doesn't exist in the API, we must preserve it from current state
+		var currentAgent = &agentModel{}
```
rubenhoenle: Same here. Please don't duplicate your whole logic into the if and else block. I don't think that's necessary.

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -2294,7 +2298,7 @@ func TestAccServerMin(t *testing.T) {
 				},
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"boot_volume", "network_interfaces"}, // Field is not mapped as it is only relevant on creation
+				ImportStateVerifyIgnore: []string{"boot_volume", "network_interfaces", "agent"}, // Field is not mapped as it is only relevant on creation
```
rubenhoenle:
```suggestion
				ImportStateVerifyIgnore: []string{"boot_volume", "network_interfaces"}, // Field is not mapped as it is only relevant on creation
```

Since the `agent.provisioning_policy` field has the default value (`INHERIT`) in the min acc test case, the import should work properly, doesn't it?

stackit/internal/services/iaas/server/datasource.go
aeter: Done - removed `provisioning_policy` from the datasource. Reran the tests locally; with defaults, it looks like this:

```
server_info_from_data = {
  "provisioned" = tobool(null)
}
server_info_from_resource = {
  "provisioned" = tobool(null)
  "provisioning_policy" = "INHERIT"
}
```

stackit/internal/services/iaas/testdata/resource-server-max.tf
```diff
@@ -79,4 +79,7 @@ resource "stackit_server" "server" {
   labels = {
     "acc-test" : var.label
   }
+  agent = {
+    provisioning_policy = "ALWAYS"
```
rubenhoenle: Please use a variable instead of hardcoded values.

https://github.com/stackitcloud/terraform-provider-stackit/blob/13a6f356f5835b359ef9f92923ae3c07d5936597/stackit/internal/services/iaas/iaas_acc_test.go#L138-L152

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -2425,6 +2430,10 @@ func TestAccServerMax(t *testing.T) {
 					resource.TestCheckResourceAttr("stackit_key_pair.key_pair", "name", testutil.ConvertConfigVariable(testConfigServerVarsMax["name_not_updated"])),
 					resource.TestCheckResourceAttr("stackit_key_pair.key_pair", "public_key", testutil.ConvertConfigVariable(testConfigServerVarsMax["public_key"])),
 
+					// Agent
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioning_policy", "ALWAYS"),
```
rubenhoenle:
```suggestion
					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioning_policy", testutil.ConvertConfigVariable(testConfigServerVarsMax["agent_provisioning_policy"])),
```

If you use a variable you can just validate against the variable here without having to hardcode values

aeter:
updated per review

stackit/internal/services/iaas/server/resource.go
```diff
@@ -276,6 +290,35 @@ func (r *serverResource) Schema(_ context.Context, _ resource.SchemaRequest, res
...
+			"agent": schema.SingleNestedAttribute{
+				Description: "The STACKIT Server Agent configured for the server",
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.RequiresReplace(),
+				},
+				Attributes: map[string]schema.Attribute{
+					"provisioned": schema.BoolAttribute{
+						Description: "Whether a STACKIT Server Agent should be provisioned at the server",
```
rubenhoenle:
```suggestion
						Description: "Whether a STACKIT Server Agent is provisioned at the server",
```

stackit/internal/services/iaas/server/resource.go
```diff
@@ -982,6 +1025,31 @@ func mapFields(ctx context.Context, serverResp *iaas.Server, model *Model, regio
 		model.NetworkInterfaces = types.ListNull(types.StringType)
 	}
 
+	// agent{...} block, determine the intent policy from Terraform state
+	currentPolicy := types.StringValue("INHERIT")
+	if !(model.Agent.IsNull() || model.Agent.IsUnknown()) {
+		var currentAgent agentModel
+		if diags := model.Agent.As(ctx, &currentAgent, basetypes.ObjectAsOptions{}); !diags.HasError() {
```
rubenhoenle:
```suggestion
		diags := model.Agent.As(ctx, &currentAgent, basetypes.ObjectAsOptions{})
		if diags.HasError() {
		    return fmt.Errorf("failed to map agent: %w", core.DiagsToError(diags))
		}
```

I would suggest to **not** fail silently here. Would there be any reason to do so?

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -2223,6 +2225,8 @@ func TestAccServerMin(t *testing.T) {
 					resource.TestCheckResourceAttrSet("stackit_server.server", "created_at"),
 					resource.TestCheckResourceAttrSet("stackit_server.server", "launched_at"),
 					resource.TestCheckResourceAttrSet("stackit_server.server", "updated_at"),
+					resource.TestCheckResourceAttr("stackit_server.server", "agent.provisioning_policy", "INHERIT"),
+					resource.TestCheckNoResourceAttr("stackit_server.server", "agent.provisioned"),
```
rubenhoenle, line 2229: Shouldn't this be either true or false always?

What does a `nil` response for the provisioned field from the API mean?

stackit/internal/services/iaas/server/resource.go
```diff
@@ -276,6 +290,35 @@ func (r *serverResource) Schema(_ context.Context, _ resource.SchemaRequest, res
...
+				Attributes: map[string]schema.Attribute{
+					"provisioned": schema.BoolAttribute{
+						Description: "Whether a STACKIT Server Agent should be provisioned at the server",
+						Computed:    true,
+						PlanModifiers: []planmodifier.Bool{
+							boolplanmodifier.UseStateForUnknown(),
+						},
+					},
+					"provisioning_policy": schema.StringAttribute{
+						Description: "Agent provisioning policy: \"ALWAYS\", \"NEVER\", or \"INHERIT\". \"INHERIT\" follows the image default value.",
```
rubenhoenle:
```suggestion
						Description: "Agent provisioning policy: `ALWAYS`, `NEVER`, or `INHERIT`. `INHERIT` follows the image default value.",
```

You can use markdown formatting here 😅

aeter: I want quotes in the docs. Like:   

"INHERIT"

and not like

INHERIT

stackit/internal/services/iaas/iaas_acc_test.go
aeter: The default is "INHERIT", it's neither `true` nor `false`. The default is not `false`.

A `nil` response from API (i.e. a missing 'agent' key) means using local tfstate and defaulting to `INHERIT`.

rubenhoenle: Sorry, my question was independent from Terraform - what does it mean technically if the API response is missing the `provisioned.provisioned` field? Does it mean an agent was provisioned or not?

So in fact, what's the difference between a server for which in the API response the `provisioned.provisioned` is null/missing vs. a server for which the API response has `provisioned.provisioned`  set to `false`?

aeter: I think the current IAAS API gives 2 responses:

1. like: `{"id": "...", "network": "...", "agent": {"provisioned": true}}`
For `true`, the agent is sure to be provisioned on the server. For `false` - it's sure the agent is not provisioned on the server. 
This API response is seen only when the server was explicitly created (POST /servers) by specifying `{"agent": {"provisioned": true|false}}` next to the other keys in the JSON payload.

2. like `{"id": "...", "network": "..."}` - no `agent` key
The agent **may** be provisioned. Depends on the default for the server image (true|false). Also, I think, this image default could be changeable with enough rights in the API.
This API response is seen when the server was created (POST /servers) without mentioning an `agent` key in the JSON payload.

rubenhoenle: So for the 2nd case you mentioned the question to "Is there an agent provisioned on the server?" is "Schrödingers cat"? :astonished:

aeter: Yes, I think so. Thus the 3 cases (inherit, always, never) at the terraform provider's PR code. The /images api cannot be used to determine the provisioning state reliably (as the image default value may be changed). I think the /command Run Commands apis currently don't provide an endpoint for the agent provisioning state either. In the future, it's likely there will be api improvements.

stackit/internal/services/iaas/server/resource.go
rubenhoenle: And I would like you to stick to our conventions :sweat_smile: 

https://github.com/stackitcloud/terraform-provider-stackit/blob/7fb8cdbed265c804bb6a3f6d2da9bb3f51d1fdee/docs/resources/logs_instance.md?plain=1#L66

You could even use these util func to format your values if you want:

https://github.com/stackitcloud/terraform-provider-stackit/blob/7fb8cdbed265c804bb6a3f6d2da9bb3f51d1fdee/stackit/internal/utils/utils.go#L141-L152

rubenhoenle: Please don't resolve such conversations silently if you want to discuss about the comments instead of resolving them in your code. Makes review pretty hard if I have to check every resolved comment over and over again. Thanks :blush:

aeter: OK, I'll change it. 

Note: If this is something you'd really want to standardize, it could make sense to setup a separate PR to fix it everywhere once and for all - the majority of the code currently is a mixture of quote styles, like these examples of many styles in the same file.

```
docs/data-sources/dns_record_set.md
42:- `id` (String) Terraform's internal data source. ID. It is structured as "`project_id`,`zone_id`,`record_set_id`".
47:- `type` (String) The record set type. E.g. `A` or `CNAME`
54:- `read` (String) A string that can be [parsed as a duration](https://pkg.go.dev/time#ParseDuration) consisting of numbers and unit suffixes, such as "30s" or "2h45m". Valid time units are "s" (seconds), "m" (minutes), "h" (hours).

docs/data-sources/application_load_balancer.md
35:- `id` (String) Terraform's internal resource ID. It is structured as `project_id`,`region`,`name`.
53:- `description` (String) The error description contains additional helpful user information to fix the error state of the Application Load Balancer. For example the IP 45.135.247.139 does not exist in the project, then the description will report: Floating IP "45.135.247.139" could not be found.
67:- `waf_config_name` (String) Enable Web Application Firewall (WAF), referenced by name. See "Application Load Balancer - Web Application Firewall API" for more information.
119:- `exact_match` (String) Exact path match. Only a request path exactly equal to the value will match, e.g. '/foo' matches only '/foo', not '/foo/bar' or '/foobar'.
120:- `prefix` (String) Prefix path match. Only matches on full segment boundaries, e.g. '/foo' matches '/foo' and '/foo/bar' but NOT '/foobar'.
```

Note2: there are good technical reasons to use quotes (avoiding code-in-code formatting at the tf cli, tfplugindocs, etc.) - but let's go with backticks, since this is what you require for this PR.

rubenhoenle (dismissed):

stackit/internal/services/iaas/server/datasource_test.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/stackitcloud/stackit-sdk-go/services/iaas"
 )
 
+var expectedNullAgentData, _ = types.ObjectValue(agentDataTypes, map[string]attr.Value{
```
Manuelvaas: Please use the types.ObjectValueMust

stackit/internal/services/iaas/server/resource_test.go
```diff
@@ -20,6 +20,11 @@ const (
 	testTimestampValue    = "2006-01-02T15:04:05Z"
 )
 
+var expectedNullAgent, _ = types.ObjectValue(agentTypes, map[string]attr.Value{
```
Manuelvaas: Here as well

Manuelvaas (changes requested):

rubenhoenle (approved):

Manuelvaas (approved):

PR #1164, opened 2026-01-28
feat(iaas): add image share resource

h3adex:
## Description

Problems with this resource:

1. When you set parent_organization = true, you are telling the API to share the image with every project in your organization. projects = null (Because they are mutually exclusive, user don't set this). The API calculates which projects belongs to your organization and returns that list of projects in the response (e.g., projects: ["proj-A", "proj-B"]). To fix this we ignore the data the API returns for projects whenever parent_organization is active (mapShareFields).

2. The API treats image sharing as a single global configuration. Creating a new share silently overwrites any existing one. Fix: Added a Check-Then-Act validation in Create (using authorization resource utils right now, should be moved to general utils?)

3. The API rejects null for projects (requires []), but throws a 400 Conflict if both projects and parentOrganization keys are present in the payload, even if one is empty/null. Fix: Implemented custom payload serialization (toSetImageSharePayload) that explicitly sets unused fields to nil (omitting the key) while initializing used fields

Open Problem (unlikely case a user does that but still open):

Resource1: Sets parent_organization = false.
Resource2: runs its pre-create check, it calls the API. My duplicate check logic interprets this as "No resource exists," effectively making Resource 1 invisible to the conflict detector.

```hcl
resource "stackit_image" "image" {
  project_id      = var.stackit_project_id
  name            = "dev"
  disk_format     = "raw"
  local_file_path = "./files/test-512k.img"
}

resource "stackit_image_share" "resource1" {
  project_id          = var.stackit_project_id
  image_id            = stackit_image.image.image_id
  // a user says explicit this image share should not be shared accros parent org. Therefor also projects
  parent_organization = false
}

// duplicate func won't detect stackit_image_share.share. On next apply they will update each other :-D
resource "stackit_image_share" "resource2" {
  project_id          = var.stackit_project_id
  image_id            = stackit_image.image.image_id
  projects = [
    "e8072f14-30e8-4d0e-8c24-a87322e507bf",
    "f8f925f1-2d67-49e9-9c50-2efcfb261899",
  ]
}
```

I don't have a solution for this problem. We could add this resource as beta and add a comment to the documentation which explains that each image can only have a single share resource.

## Checklist

- [ ] Issue was linked above


h3adex:
remove stale

rubenhoenle:
> @h3adex please fix the failing CI pipeline

@h3adex ping

PR #1165, opened 2026-01-28
Onboard logs service access token

GokceGK:
## Description

relates to STACKITTPR-449

## Checklist

- [x] Issue was linked above

### DIFF
--- a/stackit/internal/services/logs/logs_acc_test.go
+++ b/stackit/internal/services/logs/logs_acc_test.go
@@ -410,6 +410,7 @@ func TestAccLogsAccessTokenMin(t *testing.T) {
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "id"),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "access_token_id"),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "creator"),
+					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "access_token"),
 					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "expires", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["expires"])),
 					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "status", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["status"])),
 				),
@@ -482,6 +483,7 @@ func TestAccLogsAccessTokenMin(t *testing.T) {
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "creator"),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "expires"),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "status"),
+					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "access_token"),
 				),
 			},
 			// Deletion handled by framework
@@ -531,6 +533,7 @@ func TestAccLogsAccessTokenMax(t *testing.T) {
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "id"),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "access_token_id"),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "creator"),
+					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "access_token"),
 					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "expires", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMax["expires"])),
 					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "status", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMax["status"])),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "valid_until"),
@@ -607,6 +610,7 @@ func TestAccLogsAccessTokenMax(t *testing.T) {
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "creator"),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "expires"),
 					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "status"),
+					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "access_token"),
 				),
 			},
 			// Deletion handled by framework
--- a/docs/data-sources/logs_access_token.md
+++ b/docs/data-sources/logs_access_token.md
@@ -45,5 +45,5 @@ data "stackit_logs_access_token" "accessToken" {
 - `expires` (Boolean) Indicates if the access token can expire
 - `id` (String) Terraform's internal resource identifier. It is structured as "`project_id`,`region`,`instance_id`,`access_token_id`".
 - `permissions` (List of String) The access permissions granted to the access token. Possible values: `read`, `write`.
-- `status` (String) The status of the access token, possible values: Possible values are: `active`, `expired`.
+- `status` (String) The status of the access token. Possible values are: `active`, `expired`.
 - `valid_until` (String) The date and time until an access token is valid to (inclusively)
--- a/docs/resources/logs_access_token.md
+++ b/docs/resources/logs_access_token.md
@@ -73,5 +73,5 @@ import {
 - `creator` (String) The user who created the access token
 - `expires` (Boolean) Indicates if the access token can expire
 - `id` (String) Terraform's internal resource identifier. It is structured as "`project_id`,`region`,`instance_id`,`access_token_id`".
-- `status` (String) The status of the access token, possible values: Possible values are: `active`, `expired`.
+- `status` (String) The status of the access token. Possible values are: `active`, `expired`.
 - `valid_until` (String) The date and time until an access token is valid to (inclusively)


examples/resources/stackit_logs_access_token/resource.tf
```diff
@@ -0,0 +1,27 @@
...
+  region       = "eu01"
+  display_name = "logs-access-token-example"
+  lifetime     = 30
+  permissions = [
+    "write"
+  ]
+  description = "Example description"
+}
+
+# Only use the import statement, if you want to import an existing logs instance
```
marceljk:
```suggestion
# Only use the import statement, if you want to import an existing logs access token
```
Please add here also an hint, that the actual access token will not be imported because it can be only read during the creation and that probably lifetime needs to be added to ignore_changes. 
Similar like for example in the server resource:
https://github.com/stackitcloud/terraform-provider-stackit/blob/d772e2e188b6b8eaa4962933bad88ed783147ff3/docs/resources/server.md?plain=1#L381-L388

GokceGK: Made the change and added the extra info about lifetime field to ignore

stackit/internal/services/logs/accesstoken/datasource.go
```diff
@@ -0,0 +1,237 @@
...
+	accessTokenResponse, err := d.client.GetAccessToken(ctx, projectID, region, instanceID, accessTokenID).Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		ok := errors.As(err, &oapiErr)
+		if ok && oapiErr.StatusCode == http.StatusNotFound {
+			resp.State.RemoveResource(ctx)
+			return
+		}
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading Logs access token", fmt.Sprintf("Calling API: %v", err))
+		return
```
marceljk: You can adjust the error handling in datasources like here:
https://github.com/stackitcloud/terraform-provider-stackit/blob/53a36978504717f34b06f951bb8fc65eba69c3c2/stackit/internal/services/iaas/network/datasource.go#L220-L231

This makes it easier to handle different kinds of error, like for example 403

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -0,0 +1,519 @@
...
+	if !ok {
+		return
+	}
+
+	features.CheckBetaResourcesEnabled(ctx, &providerData, &resp.Diagnostics, "stackit_logs_access_token", "resource")
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	apiClient := utils.ConfigureClient(ctx, &providerData, &resp.Diagnostics)
```
marceljk: nitpick, you can directly assign here the client on the resource struct
```suggestion
	r.client := utils.ConfigureClient(ctx, &providerData, &resp.Diagnostics)
```

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -0,0 +1,519 @@
...
+	client       *logs.APIClient
+	providerData core.ProviderData
+}
+
+func NewLogsAccessTokenResource() resource.Resource {
+	return &logsAccessTokenResource{}
+}
+
+func (r *logsAccessTokenResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	providerData, ok := conversion.ParseProviderData(ctx, req.ProviderData, &resp.Diagnostics)
```
marceljk: Similar like for r.client
```suggestion
	var ok bool
	r.providerData, ok := conversion.ParseProviderData(ctx, req.ProviderData, &resp.Diagnostics)
```

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -0,0 +1,519 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	apiClient := utils.ConfigureClient(ctx, &providerData, &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	r.client = apiClient
+	r.providerData = providerData
```
marceljk: then these lines can be removed

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -0,0 +1,519 @@
...
+				ElementType: types.StringType,
+				Required:    true,
+				PlanModifiers: []planmodifier.List{
+					listplanmodifier.RequiresReplace(),
+				},
+			},
+			"status": schema.StringAttribute{
+				Description: schemaDescriptions["status"],
+				Computed:    true,
+			},
```
marceljk, line 242: Isn't the access_token itself missing as attribute here in the schema and in the model?

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -0,0 +1,519 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	ctx = core.InitProviderContext(ctx)
+
+	instanceId := model.InstanceID.ValueString()
+	projectId := model.ProjectID.ValueString()
+	region := model.Region.ValueString()
```
marceljk: you don't need to read the region directly, you can just use the GetRegionWithOverride function
```suggestion
	region := r.providerData.GetRegionWithOverride(model.Region)
```

and remove the second call to this function, which is before the call to `CreateAccessToken()`

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -0,0 +1,519 @@
...
+		return nil, fmt.Errorf("missing model")
+	}
+
+	payload := &logs.CreateAccessTokenPayload{
+		Description: conversion.StringValueToPointer(model.Description),
+		DisplayName: conversion.StringValueToPointer(model.DisplayName),
+		Lifetime:    conversion.Int64ValueToPointer(model.Lifetime),
+	}
+
+	if !(model.Permissions.IsNull() || model.Permissions.IsUnknown()) {
```
marceljk: nitpick
```suggestion
	if !(tfutils.IsUndefined(model.Permissions)) {
```

stackit/internal/services/logs/logs_acc_test.go
```diff
@@ -330,6 +376,247 @@ data "stackit_logs_instance" "logs" {
...
+
+					// Access token data
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "project_id", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["project_id"])),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "region", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["region"])),
+					resource.TestCheckResourceAttrPair(
+						"stackit_logs_access_token.accessToken", "instance_id",
+						"stackit_logs_instance.logs", "instance_id",
+					),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "display_name", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["display_name"])),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "permissions.0", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["permissions"])),
```
marceljk, line 408: optional: you can also add a check here, if the list permissions has the size 1. So just to be sure the api has not extended it 

```suggestion
					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "permissions.0", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["permissions"])),
					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "permissions.#", "1")
```

can be also added in datasource and update and in the max testcase

marceljk, line 408: some attribute checks are missing here like: "creator", "expires", "id", "status", "valid_until".
For most of them it's enough if you check if they are just set, because you may don't know the value they will have. Except `expires`, there you should know if it's false or true and same for status, there it will be probably always "active"

GokceGK: nice catch, thank you :)

GokceGK: I am only not sure about the valid_until field as it is not required field, it might be also nil.

stackit/internal/services/logs/accesstoken/datasource.go
```diff
@@ -0,0 +1,239 @@
...
+	ctx = tflog.SetField(ctx, "access_token_id", accessTokenID)
+
+	accessTokenResponse, err := d.client.GetAccessToken(ctx, projectID, region, instanceID, accessTokenID).Execute()
+	if err != nil {
+		tfutils.LogError(
+			ctx,
+			&resp.Diagnostics,
+			err,
+			"Reading Logs access token",
+			fmt.Sprintf("Calling API: %v", err),
```
marceljk: This is the default error message and will mostly be a 404.
So the message should be something like "Access token with ID $q does not exist in project %q"

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -306,13 +311,17 @@ func (r *logsAccessTokenResource) Read(ctx context.Context, req resource.ReadReq
...
+			ctx,
+			&resp.Diagnostics,
+			err,
+			"Reading Logs access token",
+			fmt.Sprintf("Calling API: %v", err),
+			map[int]string{
+				http.StatusForbidden: fmt.Sprintf("Project with ID %q not found or forbidden access", projectID),
+			},
+		)
+		resp.State.RemoveResource(ctx)
```
marceljk: Can you revert this change? The error logging with the LogError function should only be used in the datasource, because in the datasource we don't care about the exact status code and remove there always the resource from the state. There we can easily "restore" the state by running again `$ terraform apply`.
But in the resource we should remove the resource from the state only when we get a 404. So we need to explicit check here, which status code was sent.

stackit/internal/services/logs/logs_acc_test.go
```diff
@@ -425,6 +435,7 @@ func TestAccLogsAccessTokenMin(t *testing.T) {
 					resource.TestCheckResourceAttrSet("data.stackit_logs_access_token.accessToken", "access_token_id"),
 					resource.TestCheckResourceAttr("data.stackit_logs_access_token.accessToken", "display_name", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["display_name"])),
 					resource.TestCheckResourceAttr("data.stackit_logs_access_token.accessToken", "permissions.0", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["permissions"])),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "permissions.#", "1"),
 					resource.TestCheckResourceAttrSet("data.stackit_logs_access_token.accessToken", "creator"),
 					resource.TestCheckResourceAttrSet("data.stackit_logs_access_token.accessToken", "expires"),
 					resource.TestCheckResourceAttrSet("data.stackit_logs_access_token.accessToken", "status"),
```
marceljk: The value of these two attributes can be checked as well. They should have the same value like in the resource
```suggestion
					resource.TestCheckResourceAttr("data.stackit_logs_access_token.accessToken", "expires", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["expires"])),
					resource.TestCheckResourceAttr("data.stackit_logs_access_token.accessToken", "status", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["status"])),
```

stackit/internal/services/logs/logs_acc_test.go
marceljk: Didn't tested it yet, but I assume that `valid_until` is set when `expires` is true and is not set when `expires` is false.

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -0,0 +1,532 @@
...
+				Description: schemaDescriptions["lifetime"],
+				Optional:    true,
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.RequiresReplace(),
+				},
+			},
+			"permissions": schema.ListAttribute{
+				Description: schemaDescriptions["permissions"],
+				ElementType: types.StringType,
+				Required:    true,
```
marceljk, line 231: optional: you can add here a validator to check if the size of the list is always at least 1. Because when it's an empty list, the api throws an error

stackit/internal/services/logs/logs_acc_test.go
GokceGK: yes, sounds good to me. I have this case in the max test.

stackit/internal/services/logs/accesstoken/resource.go
GokceGK: I have reverted back to old error handling

stackit/internal/services/logs/accesstoken/resource.go
GokceGK: added validator

stackit/internal/services/logs/accesstoken/datasource.go
GokceGK: thx, added the message

stackit/internal/services/logs/accesstoken/resource.go
```diff
@@ -0,0 +1,534 @@
...
+	"access_token":    "The generated access token",
+	"creator":         "The user who created the access token",
+	"description":     "The description of the access token",
+	"display_name":    "The displayed name of the access token",
+	"expires":         "Indicates if the access token can expire",
+	"valid_until":     "The date and time until an access token is valid to (inclusively)",
+	"lifetime":        "A lifetime period for an access token in days. If unset the token will not expire.",
+	"permissions":     "The access permissions granted to the access token. Possible values: `read`, `write`.",
+	"status": fmt.Sprintf(
+		"The status of the access token, possible values: %s",
```
marceljk:
```suggestion
		"The status of the access token. %s",
```
The function `FormatPossibleValues` already adds some text

stackit/internal/services/logs/logs_acc_test.go
```diff
@@ -330,6 +380,264 @@ data "stackit_logs_instance" "logs" {
...
+						"stackit_logs_instance.logs", "instance_id",
+					),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "display_name", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["display_name"])),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "permissions.0", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["permissions"])),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "permissions.#", "1"),
+					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "id"),
+					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "access_token_id"),
+					resource.TestCheckResourceAttrSet("stackit_logs_access_token.accessToken", "creator"),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "expires", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["expires"])),
+					resource.TestCheckResourceAttr("stackit_logs_access_token.accessToken", "status", testutil.ConvertConfigVariable(testConfigAccessTokenVarsMin["status"])),
```
marceljk, line 415: Can you please also add a check if the field "access_token" is set? Add this check as well in the update part and in the create and update of max test

GokceGK: thx, added it

PR #1171, opened 2026-01-30
fix(cdn): error when optional blocked_countries field is not defined as empty list

pavlov-tony:
## Description

Customer reported an error on distribution creation with non specified optional `blocked_countries` property.

relates to [STACKITCDN-1014](https://jira.schwarz/browse/STACKITCDN-1014)

## Checklist

- [x] Issue was linked above

### DIFF
--- a/stackit/internal/services/cdn/distribution/resource.go
+++ b/stackit/internal/services/cdn/distribution/resource.go
@@ -828,7 +828,7 @@ func toCreatePayload(ctx context.Context, model *Model) (*cdn.CreateDistribution
 				OriginUrl:            cfg.Backend.HttpBackend.OriginUrl,
 				OriginRequestHeaders: cfg.Backend.HttpBackend.OriginRequestHeaders,
 				Geofencing:           cfg.Backend.HttpBackend.Geofencing,
-				Type:                 cdn.PtrString("http"),
+				Type:                 cfg.Backend.HttpBackend.Type,
 			},
 		},
 		IntentId:         cdn.PtrString(uuid.NewString()),


stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -24,98 +26,150 @@ import (
...
-	"project_id":                testutil.ProjectId,
-	"config_backend_type":       "http",
-	"config_backend_origin_url": "https://test-backend-1.cdn-dev.runs.onstackit.cloud",
-	"config_regions":            "\"EU\", \"US\"",
-	"config_regions_updated":    "\"EU\", \"US\", \"ASIA\"",
-	"blocked_countries":         "\"CU\", \"AQ\"", // Do NOT use DE or AT here, because the request might be blocked by bunny at the time of creation - don't lock yourself out
-	"custom_domain_prefix":      uuid.NewString(), // we use a different domain prefix each test run due to inconsistent upstream release of domains, which might impair consecutive test runs
-	"dns_name":                  fmt.Sprintf("tf-acc-%s.stackit.gg", strings.Split(uuid.NewString(), "-")[0]),
+// 1. Define the Configuration Struct
+type distributionConfig struct {
```
rubenhoenle: Why are these changes here needed? Is it needed for the bugfix or just some refactoring?

In case it's just some refactoring and not explicitly needed, I would ask you to please split it into another PR.

And one note about acc tests: We try to use the following schema for our acc tests with embedding some `*.tf` files. So the Terraform config is not part of the Golang test code. I would like to ask you to stick to that format to keep everything consistent across the codebase. 😅 

As you can see usually we have one `Min` test with a minimal config and one `Max` test for the full-blown config for each resource.

https://github.com/stackitcloud/terraform-provider-stackit/blob/6f4cb491b1198f03639063d05ac07eb3853c6d0b/stackit/internal/services/git/git_acc_test.go#L22-L26

https://github.com/stackitcloud/terraform-provider-stackit/blob/6f4cb491b1198f03639063d05ac07eb3853c6d0b/stackit/internal/services/git/testdata/resource-min.tf#L1-L8

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -281,8 +282,18 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
 						ElementType: types.StringType,
+						// The API returns an empty list for blocked_countries even if the field is omitted
+						// (null) in the request. This causes an "inconsistent result" error in Terraform
+						// because the config is null but the state is [].
+						//
+						// By setting a Default value of an empty list, we tell Terraform to treat a missing
+						// blocked_countries block in the HCL as if the user explicitly defined
+						// blocked_countries = []. This ensures the config (empty list) matches the
+						// API response (empty list).
+						Default: listdefault.StaticValue(types.ListValueMust(types.StringType, []attr.Value{})),
```
rubenhoenle, line 296: Did you check if you have the same problem for `regions`? 😅

rubenhoenle: ah nvm, that's required

stackit/internal/services/cdn/cdn_acc_test.go
pavlov-tony: I added templating for ommiting `blocked_countries` to cover bug with test case.
But now I realized that I just can specify `null` and it would be an equivalent. I'll revert all those changes to tests. Regarding using the same approach in testing as other services - we will deliver it in separate pull request

pavlov-tony: Addressed in https://github.com/stackitcloud/terraform-provider-stackit/pull/1171/commits/217976dedcee416ac3ee88f3bb2ced673b6811ca

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -812,6 +823,14 @@ func toCreatePayload(ctx context.Context, model *Model) (*cdn.CreateDistribution
 	}
 
 	payload := &cdn.CreateDistributionPayload{
+		Backend: &cdn.CreateDistributionPayloadBackend{
+			HttpBackendCreate: &cdn.HttpBackendCreate{
+				OriginUrl:            cfg.Backend.HttpBackend.OriginUrl,
+				OriginRequestHeaders: cfg.Backend.HttpBackend.OriginRequestHeaders,
+				Geofencing:           cfg.Backend.HttpBackend.Geofencing,
+				Type:                 cdn.PtrString("http"),
```
marceljk: I would suggest to use here as well the value of cfg.
```suggestion
				Type:                 cfg.Backend.HttpBackend.Type,
```

PR #1185, opened 2026-02-03
feat(cdn): implement new bucket backend type

matheuspolitano:
## Description

Integrate the Bucket backend support into the Terraform Provider in order to enable customers to configure Object Storage buckets with authentication as origins for their CDN distributions.

https://jira.schwarz/browse/STACKITCDN-1075

## Checklist

- [x] Issue was linked above


stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -650,6 +713,18 @@ func mapFields(ctx context.Context, distribution *cdn.Distribution, model *Model
 		return core.DiagsToError(diags)
 	}
 
+	// Check if we have existing configuration to preerve secrets
```
cgoetz-inovex:
```suggestion
	// Check if we have existing configuration to preserve secrets
```
minor typo

matheuspolitano: fixed :)

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -273,6 +294,34 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
+										path.MatchRelative().AtParent().AtName("origin_url"),
+									),
+								},
+							},
+							"region": schema.StringAttribute{
+								Optional:    true,
+								Description: schemaDescriptions["config_backend_region"],
+								Validators:  []validator.String{stringvalidator.AlsoRequires((path.MatchRelative().AtParent().AtName("bucket_url")))},
+							},
+							"access_key": schema.StringAttribute{
```
marceljk: is access_key really correct? Because in the api docs is named accessKeyId and not accessKey

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -273,6 +294,34 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
+								Sensitive:   true,
+								Description: schemaDescriptions["config_backend_access_key"],
+								Validators:  []validator.String{stringvalidator.AlsoRequires((path.MatchRelative().AtParent().AtName("bucket_url")))},
+							},
+							"secret_key": schema.StringAttribute{
+								Optional:    true,
+								Sensitive:   true,
+								Description: schemaDescriptions["config_backend_secret_key"],
+								Validators:  []validator.String{stringvalidator.AlsoRequires((path.MatchRelative().AtParent().AtName("bucket_url")))},
+							},
```
marceljk: Move these two attributes to a new SingleNestedAttribute called "credentials", to stay close to the structure of the api.

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -661,6 +724,18 @@ func mapFields(ctx context.Context, distribution *cdn.Distribution, model *Model
 		return core.DiagsToError(diags)
 	}
 
+	// Check if we have existing configuration to preserve secrets
+	var oldConfig distributionConfig
+	if !model.Config.IsNull() && !model.Config.IsUnknown() {
```
marceljk: nitpick:
```suggestion
	if !utils.IsUndefined(model.Config) {
```

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -357,6 +357,112 @@ func TestAccCDNDistributionResource(t *testing.T) {
...
+	const letters = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+	bytes := make([]byte, n)
+	if _, err := cryptoRand.Read(bytes); err != nil {
+		panic("failed to generate random string: " + err.Error())
+	}
+	for i, b := range bytes {
+		bytes[i] = letters[b%byte(len(letters))]
+	}
+	return string(bytes)
+}
```
marceljk: To get a random string with a specific size you can use the following function `acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)` from the module `"github.com/hashicorp/terraform-plugin-testing/helper/acctest"`

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -357,6 +357,112 @@ func TestAccCDNDistributionResource(t *testing.T) {
...
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.backend.type", "bucket"),
+
+					// Bucket Specific Field Checks
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.backend.bucket_url", bucketUrl),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.backend.region", bucketRegion),
+
+					// CRITICAL: Check Credentials Persistence
+					// The API returns null for these, so if this passes,
+					// it means your logic in resource.go (mapFields) correctly
+					// restored the secrets from the Terraform state.
```
marceljk: is this comment a leftover?

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -357,6 +357,112 @@ func TestAccCDNDistributionResource(t *testing.T) {
...
+	// but usually accepts valid-looking strings for async creation.
+	bucketUrl := "https://my-private-bucket.s3.eu-central-1.amazonaws.com"
+	bucketRegion := "eu01"
+	accessKey := strings.ToUpper(randomString(16))
+	secretKey := randomString(40)
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: testutil.TestAccProtoV6ProviderFactories,
+		CheckDestroy:             testAccCheckCDNDistributionDestroy,
+		Steps: []resource.TestStep{
```
marceljk, line 435: Please extend these tests a bit, by adding a teststep for the datasource and check if a distribution with bucket backend can be read correct. And a step where updates are applied to the resources

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: Done :)

stackit/internal/services/cdn/cdn_acc_test.go
matheuspolitano: Yep, i removed :)

stackit/internal/services/cdn/cdn_acc_test.go
matheuspolitano: Done :)

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: Done :)

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -357,6 +358,133 @@ func TestAccCDNDistributionResource(t *testing.T) {
...
+	`, configBucketResources(bucketName, credentialsGroupName))
+}
+
+func TestAccCDNDistributionBucketResource(t *testing.T) {
+	bucketName := fmt.Sprintf("tf-acc-bucket-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))
+	credentialsGroupName := fmt.Sprintf("tf-acc-group-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))
+
+	expectedBucketUrl := fmt.Sprintf("https://%s.object.storage.eu01.onstackit.cloud", bucketName)
+
+	resource.Test(t, resource.TestCase{
```
marceljk, line 432: Thanks for extending the test with a real bucket. I really like it! Your PR looks overall good to me, but the last thing what I'm missing is, that this test will be extended a bit.
So to be specific, I would like to see a few more checks if all attribute are correctly set, similar like here:
https://github.com/stackitcloud/terraform-provider-stackit/blob/d8bf391f1881b4b46f38c5a39e6603d157c31203/stackit/internal/services/cdn/cdn_acc_test.go#L184-L210
So not just the newly added attributes should be checked, but also if the existing attributes are still correctly set (of course except the fields which relates to the http config).

And I would like to see an update step, similar like here
https://github.com/stackitcloud/terraform-provider-stackit/blob/d8bf391f1881b4b46f38c5a39e6603d157c31203/stackit/internal/services/cdn/cdn_acc_test.go#L313-L343

I think the easiest way would be to change the name, because then the credentials and the bucket should be recreated and it should apply an update on the cdn_distribution

marceljk: Got also an error when doing manually an update 


```
╷
│ Error: Update CDN distribution
│ 
│   with stackit_cdn_distribution.distribution,
│   on main.tf line 80, in resource "stackit_cdn_distribution" "distribution":
│   80: resource "stackit_cdn_distribution" "distribution" {
│ 
│ Error mapping config
╵
```

PR #1198, opened 2026-02-05
STACKITLB-798 | add ALB TF provider

david-mey-STACKIT:
## Description

[STACKITLB-798](https://jira.schwarz/browse/STACKITLB-798)

Add Application Load Balancer provider (API: https://docs.api.stackit.cloud/documentation/alb/version/v2)

## Checklist

- [x] Issue was linked above


go.mod
```diff
@@ -40,6 +41,7 @@ require (
 	github.com/stackitcloud/stackit-sdk-go/services/sqlserverflex v1.3.1
 	github.com/teambition/rrule-go v1.8.2
 	golang.org/x/mod v0.30.0
+	k8s.io/utils v0.0.0-20260108192941-914a6e750570
```
marceljk: Please don't add a new module just for a to pointer utils function.
We already have the same function in the core module of our SDK. Replace the `ptr.to()` to `utils.Ptr()`

docs/data-sources/alb.md
```diff
@@ -0,0 +1,273 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "stackit_alb Data Source - stackit"
+subcategory: ""
+description: |-
+  Setting up supporting infrastructure
+  The example below creates the supporting infrastructure using the STACKIT Terraform provider, including the network, network interface, a public IP address and server resources.
+---
+
+# stackit_alb (Data Source)
```
marceljk: Like @h3adex already suggested, I would prefer as well that the datasource and resource is named application_load_balancer. This makes it easier for first time users to understand for what resource and datasource

docs/data-sources/alb.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Setting up supporting infrastructure
+
+
+The example below creates the supporting infrastructure using the STACKIT Terraform provider, including the network, network interface, a public IP address and server resources.
```
marceljk: The description doesn't fit to the datasource

stackit/internal/services/alb/alb_acc_test.go
marceljk, line 1: max acc test is failing

```console
    alb_acc_test.go:305: Step 1/4 error: Error running apply: exit status 1
        
        Error: Error updating Application Load Balancer
        
          with stackit_alb.loadbalancer,
          on terraform_plugin_test.tf line 169, in resource "stackit_alb" "loadbalancer":
         169: resource "stackit_alb" "loadbalancer" {
        
        Calling API for create: 400 Bad Request, status code 400, Body:
        {
          "code": 3,
          "message": "one or more fields are invalid",
          "details": [
            {
              "@type": "type.googleapis.com/google.rpc.BadRequest.FieldViolation",
              "field": "loadBalancer.listeners[0].https.certificateConfig.certificateIds[0]",
              "description": "The specified certificate \"name-v1-8c81bd317af8a03b8ef0851ccb074eb17d1ad589b540446244a5e593f78ef820\" does not exist."
            }
          ]
        }
--- FAIL: TestAccALBResourceMax (117.56s)
```

stackit/internal/services/alb/utils/util.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils"
+)
+
+func ConfigureClient(ctx context.Context, providerData *core.ProviderData, diags *diag.Diagnostics) *alb.APIClient {
+	apiClientConfigOptions := []config.ConfigurationOption{
+		config.WithCustomAuth(providerData.RoundTripper),
+		utils.UserAgentConfigOption(providerData.Version),
+	}
+	if providerData.LoadBalancerCustomEndpoint != "" {
+		apiClientConfigOptions = append(apiClientConfigOptions, config.WithEndpoint(providerData.LoadBalancerCustomEndpoint))
```
marceljk: wrong custom endpoint

stackit/internal/core/core.go
```diff
@@ -37,6 +37,7 @@ type ProviderData struct {
 	GitCustomEndpoint               string
 	IaaSCustomEndpoint              string
 	KMSCustomEndpoint               string
+	ALBCustomEndpoint               string
```
marceljk: Please keep the order of the CustomEndpoint attributes alphabetic

docs/data-sources/alb.md
```diff
@@ -0,0 +1,273 @@
...
+- `listeners` (Attributes List) List of all listeners which will accept traffic. Limited to 20. (see [below for nested schema](#nestedatt--listeners))
+- `load_balancer_security_group` (Attributes) Security Group permitting network traffic from the LoadBalancer to the targets. Useful when disableTargetSecurityGroupAssignment=true to manually assign target security groups to targets. (see [below for nested schema](#nestedatt--load_balancer_security_group))
+- `networks` (Attributes Set) List of networks that listeners and targets reside in. (see [below for nested schema](#nestedatt--networks))
+- `options` (Attributes) Defines any optional functionality you want to have enabled on your Application Load Balancer. (see [below for nested schema](#nestedatt--options))
+- `plan_id` (String) Service Plan configures the size of the Application Load Balancer. Possible values are: `p10`.. This list can change in the future. Therefore, this is not an enum.
+- `private_address` (String)
+- `region` (String) The resource region. If not defined, the provider region is used. Possible values are: `eu01`, `eu02`.
```
marceljk: Please don't add the possible values for region here and also remove the validation behind it. There will be in the future more regions and if customer want to pin to an older provider version but want to use a new upcoming region, it will be not possible and the docs will be incorrect.

docs/data-sources/alb.md
```diff
@@ -0,0 +1,273 @@
...
+- `id` (String) Terraform's internal resource ID. It is structured as "`project_id`","region","`name`".
+- `labels` (Map of String) Labels represent user-defined metadata as key-value pairs. Label count cannot exceed 64 per ALB.
+- `listeners` (Attributes List) List of all listeners which will accept traffic. Limited to 20. (see [below for nested schema](#nestedatt--listeners))
+- `load_balancer_security_group` (Attributes) Security Group permitting network traffic from the LoadBalancer to the targets. Useful when disableTargetSecurityGroupAssignment=true to manually assign target security groups to targets. (see [below for nested schema](#nestedatt--load_balancer_security_group))
+- `networks` (Attributes Set) List of networks that listeners and targets reside in. (see [below for nested schema](#nestedatt--networks))
+- `options` (Attributes) Defines any optional functionality you want to have enabled on your Application Load Balancer. (see [below for nested schema](#nestedatt--options))
+- `plan_id` (String) Service Plan configures the size of the Application Load Balancer. Possible values are: `p10`.. This list can change in the future. Therefore, this is not an enum.
```
marceljk:
```suggestion
- `plan_id` (String) Service Plan configures the size of the Application Load Balancer.
```
On the datasource the information which values are possible is not necessary. If you want you can leave it there, but it's not necessary.
If you have a link to some docs with the valid plans and maybe some more information, it would be nice if you can add it here

docs/data-sources/alb.md
```diff
@@ -0,0 +1,273 @@
...
+
+- `name` (String) Application Load balancer name.
+- `project_id` (String) STACKIT project ID to which the Application Load Balancer is associated.
+
+### Read-Only
+
+- `disable_target_security_group_assignment` (Boolean) Disable target security group assignemt to allow targets outside of the given network. Connectivity to targets need to be ensured by the customer, including routing and Security Groups (targetSecurityGroup can be assigned). Not changeable after creation.
+- `errors` (Attributes Set) Reports all errors a Application Load Balancer has. (see [below for nested schema](#nestedatt--errors))
+- `external_address` (String) The external IP address where this Application Load Balancer is exposed. Not changeable after creation.
+- `id` (String) Terraform's internal resource ID. It is structured as "`project_id`","region","`name`".
```
marceljk:
```suggestion
- `id` (String) Terraform's internal resource ID. It is structured as "`project_id`,`region`,`name`".
```

docs/data-sources/application_load_balancer.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Required
+
+- `name` (String) Application Load balancer name.
+- `project_id` (String) STACKIT project ID to which the Application Load Balancer is associated.
+
+### Read-Only
+
+- `disable_target_security_group_assignment` (Boolean) Disable target security group assignemt to allow targets outside of the given network. Connectivity to targets need to be ensured by the customer, including routing and Security Groups (targetSecurityGroup can be assigned). Not changeable after creation.
+- `errors` (Attributes Set) Reports all errors a Application Load Balancer has. (see [below for nested schema](#nestedatt--errors))
```
marceljk, line 33: Are the errors really needed in terraform? For the network load balancer we also don't pass the errors

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+		"region":     "The resource region. If not defined, the provider region is used. " + utils.FormatPossibleValues(regionOptions...),
+		"disable_target_security_group_assignment": "Disable target security group assignemt to allow targets outside of the given network. Connectivity to targets need to be ensured by the customer, including routing and Security Groups (targetSecurityGroup can be assigned). Not changeable after creation.",
+		"errors":                                 "Reports all errors a Application Load Balancer has.",
+		"errors.type":                            "Enum: \"TYPE_UNSPECIFIED\" \"TYPE_INTERNAL\" \"TYPE_QUOTA_SECGROUP_EXCEEDED\" \"TYPE_QUOTA_SECGROUPRULE_EXCEEDED\" \"TYPE_PORT_NOT_CONFIGURED\" \"TYPE_FIP_NOT_CONFIGURED\" \"TYPE_TARGET_NOT_ACTIVE\" \"TYPE_METRICS_MISCONFIGURED\" \"TYPE_LOGS_MISCONFIGURED\"\nThe error type specifies which part of the Application Load Balancer encountered the error. I.e. the API will not check if a provided public IP is actually available in the project. Instead the Application Load Balancer with try to use the provided IP and if not available reports TYPE_FIP_NOT_CONFIGURED error.",
```
marceljk: please use `utils.FormatPossibleValues` like in the other descriptions and use `albSdk.AllowedLoadBalancerErrorTypesEnumValues` to get the valid values. No one wants to maintain this manually

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+		"observability_metrics":                  "Observability metrics configuration.",
+		"observability_metrics_credentials_ref":  "Credentials reference for metrics. This reference is created via the observability create endpoint and the credential needs to contain the basic auth username and password for the metrics solution the push URL points to. Then this enables monitoring via remote write for the Application Load Balancer.",
+		"observability_metrics_push_url":         "The Observability(Metrics)/Prometheus remote write push URL you want the metrics to be shipped to.",
+		"plan_id":                                "Service Plan configures the size of the Application Load Balancer. " + utils.FormatPossibleValues(servicePlanOptions...) + ". This list can change in the future. Therefore, this is not an enum.",
+		"private_network_only":                   "Application Load Balancer is accessible only via a private network ip address. Not changeable after creation.",
+		"status":                                 "Enum: \"STATUS_UNSPECIFIED\" \"STATUS_PENDING\" \"STATUS_READY\" \"STATUS_ERROR\" \"STATUS_TERMINATING\"",
```
marceljk: I'm not sure if it really makes sense to add the status. Our waithandler always waits during creation or update, until the the alb is in status "ready". We don't pass for any resource/datasource the status to the users.

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+	if !ok {
+		return
+	}
+
+	apiClient := albUtils.ConfigureClient(ctx, &r.providerData, &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	r.client = apiClient
+	tflog.Info(ctx, "Load balancer client configured")
```
marceljk:
```suggestion
	tflog.Info(ctx, "Application Load balancer client configured")
```

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+}
+
+// Schema defines the schema for the resource.
+func (r *albDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	protocolOptions := []string{"PROTOCOL_UNSPECIFIED", "PROTOCOL_HTTP", "PROTOCOL_HTTPS"}
```
marceljk: instead of defining here all valid values by hand, the generated enum list of the sdk can be used
`albSdk.AllowedListenerProtocolEnumValues`

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+	roleOptions := []string{"ROLE_UNSPECIFIED", "ROLE_LISTENERS_AND_TARGETS", "ROLE_LISTENERS", "ROLE_TARGETS"}
```
marceljk: `albSdk.AllowedNetworkRoleEnumValues`

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+	servicePlanOptions := []string{"p10"}
+	regionOptions := []string{"eu01", "eu02"}
```
marceljk:
```suggestion
```

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+							Computed:    true,
+						},
+					},
+				},
+			},
+			"options": schema.SingleNestedAttribute{
+				Description: descriptions["options"],
+				Computed:    true,
+				Attributes: map[string]schema.Attribute{
+					"acl": schema.SetAttribute{
```
marceljk: Please be close to the api names. And this is actually an objects which has a property `allowed_source_ranges`. Keep the struct like in the api, because when the api changes in the future, we can't adjust the provider like the api is adjusted
```suggestion
					"access_control": schema. SingleNestedAttribute{
```

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (r *albDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) { // nolint:gocritic // function signature required by Terraform
+	var model Model
+	diags := req.Config.Get(ctx, &model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
marceljk, line 547: Add this function here. You may need to update your branch
```suggestion
	}
	
	ctx = core.InitProviderContext(ctx)
	
```

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+			err,
+			"Reading application load balancer",
+			fmt.Sprintf("Load balancer with name %q does not exist in project %q.", name, projectId),
+			map[int]string{
+				http.StatusForbidden: fmt.Sprintf("Project with ID %q not found or forbidden access", projectId),
+			},
+		)
+		resp.State.RemoveResource(ctx)
+		return
+	}
```
marceljk, line 572:
```suggestion
	}
	
	ctx = core.LogResponse(ctx)
```

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+	ctx = tflog.SetField(ctx, "region", region)
+
+	albResp, err := r.client.GetLoadBalancer(ctx, projectId, region, name).Execute()
+	if err != nil {
+		utils.LogError(
+			ctx,
+			&resp.Diagnostics,
+			err,
+			"Reading application load balancer",
+			fmt.Sprintf("Load balancer with name %q does not exist in project %q.", name, projectId),
```
marceljk:
```suggestion
			fmt.Sprintf("Application load balancer with name %q does not exist in project %q.", name, projectId),
```

stackit/internal/services/alb/applicationloadbalancer/datasource.go
```diff
@@ -0,0 +1,587 @@
...
+		return
+	}
+
+	// Set refreshed state
+	diags = resp.State.Set(ctx, model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	tflog.Info(ctx, "Load balancer read")
```
marceljk:
```suggestion
	tflog.Info(ctx, "Application load balancer read")
```

go.mod
david-mey-STACKIT: Ah, thx didn't know this existed

docs/data-sources/alb.md
david-mey-STACKIT: Ya, will change it. I dislike that its so long, but i get why we need it

stackit/internal/services/alb/alb_acc_test.go
david-mey-STACKIT: Ah, yes thats a resource I wanted to test that not in TF yet. I'll take it out for now

docs/data-sources/alb.md
david-mey-STACKIT: Good point, will change

docs/data-sources/alb.md
david-mey-STACKIT: Removed.
We have an API endpoint: https://docs.api.stackit.cloud/documentation/alb/version/v2#tag/Project/operation/APIService_ListPlans
Should I link it?

docs/data-sources/application_load_balancer.md
david-mey-STACKIT: I think yes. We do have things that we cannot test/check during creation/update calls in the API. Partly due to  OpenStack being too slow and partly because some things can only be checked in the real network of the ALB VM. So, people should have this information in order to know that something is not working as expected and maybe have alerts or automated fixes for their setup.

stackit/internal/services/alb/applicationloadbalancer/datasource.go
david-mey-STACKIT: Ah yes, thx. didn't realize it was already in the SDK

docs/data-sources/alb.md
marceljk: Yes you can link it. You can also add a hint, that our cli can be used to fetch the available plans with `stackit beta alb plans`

stackit/internal/services/alb/applicationloadbalancer/datasource.go
david-mey-STACKIT: okay, removed

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	r.client = apiClient
+	tflog.Info(ctx, "Application Load Balancer client configured")
+}
+
+// Schema defines the schema for the resource.
+func (r *applicationLoadBalancerResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	protocolOptions := albUtils.ToStringList(albSdk.AllowedListenerProtocolEnumValues)
+	roleOptions := albUtils.ToStringList(albSdk.AllowedNetworkRoleEnumValues)
+	errorOptions := albUtils.ToStringList(albSdk.AllowedLoadBalancerErrorTypesEnumValues)
+	servicePlanOptions := []string{"p10"}
```
marceljk: same like mentioned for the datasource. check if the sdk has already the enums as list and use it here to decrease maintenance overhead

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	}
+}
+
+func (r *applicationLoadBalancerResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	// Do nothing!
+	// We do not validate the TF input, because the API does that and
+	// maintaining and syncing between them is not worth it, because
+	// 400 Bad Request error gives all the details the user needs via the API in TF, which
+	// also is the single source of truth.
+}
```
marceljk: This is an optional method which is not mandatory to implement for the terraform provider. So you can just remove it :)

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+				},
+			},
+			"region": schema.StringAttribute{
+				Description: descriptions["region"],
+				Optional:    true,
+				// must be computed to allow for storing the override value from the provider
+				Computed: true,
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(4, 6),
+				},
```
marceljk: remove this validator

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"description": schema.StringAttribute{
+							Description: descriptions["errors.description"],
+							Computed:    true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
```
marceljk, line 533: These attributes shouldn't need the plan modifier UseStateForUnknown

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+		"observability_logs_push_url":            "The Observability(Logs)/Loki remote write Push URL you want the logs to be shipped to.",
+		"observability_metrics":                  "Observability metrics configuration.",
+		"observability_metrics_credentials_ref":  "Credentials reference for metrics. This reference is created via the observability create endpoint and the credential needs to contain the basic auth username and password for the metrics solution the push URL points to. Then this enables monitoring via remote write for the Application Load Balancer.",
+		"observability_metrics_push_url":         "The Observability(Metrics)/Prometheus remote write push URL you want the metrics to be shipped to.",
+		"plan_id":                                "Service Plan configures the size of the Application Load Balancer. " + utils.FormatPossibleValues(servicePlanOptions...) + ". See available plans via STACKIT CLI 'stackit alb plans' or API https://docs.api.stackit.cloud/documentation/alb/version/v2#tag/Project/operation/APIService_ListPlans",
```
marceljk: Currently we have the alb plans still as a beta command
```suggestion
		"plan_id":                                "Service Plan configures the size of the Application Load Balancer. " + utils.FormatPossibleValues(servicePlanOptions...) + ". See available plans via STACKIT CLI 'stackit beta alb plans' or API https://docs.api.stackit.cloud/documentation/alb/version/v2#tag/Project/operation/APIService_ListPlans",
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+					mapvalidator.KeysAre(stringvalidator.LengthBetween(1, 63)),
+					mapvalidator.ValueStringsAre(stringvalidator.LengthBetween(1, 63)),
+				},
+			},
+			"plan_id": schema.StringAttribute{
+				Description: descriptions["plan_id"],
+				Required:    true,
+				Validators: []validator.String{
+					stringvalidator.OneOf(servicePlanOptions...),
+				},
```
marceljk, line 571: I would suggest to not validate the plans in the terraform provider
```suggestion
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+				Description: descriptions["listeners"],
+				Required:    true,
+				Validators: []validator.List{
+					listvalidator.SizeBetween(1, 20),
+				},
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"name": schema.StringAttribute{
+							Description: descriptions["listeners.name"],
+							Computed:    true, // will be required in v2
```
marceljk: Currently the name can not be set in terraform

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+							Validators: []validator.String{
+								stringvalidator.RegexMatches(
+									regexp.MustCompile(`^[0-9a-z](?:(?:[0-9a-z]|-){0,61}[0-9a-z])?$`),
+									"1-63 characters [0-9] & [a-z] also [-] but not at the beginning or end",
+								),
+							},
+						},
+						"http": schema.SingleNestedAttribute{
+							Description: "Configuration for HTTP traffic.",
+							Required:    true,
```
marceljk, line 617: Just to be sure that this is correct here. Is `http` always required even if `https` is configured?

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+							Attributes: map[string]schema.Attribute{
+								"hosts": schema.ListNestedAttribute{
+									Description: descriptions["hosts"],
+									Required:    true,
+									Validators: []validator.List{
+										listvalidator.SizeBetween(1, 100),
+									},
```
marceljk, line 624: In the api spec I can't that this is limited to 100. Is it possible that this increases in the future?

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+									NestedObject: schema.NestedAttributeObject{
+										Attributes: map[string]schema.Attribute{
+											"host": schema.StringAttribute{
+												Description: descriptions["host"],
+												Required:    true,
+												Validators: []validator.String{
+													stringvalidator.LengthBetween(1, 253),
+												},
```
marceljk, line 632: same here and in the following nested attributes. In the api docs I don't see these limitations and in the empty ValidateConfig function you wrote:

> 	// We do not validate the TF input, because the API does that and
	// maintaining and syncing between them is not worth it, because
	// 400 Bad Request error gives all the details the user needs via the API in TF, which
	// also is the single source of truth.
	
I'm fine with adding some kind of validators, like `stringvalidator.LengthAtLeast(1)`, but they shouldn't be too strict and should be still valid if the api changes. In this specific case they seems to be random for me, because the api docs doesn't mention these limitations.

Please check all other validators again.

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+
+	// Generate API request body from model
+	payload, err := toCreatePayload(ctx, &model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating Application Load Balancer", fmt.Sprintf("Payload for create: %v", err))
+		return
+	}
+
+	// Create a new Application Load Balancer
+	createResp, err := r.client.CreateLoadBalancer(ctx, projectId, region).CreateLoadBalancerPayload(*payload).XRequestID(uuid.NewString()).Execute()
```
marceljk: I see that this is similar in the network load balancer, but why is the XRequestId set with a random uuid, which isn't stored or logged?

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+
+// Create creates the resource and sets the initial Terraform state.
+func (r *applicationLoadBalancerResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) { // nolint:gocritic // function signature required by Terraform
+	// Retrieve values from plan
+	var model Model
+	diags := req.Plan.Get(ctx, &model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
marceljk, line 1150: You need to update your branch to get this core function
```suggestion
	}
	
	ctx = core.InitProviderContext(ctx)

```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	payload, err := toCreatePayload(ctx, &model)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating Application Load Balancer", fmt.Sprintf("Payload for create: %v", err))
+		return
+	}
+
+	// Create a new Application Load Balancer
+	createResp, err := r.client.CreateLoadBalancer(ctx, projectId, region).CreateLoadBalancerPayload(*payload).XRequestID(uuid.NewString()).Execute()
+	if err != nil {
+		errStr := prettyApiErr(err)
```
marceljk: Please keep the error handling like we have it for all resources. We have already a ticket in our backlog to improve the error output and want a generic approach which can be applied to all resources / datasources

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating Application Load Balancer", fmt.Sprintf("Calling API for create: %v", errStr))
+		return
+	}
+
```
marceljk, line 1173: You need to update your branch to get this core function
```suggestion

	ctx = core.LogResponse(ctx)
	
```

and write all attributes to the state, which are needed to fetch the resource like described here: https://github.com/stackitcloud/terraform-provider-stackit/blob/459122c88e121535d831f21311b473f87de56b83/.github/docs/contribution-guide/resource.go#L211-L221

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (r *applicationLoadBalancerResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) { // nolint:gocritic // function signature required by Terraform
+	var model Model
+	diags := req.State.Get(ctx, &model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
marceljk, line 1212:
```suggestion
	}
	
	ctx = core.InitProviderContext(ctx)

```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	ctx = tflog.SetField(ctx, "region", region)
+
+	lbResp, err := r.client.GetLoadBalancer(ctx, projectId, region, name).Execute()
+	if err != nil {
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
+		if ok && oapiErr.StatusCode == http.StatusNotFound {
+			resp.State.RemoveResource(ctx)
+			return
+		}
+		errStr := prettyApiErr(err)
```
marceljk:
```suggestion
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading Application Load Balancer", fmt.Sprintf("Calling API: %v", errStr))
+		return
+	}
+
```
marceljk, line 1237:
```suggestion

	ctx = core.LogResponse(ctx)

```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+// Update updates the resource and sets the updated Terraform state on success.
+func (r *applicationLoadBalancerResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) { // nolint:gocritic // function signature required by Terraform
+	// Retrieve values from plan
+	var model Model
+	diags := req.Plan.Get(ctx, &model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
marceljk, line 1264:
```suggestion
	}
	
	ctx = core.InitProviderContext(ctx)
	
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+
+	// Update target pool
+	updateResp, err := r.client.UpdateLoadBalancer(ctx, projectId, region, name).UpdateLoadBalancerPayload(*payload).Execute()
+	if err != nil {
+		errStr := prettyApiErr(err)
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating Application Load Balancer", fmt.Sprintf("Calling API for update: %v", errStr))
+		return
+	}
+
```
marceljk, line 1298:
```suggestion

	ctx = core.LogResponse(ctx)
	
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	ctx = tflog.SetField(ctx, "region", region)
+
+	// Delete Application Load Balancer
+	_, err := r.client.DeleteLoadBalancer(ctx, projectId, region, name).Execute()
+	if err != nil {
+		errStr := prettyApiErr(err)
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error deleting Application Load Balancer", fmt.Sprintf("Calling API for delete: %v", errStr))
+		return
+	}
+
```
marceljk, line 1349:
```suggestion

	ctx = core.LogResponse(ctx)
	
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+}
+
+// Delete deletes the resource and removes the Terraform state on success.
+func (r *applicationLoadBalancerResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) { // nolint:gocritic // function signature required by Terraform
+	var model Model
+	diags := req.State.Get(ctx, &model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
marceljk, line 1331:
```suggestion
	}

	ctx = core.InitProviderContext(ctx)

```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	ctx = tflog.SetField(ctx, "name", name)
+	ctx = tflog.SetField(ctx, "region", region)
+
+	lbResp, err := r.client.GetLoadBalancer(ctx, projectId, region, name).Execute()
+	if err != nil {
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
+		if ok && oapiErr.StatusCode == http.StatusNotFound {
+			resp.State.RemoveResource(ctx)
+			return
+		}
```
marceljk: errors.As can be used here
```suggestion
		var oapiErr *oapierror.GenericOpenAPIError
		if errors.As(err, &oapiErr) {
			if oapiErr.StatusCode == http.StatusNotFound {
				resp.State.RemoveResource(ctx)
				return
			}
		}
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+		Name:                                 conversion.StringValueToPointer(model.Name),
+		Networks:                             networksPayload,
+		Options:                              optionsPayload,
+		PlanId:                               conversion.StringValueToPointer(model.PlanId),
+		TargetPools:                          targetPoolsPayload,
+	}, nil
+}
+
+func toLabelPayload(ctx context.Context, model *Model) (albSdk.CreateLoadBalancerPayloadGetLabelsAttributeType, error) {
+	if model.Labels.IsNull() || model.Labels.IsUnknown() {
```
marceljk: nitpick
```suggestion
	if utils.IsUndefined(model.Labels) {
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	if diags.HasError() {
+		return nil, fmt.Errorf("converting labels: %w", core.DiagsToError(diags))
+	}
+
+	payload := albSdk.CreateLoadBalancerPayloadGetLabelsArgType{}
+	for key, value := range labels {
+		payload[key] = value
+	}
+
+	return &payload, nil
```
marceljk: `CreateLoadBalancerPayloadGetLabelsArgType ` is just a alias typ for map[string]string, so there is no conversion needed and labels can be directly returned

```suggestion
	return &labels, nil
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+		})
+	}
+
+	return &payload, nil
+}
+
+func toHttpPayload(ctx context.Context, listenerModel *listener) (albSdk.ListenerGetHttpAttributeType, error) {
+	if listenerModel.Http.IsNull() || listenerModel.Http.IsUnknown() {
+		return nil, nil
+	}
```
marceljk: isn't http required? If so, we should return an error here

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	payload := albSdk.CreateLoadBalancerPayloadGetLabelsArgType{}
+	for key, value := range labels {
+		payload[key] = value
+	}
+
+	return &payload, nil
+}
+
+func toListenersPayload(ctx context.Context, model *Model) (*[]albSdk.Listener, error) {
+	if model.Listeners.IsNull() || model.Listeners.IsUnknown() {
```
marceljk: nitpick
```suggestion
	if utils.IsUndefined(model.Listeners) {
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+			return nil, fmt.Errorf("converting http payload: %w", err)
+		}
+		httpsPayload, err := toHttpsPayload(ctx, &listenerModel)
+		if err != nil {
+			return nil, fmt.Errorf("converting https payload: %w", err)
+		}
+		payload = append(payload, albSdk.Listener{
+			Http:  httpPayload,
+			Https: httpsPayload,
+			//Name:          conversion.StringValueToPointer(listenerModel.Name), will be added in v2
```
marceljk: should be enabled

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	}
+
+	payload := albSdk.ListenerGetHttpArgType{
+		Hosts: hostsPayload,
+	}
+	return &payload, nil
+}
+
+func toHostsPayload(ctx context.Context, httpModel *httpALB) (albSdk.ProtocolOptionsHTTPGetHostsAttributeType, error) {
+	if httpModel.Hosts.IsNull() || httpModel.Hosts.IsUnknown() {
```
marceljk:
```suggestion
	if utils.IsUndefined(httpModel.Hosts) {
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	hostsModel := []hostConfig{}
+	diags := httpModel.Hosts.ElementsAs(ctx, &hostsModel, false)
+	if diags.HasError() {
+		return nil, fmt.Errorf("converting hosts: %w", core.DiagsToError(diags))
+	}
+
+	payload := albSdk.ProtocolOptionsHTTPGetHostsArgType{}
+	for i := range hostsModel {
+		hostModel := hostsModel[i]
+		if hostModel.Host.IsNull() || hostModel.Host.IsUnknown() {
```
marceljk:
```suggestion
		if utils.IsUndefined(hostModel.Host) {
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+		payload = append(payload, albSdk.HostConfig{
+			Host:  conversion.StringValueToPointer(hostModel.Host),
+			Rules: rulesPayload,
+		})
+	}
+	return &payload, nil
+}
+
+func toRulesPayload(ctx context.Context, hostConfigModel *hostConfig) (albSdk.HostConfigGetRulesAttributeType, error) {
+	if hostConfigModel.Rules.IsNull() || hostConfigModel.Rules.IsUnknown() {
```
marceljk:
```suggestion
	if utils.IsUndefined(hostConfigModel.Rules) {
```
You can apply this to all occurrences of the .IsNull() || .IsUnknown() checks

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+		}
+		queryParametersPayload, err := toQueryParametersPayload(ctx, &ruleModel)
+		if err != nil {
+			return nil, fmt.Errorf("converting rule payload: %w", err)
+		}
+		payload = append(payload, albSdk.Rule{
+			CookiePersistence: cookiePersistencePayload,
+			Headers:           headersPayload,
+			Path:              pathPayload,
+			PathPrefix:        nil, // will be removed in v2
```
marceljk: I saw a lot of v2 comments here. In the latest version of the alb module in our Go SDK, we already have the v2 api. It's at the moment the version v0.9.3. Please bump the alb dependency to use this version and resolve all v2 comments

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2904 @@
...
+	payload := albSdk.CertificateConfig{
+		CertificateIds: certificateConfigSet,
+	}
+	return &payload, nil
+}
+
+func toNetworksPayload(ctx context.Context, model *Model) (*[]albSdk.Network, error) {
+	if model.Networks.IsNull() || model.Networks.IsUnknown() {
+		return nil, nil
+	}
```
marceljk: the network attribute is required, so we should throw an error if it's not defined

stackit/internal/services/alb/utils/util.go
```diff
@@ -0,0 +1,42 @@
...
+	albSdk.ListenerProtocol | albSdk.LoadBalancerErrorTypes | albSdk.NetworkRole
+}
+
+func ToStringList[T sdkEnums](in []T) []string {
+	out := make([]string, len(in))
+	for i, o := range in {
+		out[i] = string(o)
+	}
+	return out
+}
```
marceljk: This isn't needed. We already have a utils function `EnumSliceToStringSlice` in the core module of our SDK, which converts an enum slice to a string slice

See example usage: https://github.com/stackitcloud/terraform-provider-stackit/blob/770998656083e829003cc3ed1c067739cc3033d3/stackit/internal/services/kms/wrapping-key/resource.go#L120

stackit/internal/testutil/testutil.go
```diff
@@ -584,12 +600,16 @@ func CreateDefaultLocalFile() os.File {
...
+	input := string(tmpByteArray)
+
+	// If it's a JSON string (starts and ends with quotes)
+	if strings.HasPrefix(input, `"`) && strings.HasSuffix(input, `"`) {
+		// Unquote converts the "escaped" string back to a raw Go string
+		// interpreting \n as a real newline, \" as a quote, etc.
+		if unquoted, err := strconv.Unquote(input); err == nil {
+			return unquoted
+		}
+	}
```
marceljk, line 693: Did you changed it because there was an issue with the previous implementation? If so, can you extend the unit tests of this function to cover the new case?

stackit/internal/services/alb/testfiles/resource-min.tf
```diff
@@ -0,0 +1,128 @@
...
+  listeners = [{
+    port = var.listener_port
+    http = {
+      hosts = [{
+        host = var.host
+        rules = [{
+          target_pool = var.target_pool_name
+          path = {
+            prefix = var.path_prefix
+          }
```
marceljk: regarding the docs path is an optional field and shouldn't be set in the min test.

stackit/internal/services/alb/testfiles/resource-min.tf
```diff
@@ -0,0 +1,128 @@
...
+  project_id = var.project_id
+  name       = var.loadbalancer_name
+  plan_id    = var.plan_id
+  target_pools = [
+    {
+      name        = var.target_pool_name
+      target_port = var.target_pool_port
+      targets = [
+        {
+          display_name = var.target_display_name
```
marceljk: regarding the docs display_name is optional and therefore not be set in the min test

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -0,0 +1,591 @@
...
+					resource.TestCheckResourceAttr("data.stackit_application_load_balancer.loadbalancer", "target_pools.0.targets.0.display_name", testutil.ConvertConfigVariable(testConfigVarsMin["target_display_name"])),
+					resource.TestCheckResourceAttrSet("data.stackit_application_load_balancer.loadbalancer", "networks.0.network_id"),
+					resource.TestCheckResourceAttrSet("data.stackit_application_load_balancer.loadbalancer", "target_pools.0.targets.0.ip"),
+					resource.TestCheckResourceAttrSet("data.stackit_application_load_balancer.loadbalancer", "external_address"),
+					resource.TestCheckResourceAttrSet("stackit_application_load_balancer.loadbalancer", "target_security_group.id"),
+					resource.TestCheckResourceAttrSet("stackit_application_load_balancer.loadbalancer", "target_security_group.name"),
+					resource.TestCheckResourceAttrSet("stackit_application_load_balancer.loadbalancer", "load_balancer_security_group.id"),
+					resource.TestCheckResourceAttrSet("stackit_application_load_balancer.loadbalancer", "load_balancer_security_group.name"),
```
marceljk: Here you check the resource instead of the datasource

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -0,0 +1,591 @@
...
+	"observability_metrics_push_url":    config.StringVariable("https://metrics.observability.dummy.stackit.cloud"),
+	"observability_credential_name":     config.StringVariable(fmt.Sprintf("tf-acc-l%s", acctest.RandStringFromCharSet(7, acctest.CharSetAlphaNum))),
+	"observability_credential_username": config.StringVariable("obs-cred-username"),
+	"observability_credential_password": config.StringVariable("obs-cred-password"),
+}
+
+func configVarsMinUpdated() config.Variables {
+	tempConfig := make(config.Variables, len(testConfigVarsMin))
+	maps.Copy(tempConfig, testConfigVarsMin)
+	tempConfig["target_pool_port"] = config.StringVariable("5431")
```
marceljk, line 116: aren't there more fields which can be updated in-place in the minimal config? like rename the target pool or change the target pool at all or change the listeners?

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -0,0 +1,591 @@
...
+			},
+			// Update
+			{
+				ConfigVariables: configVarsMinUpdated(),
+				Config:          testutil.ALBProviderConfig() + resourceMinConfig,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "project_id", testutil.ConvertConfigVariable(testConfigVarsMin["project_id"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "name", testutil.ConvertConfigVariable(testConfigVarsMin["loadbalancer_name"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "target_pools.0.target_port", testutil.ConvertConfigVariable(configVarsMinUpdated()["target_pool_port"])),
+				),
```
marceljk, line 313: please check here all fields like in the Create step. With the only difference that here the updated configVars is used

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -0,0 +1,591 @@
...
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			// Update
+			{
+				ConfigVariables: configVarsMaxUpdated(),
+				Config:          testutil.ALBProviderConfig() + resourceMaxConfig,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "project_id", testutil.ConvertConfigVariable(testConfigVarsMax["project_id"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "name", testutil.ConvertConfigVariable(testConfigVarsMax["loadbalancer_name"])),
```
marceljk, line 536: same here. you can basically copy all the checks from the Create step and replace `testConfigVarsMax` by `configVarsMaxUpdated()`

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Sure, will change. Will you take care of updating this once the CLI is GA? Is updating the CLI complex or is there an automated tool for it?

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: made validation more open via regex: ^p\d{2,4}$

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: yes, but will be once SDK for v2 is available

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: yes, I don't like it either but thats how the API was built

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Limits were added recently and are currently not changeable and not planned to be changeable. If anything it might be reduced in the future.

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: The validators are already very loosely defined, anything they check is unlikely to change in the future. The actual validation is done in the API which is much more complex. The validators should just ensure that the TF contains a values which roughly makes sense.

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Historically grown. At some point this was to ensure idempotency for POST requests. But the API Gateway stopped accepting these long ago and adds their own anyway. So, I will remove it at least here.

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Thats great, but when will it be done? The API output is not formated and the important part of what the issue is is cut out if i do the normal error handling. Hence, normal way is unusable atm. Therefore, I will keep this for now.

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: added also to update call

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: What is enabled? The name only will be here in V2 of the ALB api and the SDK is at v2beta2, or not?

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Ah, so v2 is already available, nice. Wasn't when i started with this PR
Updated

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Updated to v2

stackit/internal/services/alb/utils/util.go
david-mey-STACKIT: nice, didn't know ~string this was possible

stackit/internal/testutil/testutil.go
david-mey-STACKIT: The issue was \n wasn't converted, it only took care of quotes

david-mey-STACKIT: added test case

stackit/internal/services/alb/applicationloadbalancer/resource.go
marceljk: Yes, we will take care. Updating the CLI is in my opinion easier than the terraform provider. We actually can also move alb out from the beta command because we have it almost one year integrated and the api behind it is GA. Moving it out is more or less just moving some files and adjust some code examples.
If you want, you can contribute it

stackit/internal/services/alb/applicationloadbalancer/resource.go
marceljk: Thanks, in the update call it's not necessary to write the IDs directly into the state, because terraform knows them already from the create run and in case the update fails, Terraform can still access the instance. 

This is just in the Create call important, because if for example the waiter fails, terraform will not know any IDs without `SetAndLogStateFields ` and will not know the instance at all.

stackit/internal/services/alb/alb_acc_test.go
david-mey-STACKIT: added

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Thanks for taking care of it. Nice that the CLI is easier. Sorry, we are way behind on other thing (e.g. WAF) and I don't have time to contribute there :S

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Technically yes, but it will not happen anyway. So, It could be just removed, but this nil check makes the other nil checks much easier to test. So, I would leave it like this.

stackit/internal/services/alb/applicationloadbalancer/resource.go
marceljk: Please remove also the LengthBetween validator. We don't have for any resource a validation of the region and I prefer that we keep this consistent for alb

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -404,19 +404,18 @@ func (r *applicationLoadBalancerResource) Configure(ctx context.Context, req res
...
 		"errors":                                 "Reports all errors a Application Load Balancer has.",
-		"errors.type":                            "Enum: \"TYPE_UNSPECIFIED\" \"TYPE_INTERNAL\" \"TYPE_QUOTA_SECGROUP_EXCEEDED\" \"TYPE_QUOTA_SECGROUPRULE_EXCEEDED\" \"TYPE_PORT_NOT_CONFIGURED\" \"TYPE_FIP_NOT_CONFIGURED\" \"TYPE_TARGET_NOT_ACTIVE\" \"TYPE_METRICS_MISCONFIGURED\" \"TYPE_LOGS_MISCONFIGURED\"\nThe error type specifies which part of the Application Load Balancer encountered the error. I.e. the API will not check if a provided public IP is actually available in the project. Instead the Application Load Balancer with try to use the provided IP and if not available reports TYPE_FIP_NOT_CONFIGURED error.",
+		"errors.type":                            "Enum: " + utils.FormatPossibleValues(errorOptions...) + "\nThe error type specifies which part of the Application Load Balancer encountered the error. I.e. the API will not check if a provided public IP is actually available in the project. Instead the Application Load Balancer with try to use the provided IP and if not available reports TYPE_FIP_NOT_CONFIGURED error.",
```
marceljk: First the description and then the possible values
```suggestion
		"errors.type":                            "The error type specifies which part of the Application Load Balancer encountered the error. I.e. the API will not check if a provided public IP is actually available in the project. Instead the Application Load Balancer with try to use the provided IP and if not available reports TYPE_FIP_NOT_CONFIGURED error." + utils.FormatPossibleValues(errorOptions...) ,
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
marceljk: Alright I understand. Can you then move this function into the utils package of the provider, so that if the next service needs something similar, it can reuse your function?

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -0,0 +1,678 @@
...
+		region = testutil.Region
+	}
+	loadbalancersToDestroy := []string{}
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "stackit_loadbalancer" {
+			continue
+		}
+		// loadbalancer terraform ID: = "[project_id],[name]"
+		loadbalancerName := strings.Split(rs.Primary.ID, core.Separator)[1]
+		loadbalancersToDestroy = append(loadbalancersToDestroy, loadbalancerName)
```
marceljk: wrong structure of the ID
```suggestion
		// loadbalancer terraform ID: = "[project_id],[region],[name]"
		loadbalancerName := strings.Split(rs.Primary.ID, core.Separator)[2]
		loadbalancersToDestroy = append(loadbalancersToDestroy, loadbalancerName)
```

examples/resources/stackit_application_load_balancer/resource.tf
```diff
@@ -0,0 +1,201 @@
...
+  region     = "eu01"
+  name       = "example-load-balancer"
+  plan_id    = "p10"
+  // Hint: Automatically create an IP for the ALB lifecycle by setting ephemeral_address = true or use:
+  // external_address = "124.124.124.124"
+  labels = {
+    (var.label_key) = var.label_value
+  }
+  listeners = [{
+    port = 443
```
marceljk, line 112: name for the listener is missing. Got a validation error when I tried it

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: Done

stackit/internal/services/alb/alb_acc_test.go
david-mey-STACKIT: Done

marceljk (dismissed):
Looks good to me. Thanks for you contribution! As soon a second team member has reviewed and approved this PR, it will be merged 😃

stackit/internal/services/alb/testfiles/resource-max.tf
```diff
@@ -0,0 +1,309 @@
...
+}
+
+resource "stackit_loadbalancer_observability_credential" "observer" {
+  project_id   = var.project_id
+  display_name = var.observability_credential_name
+  password     = var.observability_credential_password
+  username     = var.observability_credential_username
+}
+
+resource "stackit_application_load_balancer" "loadbalancer" {
```
Manuelvaas, line 156: Could you add the external_address to the max test?

stackit/internal/services/alb/testfiles/resource-min.tf
```diff
@@ -0,0 +1,124 @@
...
+    stackit_network_interface.network_interface.network_interface_id
+  ]
+  # Explicit dependencies to ensure ordering
+  depends_on = [
+    stackit_network.network,
+    stackit_network_interface.network_interface
+  ]
+}
+
+resource "stackit_application_load_balancer" "loadbalancer" {
```
Manuelvaas, line 78: Could you remove region and external_address from the min test (as they are not flagged as required)

examples/data-sources/stackit_application_load_balancer/data-source.tf
```diff
@@ -0,0 +1,4 @@
+data "stackit_application_load_balancer" "example" {
+  project_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+  name       = "example-load-balancer"
+}
```
Manuelvaas: Please add a line at the end of the file

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -0,0 +1,678 @@
...
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "listeners.0.http.hosts.0.host", testutil.ConvertConfigVariable(testConfigVarsMin["host"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "listeners.0.http.hosts.0.rules.0.target_pool", testutil.ConvertConfigVariable(testConfigVarsMin["target_pool_name"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "listeners.0.protocol", testutil.ConvertConfigVariable(testConfigVarsMin["protocol_http"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "target_pools.0.name", testutil.ConvertConfigVariable(testConfigVarsMin["target_pool_name"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "target_pools.0.target_port", testutil.ConvertConfigVariable(testConfigVarsMin["target_pool_port"])),
+					resource.TestCheckResourceAttrSet("stackit_application_load_balancer.loadbalancer", "networks.0.network_id"),
```
Manuelvaas: You could actually validate that the right ip is set:
```suggestion
					resource.TestCheckResourceAttrPair("stackit_application_load_balancer.loadbalancer", "networks.0.network_id",
						"stackit_network.network", "network_id"),
```
(same for the other TestCheckResourceAttrSet)

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -0,0 +1,678 @@
...
+		ProtoV6ProviderFactories: testutil.TestAccProtoV6ProviderFactories,
+		CheckDestroy:             testAccCheckALBDestroy,
+		Steps: []resource.TestStep{
+			// Creation
+			{
+				ConfigVariables: testConfigVarsMax,
+				Config:          testutil.ALBProviderConfig() + resourceMaxConfig,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// Load balancer instance resource
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "project_id", testutil.ConvertConfigVariable(testConfigVarsMax["project_id"])),
```
Manuelvaas, line 331: You could check region here as well (also in the tests below)
```suggestion
					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "project_id", testutil.ConvertConfigVariable(testConfigVarsMax["project_id"])),
					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "region", testutil.ConvertConfigVariable(testConfigVarsMin["region"])),
```

stackit/internal/services/alb/applicationloadbalancer/resource.go
```diff
@@ -0,0 +1,2874 @@
...
+		Listeners:                            listenersPayload,
+		Name:                                 conversion.StringValueToPointer(model.Name),
+		Networks:                             networksPayload,
+		Options:                              optionsPayload,
+		PlanId:                               conversion.StringValueToPointer(model.PlanId),
+		TargetPools:                          targetPoolsPayload,
+	}, nil
+}
+
+func toLabelPayload(ctx context.Context, model *Model) (albSdk.CreateLoadBalancerPayloadGetLabelsAttributeType, error) { //nolint:gocritic //This needs to be a pointer of a pointe anyway due to how the SDK works
```
Manuelvaas, line 1420: Missing nil check for model, please also adjust all functions below to check all pointers in your arguments

stackit/internal/services/alb/testfiles/resource-max.tf
david-mey-STACKIT: no, because I'm testing the promotion from ephemeral_address to external_address in the max test update

stackit/internal/services/alb/applicationloadbalancer/resource.go
david-mey-STACKIT: It is checked in the called for all of them

go.mod
Manuelvaas, line 1: you should run a `go mod tidy`

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -396,12 +394,12 @@ func TestAccALBResourceMax(t *testing.T) {
 					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "options.observability.metrics.push_url", testutil.ConvertConfigVariable(testConfigVarsMax["observability_metrics_push_url"])),
 					resource.TestCheckResourceAttrSet("stackit_application_load_balancer.loadbalancer", "networks.0.network_id"),
```
Manuelvaas: Those network IDs could be validated as well through TestCheckResourceAttrPair

Manuelvaas (approved):

stackit/internal/services/alb/alb_acc_test.go
```diff
@@ -0,0 +1,678 @@
...
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "options.private_network_only", testutil.ConvertConfigVariable(testConfigVarsMax["private_network_only"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "options.ephemeral_address", testutil.ConvertConfigVariable(testConfigVarsMax["ephemeral_address"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "options.access_control.allowed_source_ranges.0", testutil.ConvertConfigVariable(testConfigVarsMax["acl"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "options.observability.logs.push_url", testutil.ConvertConfigVariable(testConfigVarsMax["observability_logs_push_url"])),
+					resource.TestCheckResourceAttr("stackit_application_load_balancer.loadbalancer", "options.observability.metrics.push_url", testutil.ConvertConfigVariable(testConfigVarsMax["observability_metrics_push_url"])),
+					resource.TestCheckResourceAttrPair("stackit_application_load_balancer.loadbalancer", "networks.0.network_id", "stackit_network.target_network", "network_id"),
```
Manuelvaas: 0 is the listener and 1 the target network: failing ACC Test

PR #1232, opened 2026-02-17
feat(sfs): Improve acc tests with min/max config

Manuelvaas:
## Description

relates to STACKITTPR-441

## Checklist

- [x] Issue was linked above

### DIFF
--- a/stackit/internal/services/sfs/sfs_acc_test.go
+++ b/stackit/internal/services/sfs/sfs_acc_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/stackitcloud/stackit-sdk-go/core/utils"
 	sfs "github.com/stackitcloud/stackit-sdk-go/services/sfs/v1api"
@@ -262,6 +263,11 @@ func TestAccExportPolicyMin(t *testing.T) {
 			{
 				ConfigVariables: testConfigExportPolicyVarsMinUpdated(),
 				Config:          fmt.Sprintf("%s\n%s", testutil.NewConfigBuilder().EnableBetaResources(true).BuildProviderConfig(), resourceExportPolicyMinConfig),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("stackit_sfs_export_policy.exportpolicy", plancheck.ResourceActionReplace),
+					},
+				},
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "project_id", testutil.ProjectId),
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "region", testutil.Region),
@@ -358,6 +364,11 @@ func TestAccExportPolicyMax(t *testing.T) {
 			{
 				ConfigVariables: testConfigExportPolicyVarsMaxUpdated(),
 				Config:          fmt.Sprintf("%s\n%s", testutil.NewConfigBuilder().EnableBetaResources(true).BuildProviderConfig(), resourceExportPolicyMaxConfig),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("stackit_sfs_export_policy.exportpolicy", plancheck.ResourceActionUpdate),
+					},
+				},
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "project_id", testutil.ProjectId),
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "region", testutil.Region),
@@ -465,6 +476,11 @@ func TestAccResourcePoolResourceMin(t *testing.T) {
 			{
 				ConfigVariables: testConfigResourcePoolVarsMinUpdated(),
 				Config:          fmt.Sprintf("%s\n%s", testutil.NewConfigBuilder().EnableBetaResources(true).BuildProviderConfig(), resourceResourcePoolMinConfig),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("stackit_sfs_resource_pool.resourcepool", plancheck.ResourceActionUpdate),
+					},
+				},
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttr("stackit_sfs_resource_pool.resourcepool", "project_id", testutil.ProjectId),
 					resource.TestCheckResourceAttr("stackit_sfs_resource_pool.resourcepool", "region", testutil.Region),
@@ -560,6 +576,11 @@ func TestAccResourcePoolResourceMax(t *testing.T) {
 			{
 				ConfigVariables: testConfigResourcePoolVarsMaxUpdated(),
 				Config:          fmt.Sprintf("%s\n%s", testutil.NewConfigBuilder().EnableBetaResources(true).BuildProviderConfig(), resourceResourcePoolMaxConfig),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("stackit_sfs_resource_pool.resourcepool", plancheck.ResourceActionUpdate),
+					},
+				},
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttr("stackit_sfs_resource_pool.resourcepool", "project_id", testutil.ProjectId),
 					resource.TestCheckResourceAttr("stackit_sfs_resource_pool.resourcepool", "region", testutil.Region),
@@ -660,6 +681,11 @@ func TestAccShareResourceMin(t *testing.T) {
 			{
 				ConfigVariables: testConfigShareVarsMinUpdated(),
 				Config:          fmt.Sprintf("%s\n%s", testutil.NewConfigBuilder().EnableBetaResources(true).BuildProviderConfig(), resourceShareMinConfig),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("stackit_sfs_share.share", plancheck.ResourceActionUpdate),
+					},
+				},
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttr("stackit_sfs_share.share", "project_id", testutil.ProjectId),
 					resource.TestCheckResourceAttr("stackit_sfs_share.share", "region", testutil.Region),
@@ -766,6 +792,11 @@ func TestAccShareResourceMax(t *testing.T) {
 			{
 				ConfigVariables: testConfigShareVarsMaxUpdated(),
 				Config:          fmt.Sprintf("%s\n%s", testutil.NewConfigBuilder().EnableBetaResources(true).BuildProviderConfig(), resourceShareMaxConfig),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("stackit_sfs_share.share", plancheck.ResourceActionUpdate),
+					},
+				},
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttr("stackit_sfs_share.share", "project_id", testutil.ProjectId),
 					resource.TestCheckResourceAttr("stackit_sfs_share.share", "region", testutil.Region),
--- a/stackit/internal/services/sfs/sfs_acc_test.go
+++ b/stackit/internal/services/sfs/sfs_acc_test.go
@@ -68,10 +68,12 @@ var testConfigExportPolicyVarsMax = config.Variables{
 	"name":       config.StringVariable("tf-acc-" + acctest.RandStringFromCharSet(8, acctest.CharSetAlpha)),
 	"rules": config.ListVariable(
 		config.ObjectVariable(map[string]config.Variable{
-			"ip_acl": config.ListVariable(config.StringVariable(exportPolicyMaxIpAcl1), config.StringVariable(exportPolicyMaxIpAcl2)),
-			"order":  config.IntegerVariable(1),
+			"description": config.StringVariable("Some description"),
+			"ip_acl":      config.ListVariable(config.StringVariable(exportPolicyMaxIpAcl1), config.StringVariable(exportPolicyMaxIpAcl2)),
+			"order":       config.IntegerVariable(1),
 		}),
 	),
+	"region": config.StringVariable(testutil.Region),
 }
 
 var testConfigExportPolicyVarsMaxUpdated = func() config.Variables {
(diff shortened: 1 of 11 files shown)


stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,221 +2,100 @@ package sfs_test
...
-	`))
-	var buffer strings.Builder
-	if err := tmpl.ExecuteTemplate(&buffer, "config", configParams); err != nil {
-		panic(fmt.Sprintf("cannot render template: %v", err))
-	}
-	return buffer.String()
-}
+// EXPORT POLICY - MAX
 
 var (
```
rubenhoenle: could be const

stackit/internal/services/sfs/testdata/share-max.tf
```diff
@@ -0,0 +1,35 @@
...
+variable "space_hard_limit_gigabytes" {}
+
+resource "stackit_sfs_resource_pool" "resourcepool" {
+  project_id        = var.project_id
+  name              = var.resource_pool_name
+  availability_zone = "eu01-m"
+  performance_class = "Standard"
+  size_gigabytes    = 512
+  ip_acl            = ["192.168.42.1/32"]
+  region            = "eu01"
```
marceljk: Avoid hardcoded regions in the test configs, because this can potentially fail if we change something in the future. The default region is already configured in the `SFSProviderConfig()` and this should be enough. If someone would change there the region to e.g. "eu02", the test case would fail, because the resource pool would be still in eu02 and can't be used by the sfs_share resource.

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
+)
+
+var testConfigExportPolicyVarsMax = config.Variables{
+	"project_id": config.StringVariable(testutil.ProjectId),
+	"name":       config.StringVariable("tf-acc-" + acctest.RandStringFromCharSet(8, acctest.CharSetAlpha)),
+	"rules": config.ListVariable(
+		config.ObjectVariable(map[string]config.Variable{
+			"ip_acl": config.ListVariable(config.StringVariable(exportPolicyMaxIpAcl1), config.StringVariable(exportPolicyMaxIpAcl2)),
+			"order":  config.IntegerVariable(1),
+		}),
```
marceljk: The rules have some more optional fields (description, read_only, set_uuid and super_user). They should be set in the max test as well.

You can also just add a second rule, which has these fields set and let the first rule like it is, in order to test if the default values are correctly set.

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
+// Share - MIN
 
-func nsfShareConfig(configParams map[string]string) string {
-	tmpl := template.Must(template.New("config").
-		Parse(`
-		{{.providerConfig}}
+var testConfigShareVarsMin = config.Variables{
+	"project_id":                 config.StringVariable(testutil.ProjectId),
+	"name":                       config.StringVariable("tf-acc-" + acctest.RandStringFromCharSet(8, acctest.CharSetAlpha)),
+	"region":                     config.StringVariable("eu01"),
```
marceljk: Looks like the region var isn't used in the config. For the min config I also wouldn't set it because it's optional and will fallback to the default value.
For the max tests this can be set with the value from `testutil.Region`. Otherwise the cleanup stuff will fail

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
+					}
+					`,
+					testutil.SFSProviderConfig(), resourceExportPolicyMinConfig,
+				),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "project_id", testutil.ProjectId),
+					resource.TestCheckResourceAttrSet("stackit_sfs_export_policy.exportpolicy", "id"),
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "name", testutil.ConvertConfigVariable(testConfigExportPolicyVarsMin["name"])),
+
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.#", "0"),
```
marceljk: In the datasource step you doesn't need to check the resource because it was already done in the previous step. Instead you should check if all the datasource fields have the same value like it was in the resource defined.

The easiest way to do it is, to copy the checks from the create step and add the prefix "data." in all checks

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
+					resource.TestCheckResourceAttrSet("stackit_sfs_export_policy.exportpolicy", "id"),
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "name", testutil.ConvertConfigVariable(testConfigExportPolicyVarsMin["name"])),
 
-	`))
-	var buffer strings.Builder
-	if err := tmpl.ExecuteTemplate(&buffer, "config", configParams); err != nil {
-		panic(fmt.Sprintf("cannot render template: %v", err))
-	}
-	return buffer.String()
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.#", "0"),
```
marceljk, line 182: Add here a check if the policy_id and region is set

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
+			},
+			// Update
+			{
+				ConfigVariables: testConfigExportPolicyVarsMinUpdated(),
+				Config:          fmt.Sprintf("%s\n%s", testutil.SFSProviderConfig(), resourceExportPolicyMinConfig),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "project_id", testutil.ProjectId),
+					resource.TestCheckResourceAttrSet("stackit_sfs_export_policy.exportpolicy", "id"),
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "name", testutil.ConvertConfigVariable(testConfigExportPolicyVarsMinUpdated()["name"])),
+
```
marceljk, line 248: add check for policy_id and region

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
+					return fmt.Sprintf("%s,%s,%s", testutil.ProjectId, testutil.Region, policyId), nil
+				},
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			// Update
+			{
+				ConfigVariables: testConfigExportPolicyVarsMinUpdated(),
+				Config:          fmt.Sprintf("%s\n%s", testutil.SFSProviderConfig(), resourceExportPolicyMinConfig),
+				Check: resource.ComposeAggregateTestCheckFunc(
```
marceljk: It's possible to set which action we expect here, because so far we didn't checked during the update, if a replace was trigger or just an in-place update. 
See example from iaas acc test:
https://github.com/stackitcloud/terraform-provider-stackit/blob/285c7e38f5bb5ced2b62b511b84cbf34afdfd716/stackit/internal/services/iaas/iaas_acc_test.go#L5040-L5044 

To be honest, we haven't set this yet in many tests, but I think we should start doing it.

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
+
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.#", "1"),
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.order", "1"),
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.#", "2"),
-					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.0", exportPolicyResource["ip_acl_1"]),
-					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.1", exportPolicyResource["ip_acl_2"]),
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.0", exportPolicyMaxIpAcl1),
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.1", exportPolicyMaxIpAcl2),
 
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.read_only", "false"),
```
marceljk: check for description field is missing

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
 		Steps: []resource.TestStep{
 			// Creation
 			{
-				Config: resourceConfigExportPolicy(),
+				ConfigVariables: testConfigExportPolicyVarsMax,
+				Config:          fmt.Sprintf("%s\n%s", testutil.SFSProviderConfig(), resourceExportPolicyMaxConfig),
 				Check: resource.ComposeAggregateTestCheckFunc(
-					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "project_id", exportPolicyResource["project_id"]),
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "project_id", testutil.ProjectId),
 					resource.TestCheckResourceAttrSet("stackit_sfs_export_policy.exportpolicy", "id"),
```
marceljk, line 269: check for policy_id is missing

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -2,263 +2,334 @@ package sfs_test
...
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.order", "1"),
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.#", "2"),
-					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.0", exportPolicyResource["ip_acl_1"]),
-					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.1", exportPolicyResource["ip_acl_2"]),
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.0", exportPolicyMaxIpAcl1),
+					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.ip_acl.1", exportPolicyMaxIpAcl2),
 
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.read_only", "false"),
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.set_uuid", "false"),
 					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "rules.0.super_user", "true"),
```
marceljk: most of the checks here are only checking the resource and not the actual datasource

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -275,25 +346,26 @@ func TestAccExportPolicyResource(t *testing.T) {
 			},
 			// Update
 			{
-				Config: resourceConfigUpdateExportPolicy(),
+				ConfigVariables: testConfigExportPolicyVarsMaxUpdated(),
+				Config:          fmt.Sprintf("%s\n%s", testutil.SFSProviderConfig(), resourceExportPolicyMaxConfig),
 				Check: resource.ComposeAggregateTestCheckFunc(
-					resource.TestCheckResourceAttr("stackit_sfs_export_policy.exportpolicy", "project_id", exportPolicyResource["project_id"]),
```
marceljk, line 280: check for policy_id is missing and description in the rules

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -17,248 +21,343 @@ import (
...
+	"name":       config.StringVariable("tf-acc-" + acctest.RandStringFromCharSet(8, acctest.CharSetAlpha)),
 }
 
-func resourceConfigUpdateExportPolicy() string {
-	return fmt.Sprintf(`
-		%s
+var testConfigExportPolicyVarsMinUpdated = func() config.Variables {
+	updatedConfig := config.Variables{}
+	maps.Copy(updatedConfig, testConfigExportPolicyVarsMin)
+	updatedConfig["name"] = config.StringVariable("tf-acc-updated")
```
marceljk: OPTIONAL: just a personal preference, which makes it easier to see in the Portal/CLI the current state of the acc tests, because the name will be just extended.
```suggestion
	updatedConfig["name"] = config.StringVariable(fmt.Sprintf("%s-updated", testutil.ConvertConfigVariable(updatedConfig["name"])))
```

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -17,248 +21,343 @@ import (
...
-		"name":                       fmt.Sprintf("acc-sfs-share-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)),
-		"project_id":                 testutil.ProjectId,
-		"region":                     "eu01",
-		"space_hard_limit_gigabytes": "42",
-	}
+var testConfigResourcePoolVarsMinUpdated = func() config.Variables {
+	updatedConfig := config.Variables{}
+	maps.Copy(updatedConfig, testConfigResourcePoolVarsMin)
+	updatedConfig["performance_class"] = config.StringVariable("Premium")
+	updatedConfig["acl"] = config.ListVariable(
```
marceljk: "size_gigabytes" can be updated as well

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -17,248 +21,343 @@ import (
...
+	"size_gigabytes":        config.IntegerVariable(500),
+	"snapshots_are_visible": config.BoolVariable(true),
+}
 
+var testConfigResourcePoolVarsMaxUpdated = func() config.Variables {
+	updatedConfig := config.Variables{}
+	maps.Copy(updatedConfig, testConfigResourcePoolVarsMax)
+	updatedConfig["performance_class"] = config.StringVariable("Premium")
+	updatedConfig["snapshots_are_visible"] = config.BoolVariable(false)
+	updatedConfig["acl"] = config.ListVariable(
```
marceljk: size_gigabytes can be updated here as well

stackit/internal/services/sfs/sfs_acc_test.go
```diff
@@ -17,248 +21,343 @@ import (
...
+var testConfigExportPolicyVarsMax = config.Variables{
+	"project_id": config.StringVariable(testutil.ProjectId),
+	"name":       config.StringVariable("tf-acc-" + acctest.RandStringFromCharSet(8, acctest.CharSetAlpha)),
+	"rules": config.ListVariable(
+		config.ObjectVariable(map[string]config.Variable{
+			"description": config.StringVariable("Some description"),
+			"ip_acl":      config.ListVariable(config.StringVariable(exportPolicyMaxIpAcl1), config.StringVariable(exportPolicyMaxIpAcl2)),
+			"order":       config.IntegerVariable(1),
+		}),
+	),
```
marceljk: Please avoid the usage of `config.ListVariable` and `config.ObjectVariable`. It's not wrong to use it but it has the disadvantage, that the values can't be reused and in the checks the values needs to be set manually like here:

https://github.com/stackitcloud/terraform-provider-stackit/blob/918a9ed495a6d77dfbf110019bb80343e368b0fd/stackit/internal/services/sfs/sfs_acc_test.go#L308-L316

Instead the checks should look like this one:
https://github.com/stackitcloud/terraform-provider-stackit/blob/918a9ed495a6d77dfbf110019bb80343e368b0fd/stackit/internal/services/sfs/sfs_acc_test.go#L306

This makes it easier to update the tests in the future, because you need to update a value just once and not at multiple places.

---

To configure objects or lists for your acc tests, you can add additional variables for each of them. Similar like we did it here:
https://github.com/stackitcloud/terraform-provider-stackit/blob/53a36978504717f34b06f951bb8fc65eba69c3c2/stackit/internal/services/iaas/testdata/resource-network-area-region-max.tf#L16-L33

Then it's possible to check the attributes like here, without copying the values:
https://github.com/stackitcloud/terraform-provider-stackit/blob/c7fabc1d71c1babd3741dc9b8884f043b7b2d5d4/stackit/internal/services/iaas/iaas_acc_test.go#L1436-L1449

marceljk (approved):

PR #1246, opened 2026-02-20
feat(opensearch) instance + credential, save IDs immediately after cr…

cgoetz-inovex:
…eate

## Description

relates to #1234

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/.github/docs/contribution-guide/package_test.go
+++ b/.github/docs/contribution-guide/package_test.go
@@ -66,7 +66,7 @@ resource "stackit_foo" "foo" {
 				PreConfig: func() {
 					/* Setup mock responses for refresh and delete.
 					   The refresh response fails with an error, but we want to verify that the URL contains the correct IDs.
-					   After the test TF will automatically destroy the resource. So we set up mocks to simulate a successful dlete.
+					   After the test TF will automatically destroy the resource. So we set up mocks to simulate a successful delete.
 					*/
 					s.Reset(
 						testutil.MockResponse{


marceljk (approved):

Fyusel (approved):

stackit/internal/services/opensearch/opensearch_acc_test.go
```diff
@@ -235,6 +238,153 @@ func TestAccOpenSearchResource(t *testing.T) {
...
+						testutil.MockResponse{
+							Description: "create instance",
+							ToJsonBody: &opensearch.CreateInstanceResponse{
+								InstanceId: utils.Ptr(instanceId),
+							},
+						},
+						testutil.MockResponse{Description: "failing waiter", StatusCode: http.StatusInternalServerError},
+					)
+				},
+				Config:      tfConfig,
```
rubenhoenle: IMO these test cases belong into the unit test files of each resource if we want to keep them. They are no e2e-tests which require credentials.

Furthermore wouldn't you need to verify the fields of the resource here (instance_id, region, project_id, ...) here for this test case to make sense?

cgoetz-inovex: moving theses tests into resource specific test files leads to an import cycle:
```
package github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/opensearch/credential
	imports github.com/stackitcloud/terraform-provider-stackit/stackit/internal/testutil from resource_test.go
	imports github.com/stackitcloud/terraform-provider-stackit/stackit from testutil.go
	imports github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/opensearch/credential from provider.go: import cycle not allowed in test
```
moving mockserver from testutil to its own package doesn't resolve the cycle.

This is done in the 2nd test step:
```
|								expected := fmt.Sprintf("/v1/projects/%s/instances/%s", projectId, instanceId)
								if req.URL.Path != expected {
									t.Errorf(fmt.Sprintf("unexpected URL path: got %s, want %s", req.URL.Path, expected), http.StatusBadRequest)
								}
								w.WriteHeader(http.StatusInternalServerError)
```
Background why its done this way:
- creation fails in waiter, this leaves the IDs in the state
- the test step is successful because of `ExpectError` but checks in this step would not be executed by TFs test framework
- so we need a 2nd test step to check the state after creation
- we use a refresh, TF creates a read call with the IDs saved in the state
- we verify these when mocking the read endpoint

rubenhoenle: Regarding the location of the tests: Got the problem with the import cycle. 

But I hope you still got my point that it's wrong to have them in the e2e acc test files? These tests don't need credentials and don't create real resources, so they should be executed by CI on each PR in my opinion. So a solution must be found here.

--- 

Regarding the validation: This is not trivial to me. Maybe you can explain this using some comments in your code. You're not doing it just for me, we have a lot of contributors which haven't ever developed a Terraform provider. We should keep the entry barrier as low as possible. :sweat_smile:

cgoetz-inovex: Tests are executed by CI on each PR.
So it's just the colocation with the acceptance tests? Should i move them to a new file `<service>/<service>_test.go`?

For the first services, tested this way, I added a comment to the tests themselves. Stopped doing it, in order to not copy the same comment everywhere.
Should I add a section 'Test...SavesIDsOnError - Tests' to the readme instead?

rubenhoenle: > Should i move them to a new file <service>/<service>_test.go?

Sounds good :blush: 

---

> Should I add a section 'Test...SavesIDsOnError - Tests' to the readme instead?

Maybe it's better to have it in our contribution guide: https://github.com/stackitcloud/terraform-provider-stackit/blob/e66353da3c5f6e81384f2067eb80f60b7058b899/CONTRIBUTION.md

There are code examples for the contribution guide which can be extended with a blueprint of this test (note to mark it as "only needed when the API is async"): https://github.com/stackitcloud/terraform-provider-stackit/tree/e66353da3c5f6e81384f2067eb80f60b7058b899/.github/docs/contribution-guide incl

cgoetz-inovex: done, also moved the existing tests in this PR

.github/docs/contribution-guide/package_test.go
```diff
@@ -0,0 +1,91 @@
...
+					)
+				},
+				Config:      tfConfig,
+				ExpectError: regexp.MustCompile("Error creating foo.*"),
+			},
+			{
+				PreConfig: func() {
+					/* Setup mock responses for refresh and delete.
+					   The refresh response fails with an error, but we want to verify that the URL contains the correct IDs.
+					   After the test TF will automatically destroy the resource. So we set up mocks to simulate a successful dlete.
```
rubenhoenle:
```suggestion
					   After the test TF will automatically destroy the resource. So we set up mocks to simulate a successful delete.
```

PR #1250, opened 2026-02-23
feat(iam): attribute to query service- accounts by email_regex

h3adex:
## Description

This PR adds the capability to query service accounts by `email_regex`. This feature addresses the following use case:

Each project with ske enabled has a service account named `eu01-<random-identifier-7-to-10-chars>@ske.sa.stackit.cloud`. When rolling out volume encryption storage classes across a large number of clusters, you need an automated way to retrieve the email address of this service account. This is the immediate use case we need to support. However, additional use cases may arise in the future, as internal STACKIT services also rely on service account creation. The SKE API does not expose the service account email and currently there is no priority to export this as a attribute.

While implementing this PR, I also improved the test coverage for the `parseNameFromEmail` functionality and updated our acceptance tests to follow the latest format. Below is the code I used to test this feature:

```hcl
resource "stackit_service_account" "sa" {
  project_id = "xxx"
  name       = "satest01"
}

data "stackit_service_account" "sa" {
  project_id  = "xxx"
  email_regex = ".*@sa\\.stackit\\.cloud$"
}

data "stackit_service_account" "sa_ske" {
  project_id  = "xxx"
  email_regex = ".*@ske\\.sa\\.stackit\\.cloud$"
}

data "stackit_service_account" "sa_email" {
  project_id = "xxx"
  email      = stackit_service_account.sa.email
}
```

## Checklist

- [x] Issue was linked above
Screenshot 2026-02-23 at 10 53 51

### DIFF
--- a/stackit/internal/services/serviceaccount/account/datasource.go
+++ b/stackit/internal/services/serviceaccount/account/datasource.go
@@ -3,12 +3,15 @@ package account
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/conversion"
 	serviceaccountUtils "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/serviceaccount/utils"
 
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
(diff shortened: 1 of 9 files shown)


stackit/internal/services/serviceaccount/account/resource.go
```diff
@@ -323,10 +323,11 @@ func mapFields(resp *serviceaccount.ServiceAccount, model *Model) error {
 	return nil
 }
 
-// parseNameFromEmail extracts the name component from an email address.
-// The email format must be `name-<random7characters>@sa.stackit.cloud`.
+// parseNameFromEmail extracts the name component from a service account email address.
+// The expected email format is `name-<random7to10characters>@sa.stackit.cloud`
```
h3adex, line 327: I’ve adjusted the regex. When this resource/datasource was first released, the standard was 7 random characters. The API has since increased the length to 8. I adjusted the regex to support up to 10. This function is not necessary, it’s merely a nice-to-have to have the name in the state for imports and datasources.

stackit/internal/services/serviceaccount/account/datasource.go
```diff
@@ -56,15 +65,13 @@ func (r *serviceAccountDataSource) Metadata(_ context.Context, req datasource.Me
...
 	descriptions := map[string]string{
-		"id":         "Terraform's internal resource ID, structured as \"`project_id`,`email`\".",
-		"project_id": "STACKIT project ID to which the service account is associated.",
-		"name":       "Name of the service account.",
-		"email":      "Email of the service account.",
+		"id":          "Terraform's internal resource ID, structured as \"`project_id`,`email`\".",
+		"project_id":  "STACKIT project ID to which the service account is associated.",
+		"name":        "Name of the service account.",
+		"email":       "Email of the service account. Either email or email_regex must be provided.",
+		"email_regex": "Regular expression to match the email of the service account. The first service account matching this regex will be used. Either email or email_regex must be provided.",
```
rubenhoenle, line 72: IMO this email regex field must be a seperate datasource to read all service account emails matching that regex. This isn't the way as it's not reproducible across multiple runs. We don't have any control in which order the service accounts get returned by the API.

PR #1271, opened 2026-03-04
chore(resourcemanager): use new multi API version support SDK structure

rubenhoenle:
relates to STACKITSDK-351

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/stackitcloud/terraform-provider-stackit
 
 go 1.24.0
 
-replace github.com/stackitcloud/stackit-sdk-go/services/resourcemanager => /home/ruben.hoenle/Developer/stackit-sdk-go/services/resourcemanager
-
 require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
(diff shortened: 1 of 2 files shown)


stackit/internal/services/resourcemanager/folder/resource.go
```diff
@@ -206,26 +206,21 @@ func (r *folderResource) Create(ctx context.Context, req resource.CreateRequest,
...
-		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating folder", "Container ID is missing")
-		return
-	}
-
 	// This sleep is currently needed due to the IAM Cache.
 	time.Sleep(10 * time.Second)
 	ctx = utils.SetAndLogStateFields(ctx, &resp.Diagnostics, &resp.State, map[string]any{
-		"container_id": *folderCreateResp.ContainerId,
+		"container_id": folderCreateResp.ContainerId,
 	})
```
marceljk: It doesn't relate to your changes, but this would be a quick fix.

Can you change the order of time.Sleep and the SetAndLogStateFields. And the time.Sleep is context blocking at the moment, so it could potentially result in a force kill of the process, because it doesn't react here on a cancelled timeout from e.g. a SIGTERM. This can overall result is lost data. So the time.Sleep should be replaced by a `time.After()`

So my suggestion is, to change it to this:
```golang
	ctx = utils.SetAndLogStateFields(ctx, &resp.Diagnostics, &resp.State, map[string]any{
		"container_id": folderCreateResp.ContainerId,
	})
	
	select {
	case <-ctx.Done():
		return
	case <-time.After(10 * time.Second): // This sleep is currently needed due to the IAM Cache.
		// continue
	}
```

rubenhoenle: as discussed, will do it in a follow-up

PR #1313, opened 2026-03-19
Refac/stackittpr 229 provider config builder

cgoetz-inovex:
## Description

relates to #1234

## Checklist

- [ ] Issue was linked above


stackit/internal/testutil/testutil.go
```diff
@@ -63,540 +66,199 @@ var (
...
-		return `
-		provider "stackit" {
-			default_region = "eu01"
-		}`
+// NewConfigBuilder creates a new ConfigBuilder with enabled beta resources and region eu01 as default.
+// All custom endpoints defined in TF_ACC_*_CUSTOM_ENDPOINT env vars are also set.
+func NewConfigBuilder() *ConfigBuilder {
+	b := &ConfigBuilder{
+		region:              "eu01",
+		enableBetaResources: true,
```
rubenhoenle, line 160: Do we really want to have beta resources enabled by default? 

Should be exactly the other way around from my understanding.

cgoetz-inovex: The original was opt-in to beta resources, used `true` here to save us some configuring in tests.
Is there a downside to this?

rubenhoenle: Since the beta resources aren't enabled by default in our provider by default I would expect the same in the test suite.

But I'm a general opt-in guy I would say xD

stackit/internal/testutil/testutil.go
```diff
@@ -63,540 +66,199 @@ var (
...
+	endpoint := b.endpoints[service.providerName]
+	if endpoint != "" {
+		opts = append(opts, sdkConf.WithEndpoint(endpoint))
+	}
+	tokenEndPoint := b.endpoints[TokenCustomEndpoint.providerName]
+	if tokenEndPoint != "" {
+		opts = append(opts, sdkConf.WithTokenEndpoint(tokenEndPoint))
+	}
+	if b.region != "" {
+		opts = append(opts, sdkConf.WithRegion(b.region))
```
rubenhoenle, line 259: Actually the `WithRegion` is only needed for a subset of services. 

You could even get an error for global APIs if you call it.

https://github.com/stackitcloud/stackit-sdk-go/blob/2693900b73e07a4fa3eef6ff039c7d08db72fda2/core/config/config.go#L572-L574

rubenhoenle: So I think it has to stay out of here

stackit/internal/testutil/testutil.go
```diff
@@ -63,540 +66,199 @@ var (
...
+	return bs.String()
 }
 
-func ScfProviderConfig() string {
-	if ScfCustomEndpoint == "" {
-		return `
-		provider "stackit" {
-			default_region = "eu01"
-		}`
+func (b *ConfigBuilder) BuildClientOptions(service customEndpointConfig) []sdkConf.ConfigurationOption {
```
rubenhoenle, line 245: Pretty nice one btw, thanks for that! 😊

rubenhoenle (approved):
See comments above, rest looks great

stackit/internal/testutil/testutil.go
cgoetz-inovex: changed in https://github.com/stackitcloud/terraform-provider-stackit/pull/1332

stackit/internal/testutil/testutil.go
cgoetz-inovex: added an additional bool parameter to this func in https://github.com/stackitcloud/terraform-provider-stackit/pull/1332

PR #1314, opened 2026-03-19
feat: Support Federated Identity Providers for service accounts

### LINKED ISSUE #1433: FR: Add workload identity federation for service accounts
## Problem description

I'd like to configure  Workload identity federation for service accounts with the STACKIT Terraform provider, so I have the config in git, instead of manually configuring it through the STACKIT portal (as socumented in https://docs.stackit.cloud/platform/access-and-identity/service-accounts/how-tos/manage-service-account-federations/#create-a-federated-identity-provider).

## Proposed solution

Add the neccessary resources to configure Workload identity federation for service accounts.

## Alternative solutions (optional)

- Manually configuring it through the STACKIT portal (as socumented in https://docs.stackit.cloud/platform/access-and-identity/service-accounts/how-tos/manage-service-account-federations/#create-a-federated-identity-provider).

JorTurFer:
## Description

As STACKIT supports workload identity federation via public API, this PR supports it as part of terraform

## Checklist

Fixes https://github.com/stackitcloud/terraform-provider-stackit/issues/1433


stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,457 @@
...
+	apiResp, err := r.client.DefaultAPI.CreateFederatedIdentityProvider(ctx, projectId, serviceAccountEmail).
+		CreateFederatedIdentityProviderPayload(*payload).
+		Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		ok := errors.As(err, &oapiErr)
+		if ok {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Create", fmt.Sprintf("failed to create federated identity provider: %s", oapiErr.Error()))
+		} else {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Create", fmt.Sprintf("failed to create federated identity provider: %v", err))
```
Fyusel: It should be enough here to just use the core.LogAndAddError().
We are using this in all other Create() methods as well.

e.g.

```
if err != nil {
 core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating federated identity provider", fmt.Sprintf("Calling API: %v", err))
 return
}
```

Fyusel: And I would change the "Create" to "Error creating federated identity provider" for all other occurrences

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,457 @@
...
+	projectId := model.ProjectId.ValueString()
+	serviceAccountEmail := model.ServiceAccountEmail.ValueString()
+	federationId := model.FederationId.ValueString()
+
+	apiResp, err := r.client.DefaultAPI.ListFederatedIdentityProviders(ctx, projectId, serviceAccountEmail).
+		Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		ok := errors.As(err, &oapiErr)
+		if ok {
```
Fyusel: Can be simplified to

```suggestion
		if errors.As(err, &oapiErr) {
```

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,457 @@
...
+			if oapiErr.StatusCode == 404 || oapiErr.StatusCode == 403 {
+				resp.State.RemoveResource(ctx)
+				return
+			}
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Read", fmt.Sprintf("failed to list federated identity providers: %s", oapiErr.Error()))
+		} else {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Read", fmt.Sprintf("failed to list federated identity providers: %v", err))
```
Fyusel: Same here, one Log is enough

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,457 @@
...
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Update", fmt.Sprintf("failed to build update payload: %v", err))
+		return
+	}
+
+	apiResp, err := r.client.DefaultAPI.PartialUpdateServiceAccountFederatedIdentityProvider(ctx, projectId, serviceAccountEmail, federationId).
+		PartialUpdateServiceAccountFederatedIdentityProviderPayload(*payload).
+		Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
```
Fyusel: same here

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,457 @@
...
+		ok := errors.As(err, &oapiErr)
+		if ok {
+			if oapiErr.StatusCode == 404 || oapiErr.StatusCode == 403 {
+				return
+			}
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Delete", fmt.Sprintf("failed to delete federated identity provider: %s", oapiErr.Error()))
+		} else {
+			core.LogAndAddError(ctx, &resp.Diagnostics, "Delete", fmt.Sprintf("failed to delete federated identity provider: %v", err))
```
Fyusel: one Log message is enough

stackit/internal/services/serviceaccount/federated_identity_provider/resource_test.go
```diff
@@ -0,0 +1,209 @@
+package federated_identity_provider_test
+
```
Fyusel, line 2: Can you add the acceptance tests in this file to the serviceaccount_acc_test.go please?

stackit/internal/services/serviceaccount/federated_identity_provider/resource_test.go
```diff
@@ -0,0 +1,209 @@
...
+					resource.TestCheckResourceAttr("stackit_service_account_federated_identity_provider.provider", "assertions.1.value", "user@example.com"),
+					resource.TestCheckResourceAttrSet("stackit_service_account_federated_identity_provider.provider", "id"),
+				),
+			},
+		},
+	})
+}
+
+func testAccFederatedIdentityProviderConfig(name string) string {
+	return fmt.Sprintf(`
```
Fyusel: And we are using _.tf_ files under the _testdata_ folder instead of _Sprintf_. Can you move this code out into the _testdata_ of the serviceaccount service?

stackit/internal/services/serviceaccount/federated_identity_provider/resource_test.go
```diff
@@ -0,0 +1,209 @@
...
+			service_account_email = stackit_service_account.sa.email
+			name                  = "%s"
+			issuer                = "https://example.com"
+		}
+	`, testutil.NewConfigBuilder().EnableBetaResources(true).BuildProviderConfig(), testutil.ProjectId, name)
+}
+
+func testAccFederatedIdentityProviderConfigWithAssertions() string {
+	return fmt.Sprintf(`
+		%s
```
Fyusel: Same here :-)

stackit/internal/services/serviceaccount/federated_identity_provider/mapper_test.go
```diff
@@ -0,0 +1,355 @@
+package federated_identity_provider
+
```
Fyusel: This file should be named: _resource_test.go_

stackit/internal/services/serviceaccount/federated_identity_provider/model.go
```diff
@@ -0,0 +1,21 @@
+package federated_identity_provider
+
+import "github.com/hashicorp/terraform-plugin-framework/types"
+
+// Model describes the resource data model.
+type Model struct {
```
Fyusel: In the other resources the model is included in the _resource.go_

stackit/internal/services/serviceaccount/federated_identity_provider/schema_validators.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (v assertionsValidator) ValidateList(ctx context.Context, req validator.ListRequest, resp *validator.ListResponse) { //nolint:gocritic // function signature required by Terraform
+	// Skip validation when the value is null or unknown, for example during plan with computed values.
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
+
+	// Define a lightweight model to extract only the "item" field from list objects.
+	type assertionModel struct {
```
Fyusel: The same model is defined in _model.go_. You can use this instead of redefining it here.

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,457 @@
...
+		}
+		return
+	}
+}
+
+func (r *ServiceAccountFederatedIdentityProviderResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
```
Fyusel: As far as I can see this function does not return a good error message what IDs(attributes) are needed.

E.g. looking at the other resources we have something like this:

	
    idParts := strings.Split(req.ID, core.Separator)
	if len(idParts) != 3 || idParts[0] == "" || idParts[1] == "" || idParts[2] == "" {
		core.LogAndAddError(ctx, &resp.Diagnostics,
			"Error importing image",
			fmt.Sprintf("Expected import identifier with format: [project_id],[region],[image_id]  Got: %q", req.ID),
		)
		return
	}

	ctx = utils.SetAndLogStateFields(ctx, &resp.Diagnostics, &resp.State, map[string]any{
		"project_id": idParts[0],
		"region":     idParts[1],
		"image_id":   idParts[2],
	})

	tflog.Info(ctx, "Image state imported")

Fyusel (commented):
Hi @JorTurFer,
had a first look 👍 
What about a datasource?

stackit/internal/services/serviceaccount/federated_identity_provider/resource_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+		resource "stackit_service_account" "sa" {
+			project_id = "%s"
+			name       = "test-sa-with-assertions"
```
Fyusel: This name is too long

> Attribute name string length must be at most 20, got: 23

stackit/internal/services/serviceaccount/federated_identity_provider/resource_test.go
```diff
@@ -0,0 +1,209 @@
...
+			assertions = [
+				{
+					item     = "iss"
+					operator = "equals"
+					value    = "https://example.com"
+				},
+				{
+					item     = "sub"
+					operator = "equals"
+					value    = "user@example.com"
```
Fyusel: > The 'assertions' list must contain at least one block where the 'item' field is exactly "aud".

stackit/internal/services/serviceaccount/federated_identity_provider/resource_test.go
```diff
@@ -0,0 +1,209 @@
...
+		resource "stackit_service_account" "sa" {
+			project_id = "%s"
+			name       = "test-sa"
+		}
+
+		resource "stackit_service_account_federated_identity_provider" "provider" {
+			project_id            = stackit_service_account.sa.project_id
+			service_account_email = stackit_service_account.sa.email
+			name                  = "%s"
+			issuer                = "https://example.com"
```
Fyusel: Running this test I get:

> The argument "assertions" is required, but no definition was found.

JorTurFer:
> What about a datasource?

As this is something that we are waiting for integrations, I'd prefer to not depend on the datasource. I'm willing to open a follow up PR adding the datasource but I'm not sure it makes sense as it's not unique, so a customer can just create another one and eventually delete the duplicated, but as I said, if you think that it's useful, it's fine for me and I'm willing to create it in another PR or as part of this if they must be added at same time

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,453 @@
...
+	Item     types.String `tfsdk:"item"`
+	Operator types.String `tfsdk:"operator"`
+	Value    types.String `tfsdk:"value"`
+}
+
+func NewServiceAccountFederatedIdentityProviderResource() resource.Resource {
+	return &ServiceAccountFederatedIdentityProviderResource{}
+}
+
+type ServiceAccountFederatedIdentityProviderResource struct {
```
Fyusel:
```suggestion
type serviceAccountFederatedIdentityProviderResource struct {
```

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,453 @@
...
+								stringvalidator.OneOf("equals"),
+							},
+						},
+						"value": schema.StringAttribute{
+							Required:    true,
+							Description: descriptions["assertions.value"],
+						},
+					},
+				},
+				Required: true,
```
Fyusel, line 148: Is this correct? In the API this is not listed as required (https://docs.api.stackit.cloud/documentation/service-account/version/v2#tag/v2/operation/post-projects-projectId-service-accounts-federations)

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,453 @@
...
+
+func (r *ServiceAccountFederatedIdentityProviderResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+}
+
+func mapFields(ctx context.Context, apiResp *serviceaccount.FederatedIdentityProvider, model *Model, projectId, serviceAccountEmail string) error {
+	if apiResp == nil {
+		return fmt.Errorf("apiResp is nil")
+	}
+
```
Fyusel, line 358: Missing:
if model == nil {
		return fmt.Errorf("model input is nil")
	}

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+		payload.Assertions = assertionsPayload
+	}
+
+	return payload, nil
+}
+
+func toUpdatePayload(ctx context.Context, model *Model) (*serviceaccount.PartialUpdateServiceAccountFederatedIdentityProviderPayload, error) {
+	payload := &serviceaccount.PartialUpdateServiceAccountFederatedIdentityProviderPayload{}
+
```
Fyusel: Missing:
if model == nil {
		return nil, fmt.Errorf("nil model")
	}

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
Fyusel: @JorTurFer This is still open

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
JorTurFer: it is required indeed, there is a validation on backend that rejects the request if there isn't any `assertion` and also if there isn't any assertion for `aud` in specific.

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
```diff
@@ -0,0 +1,458 @@
...
+		return
+	}
+
+	ctx = core.InitProviderContext(ctx)
+
+	projectId := model.ProjectId.ValueString()
+	serviceAccountEmail := model.ServiceAccountEmail.ValueString()
+	federationId := model.FederationId.ValueString()
+
+	apiResp, err := r.client.DefaultAPI.ListFederatedIdentityProviders(ctx, projectId, serviceAccountEmail).
```
Fyusel: We need a GET endpoint here. Then the filtering below can be removed.

Additionally a datasource for thsi GET endpoint can be implemented.
A datasource for the list endpoint is not needed right now since we need to check how a good approach looks like for pagination.

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
JorTurFer: I missed this, sorry. Now it's fixed

stackit/internal/services/serviceaccount/federated_identity_provider/resource.go
Fyusel: Ok, can you update the API doc as well, please

Fyusel (changes requested):
To summarize:
- GET endpoint is needed (Read() in resource needs to be updated)
- datasource for GET endpoint is needed
- datasource for LIST is not needed

stackit/internal/services/serviceaccount/federated_identity_provider/datasource.go
```diff
@@ -0,0 +1,156 @@
...
+			"federation_id": schema.StringAttribute{
+				Description: descriptions["federation_id"],
+				Required:    true,
+				Validators: []validator.String{
+					validate.UUID(),
+				},
+			},
+			"name": schema.StringAttribute{
+				Computed:    true,
+				Description: descriptions["name"],
```
Fyusel, line 70: This string is not existent and is therefore "" (see the generated docs as well).

Either use the one in resource.go if the descriptions are the same or add them here.
(make descriptions global in resource and use it there and in the datasource if possible)

JorTurFer: done!

cgoetz-inovex (approved):

Fyusel (approved):

PR #1333, opened 2026-03-27
feat(cdn): implement redirects

matheuspolitano:
## Description

[STACKITCDN-1104](https://jira.schwarz/browse/STACKITCDN-1104)

In case there is no issue present for your PR, please consider creating one.
At least please give us some description what you are trying to achieve and why your change is needed. -->

relates to #1234

## Checklist

- [x] Issue was linked above
- [x] Code format was applied: `make fmt`
- [x] Examples were added / adjusted (see `examples/` directory)
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI)
- [x] Unit tests got implemented or updated
- [x] Acceptance tests got implemented or updated (see e.g. [here](https://github.com/stackitcloud/terraform-provider-stackit/blob/f5f99d170996b208672ae684b6da53420e369563/stackit/internal/services/dns/dns_acc_test.go))
- [x] Unit tests are passing: `make test` (will be checked by CI)
- [x] No linter issues: `make lint` (will be checked by CI)

PR #1372, opened 2026-04-13
feat(cdn): Add support for configuring WAF

matheuspolitano:
## Description

https://jira.schwarz/browse/STACKITCDN-723

I want to add support for the waf configuration block in the CDN distribution resource,so that users can programmatically manage security tiers, paranoia levels, and granular rule overrides (Enabled/Disabled/Log-only) via Terraform.

## Checklist

- [x] Issue was linked above


matheuspolitano:
@Fyusel  @marceljk  Kindly prioritize the review of PR #[1333](https://github.com/stackitcloud/terraform-provider-stackit/pull/1333), as the current pull request builds upon those changes.

docs/resources/cdn_distribution.md
```diff
@@ -56,6 +56,57 @@ resource "stackit_cdn_distribution" "example_bucket_distribution" {
...
+
+    # WAF Configuration
+    # 
+    # Precedence Hierarchy: Specific Rules > Groups > Collections
+    # In this example, the entire "@builtin/crs/request" collection is ENABLED.
+    # However, because specific Rule IDs have a higher precedence, the rule 
+    # "@builtin/crs/request/942151" is explicitly DISABLED, overriding the collection setting.
+    # 
+    # To view all available collections, groups, and rules, consult the API documentation:
+    # https://internal-docs.api.eu01.stackit.cloud/documentation/cdn/version/v1#tag/WAF/operation/ListWafCollections
```
h3adex, line 86: internal docs are not reachable from external user

docs/data-sources/cdn_distribution.md
```diff
@@ -74,6 +76,58 @@ Read-Only:
...
+- `allowed_http_versions` (List of String) Restricts which HTTP protocol versions are accepted. If provided, the list must contain at least one item. If omitted, the API applies the following defaults: `HTTP/1.0`, `HTTP/1.1`, `HTTP/2`, `HTTP/2.0`.
+- `allowed_request_content_types` (List of String) Restricts which Content-Type headers are accepted in request bodies. If provided, the list must contain at least one item. If omitted, the API applies the following defaults: `application/x-www-form-urlencoded`, `multipart/form-data`, `multipart/related`, `text/xml`, `application/xml`, `application/soap+xml`, `application/x-amf`, `application/json`, `application/octet-stream`, `application/csp-report`, `application/xss-auditor-report`, `text/plain`.
+- `disabled_rule_collection_ids` (List of String) List of WAF Collection IDs explicitly disabled. Can be set to an empty list to clear previously set rules. Precedence hierarchy: Specific Rules override Groups, and Groups override Collections. To view available rule collections, please consult the API documentation: https://internal-docs.api.eu01.stackit.cloud/documentation/cdn/version/v1#tag/WAF/operation/ListWafCollections
```
Fyusel: Same here with the internal-docs
Should all be replaced

docs/data-sources/cdn_distribution.md
```diff
@@ -74,6 +76,58 @@ Read-Only:
...
+- `disabled_rule_group_ids` (List of String) List of WAF Rule Group IDs explicitly disabled. Can be set to an empty list to clear previously set rules. Precedence hierarchy: Specific Rules override Groups, and Groups override Collections. To view available rule groups, please consult the API documentation: https://internal-docs.api.eu01.stackit.cloud/documentation/cdn/version/v1#tag/WAF/operation/ListWafCollections
```
Fyusel: The sentence "Precendence hierarchy" is written for almost every attribute and is almost the same. Does it make sense to have this for every attribute or should this just be mentioned somewhere "globally"?

docs/resources/cdn_distribution.md
```diff
@@ -131,6 +184,67 @@ Optional:
...
+- `allowed_http_versions` (List of String) Restricts which HTTP protocol versions are accepted. If provided, the list must contain at least one item. If omitted, the API applies the following defaults: `HTTP/1.0`, `HTTP/1.1`, `HTTP/2`, `HTTP/2.0`.
+- `allowed_request_content_types` (List of String) Restricts which Content-Type headers are accepted in request bodies. If provided, the list must contain at least one item. If omitted, the API applies the following defaults: `application/x-www-form-urlencoded`, `multipart/form-data`, `multipart/related`, `text/xml`, `application/xml`, `application/soap+xml`, `application/x-amf`, `application/json`, `application/octet-stream`, `application/csp-report`, `application/xss-auditor-report`, `text/plain`.
+- `disabled_rule_collection_ids` (List of String) List of WAF Collection IDs explicitly disabled. Can be set to an empty list to clear previously set rules. Precedence hierarchy: Specific Rules override Groups, and Groups override Collections. To view available rule collections, please consult the API documentation: https://internal-docs.api.eu01.stackit.cloud/documentation/cdn/version/v1#tag/WAF/operation/ListWafCollections
```
Fyusel: Same here

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1132,3 +1784,48 @@ func validateCountryCode(country string) (string, error) {
 
 	return upperCountry, nil
 }
+
+// getSortedWafList extracts strings from HCL list, sorts them and returns the slice
+func getSortedWafList(ctx context.Context, tfList basetypes.ListValue) []string {
```
Fyusel: I would move those three helper functions to utils package

marceljk:
> @Fyusel @marceljk Kindly prioritize the review of PR #[1333](https://github.com/stackitcloud/terraform-provider-stackit/pull/1333), as the current pull request builds upon those changes.

#1333 is merged now. Please rebase this PR

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -268,6 +402,192 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
+							"allowed_http_methods": schema.ListAttribute{
+								Optional:    true,
+								Computed:    true,
+								ElementType: types.StringType,
+								Description: schemaDescriptions["waf_allowed_http_methods"],
+								Default:     listdefault.StaticValue(types.ListValueMust(types.StringType, defaultWafConfigAllowedHttpMethods)),
+								Validators: []validator.List{
+									listvalidator.SizeAtLeast(1),
+								},
+							},
```
marceljk, line 514: Why have these attributes a default value? The API doesn't have them as default value, so I would avoid it here as well. 

These default values could result in potential issues for us. E.g. the api supports a new content_type, it wouldn't be here part of the list. When we update this list with a newer terraform version, it could lead to unexpected behavior for some customers, because then the list contains maybe a content type which they don't want.

Same if for example, the api will not allow anymore HTTP/1.0 for the allowed_http_version. This will break then all older version of the terraform provider.

I would be fine with it, if the API has the same behavior. But based on the api docs, it looks like users need always explicit set a value in these list.

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -268,6 +402,192 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
+								Optional:    true,
+								Computed:    true,
+								Description: schemaDescriptions["waf_type"],
+								Default:     stringdefault.StaticString("FREE"),
+							},
+							"paranoia_level": schema.StringAttribute{
+								Optional:    true,
+								Computed:    true,
+								Description: schemaDescriptions["waf_paranoia_level"],
+								Default:     stringdefault.StaticString("L1"),
```
marceljk: Same here. Based on the api docs these attributes don't have a default value

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -590,35 +965,89 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
...
+			DisabledRuleGroupIds:      []string{},
+			LogOnlyRuleGroupIds:       []string{},
+			EnabledRuleCollectionIds:  []string{},
+			DisabledRuleCollectionIds: []string{},
+			LogOnlyRuleCollectionIds:  []string{},
+			// Intentionally omitted (nil) to avoid the 422 Unprocessable Entity error:
+			// AllowedHttpVersions, AllowedRequestContentTypes, AllowedHttpMethods
+		}
+		configPatch.Waf = &wafPatch
+	}
```
marceljk: Why do you need the state here? Is it needed to set WAFMODE_DISABLED or can `configPatch.Waf` set to `null` to remove it?

Nevertheless, I think you can check here also `configPlanModel.Waf.IsNull()` instead of `!utils.IsUndefined(configStateModel.Waf)` and would have the same outcome but without the need accessing the state

stackit/internal/services/cdn/distribution/resource.go
marceljk: I agree with it. But I would suggest to move the `mapWafListToHCL()` function, to the conversion package 

https://github.com/stackitcloud/terraform-provider-stackit/blob/27ed51088282736f22625d4b6cc15f82eabfe804/stackit/internal/conversion/conversion.go#L1-L263

And please give these functions a more generic name and add unit tests

docs/resources/cdn_distribution.md
matheuspolitano: It's reachable

docs/data-sources/cdn_distribution.md
matheuspolitano: It is possible for external users to access our docs.

stackit/internal/services/cdn/distribution/resource.go

stackit/internal/services/cdn/distribution/resource.go

docs/data-sources/cdn_distribution.md
matheuspolitano: I did this on purpose; to improve it, I moved them to collection_ids.

docs/resources/cdn_distribution.md
matheuspolitano: Same

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: I change the approach, now when it's not defined we let them null and the api manage the default values

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: We are no longer persisting states once the WAF is removed, it is only disabled and change the type to free a pushed the API

stackit/internal/conversion/conversion.go
```diff
@@ -263,3 +263,18 @@ func ParseEphemeralProviderData(ctx context.Context, providerData any, diags *di
 	}
 	return stackitProviderData, true
 }
+
+// SortedStringsToListValue guarantees the returned HCL List is sorted
+func SortedStringsToListValue(items []string) basetypes.ListValue {
+	if len(items) == 0 {
```
marceljk:
```suggestion
func SortedStringsToListValue(items []string) basetypes.ListValue {
	if items == nil {
		return types.ListNull(types.StringType)
	}
	if len(items) == 0 {
```
I would return here a null list in case the string slice is also null. null list != empty list
This will also need some adjustment to the tests

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -403,6 +460,98 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
 					},
+					"waf": schema.SingleNestedAttribute{
+						Description: schemaDescriptions["config_waf"],
+						Optional:    true,
+						Computed:    true,
+						Attributes: map[string]schema.Attribute{
+							"mode": schema.StringAttribute{
+								Required:    true,
+								Description: schemaDescriptions["waf_mode"],
+								Validators:  []validator.String{stringvalidator.OneOf(string(cdnSdk.WAFMODE_DISABLED), string(cdnSdk.WAFMODE_ENABLED), string(cdnSdk.WAFMODE_LOG_ONLY))},
```
marceljk: We have in our SDK also a slice with all the enum. Use it here instead of referencing each enum by hand
```suggestion
								Validators:  []validator.String{stringvalidator.OneOf(sdkUtils.EnumSliceToStringSlice(cdnSdk.AllowedWafModeEnumValues)...)},
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -403,6 +460,98 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
+							},
+							"type": schema.StringAttribute{
+								Required:    true,
+								Description: schemaDescriptions["waf_type"],
+								Validators:  []validator.String{stringvalidator.OneOf(string(cdnSdk.WAFTYPE_PREMIUM), string(cdnSdk.WAFTYPE_FREE))},
```
marceljk: same here

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -403,6 +460,98 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
+							},
+							"paranoia_level": schema.StringAttribute{
+								Optional:    true,
+								Description: schemaDescriptions["waf_paranoia_level"],
+								Validators:  []validator.String{stringvalidator.OneOf(string(cdnSdk.WAFPARANOIALEVEL_L1), string(cdnSdk.WAFPARANOIALEVEL_L2), string(cdnSdk.WAFPARANOIALEVEL_L3), string(cdnSdk.WAFPARANOIALEVEL_L4))},
```
marceljk: same here

stackit/internal/services/cdn/distribution/resource.go
marceljk: But for this you don't need the state. You can just set `WafConfigPatch` always to mode=disabled and type=free, when in the terraform config nothing related to Waf is defined. 

Then you can remove the added stateModel and revert all renamings from Model -> PlanModel. I would actually prefer to avoid here, handling the state and config model, because this can lead in the future to issue when someone mixes them. And this resource is already really big.

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: DOne

stackit/internal/services/cdn/distribution/resource.go

stackit/internal/services/cdn/distribution/resource.go

stackit/internal/conversion/conversion.go
matheuspolitano: Done

stackit/internal/services/cdn/distribution/datasource.go
```diff
@@ -512,6 +593,51 @@ func mapDataSourceFields(ctx context.Context, distribution *cdnSdk.Distribution,
...
+	// Helper to unconditionally map set fields for WAF
+	mapWafSet := func(apiList []string) types.Set {
+		if apiList != nil {
+			setVal, diags := types.SetValueFrom(ctx, types.StringType, apiList)
+			if !diags.HasError() {
+				return setVal
+			}
+		}
+		return types.SetNull(types.StringType)
+	}
```
marceljk: This should be moved into a generic utils function, because this is not waf specific and can be used in other places as well

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -759,10 +915,46 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
 		Redirects:        redirectsConfig,
 	}
 
-	if !utils.IsUndefined(configModel.Optimizer) {
+	modeDisabled := cdnSdk.WafMode(cdnSdk.WAFMODE_DISABLED)
+	typeFree := cdnSdk.WafType(cdnSdk.WAFTYPE_FREE)
+	configPatch.Waf = &cdnSdk.WafConfigPatch{
+		Mode: &modeDisabled,
+		Type: &typeFree,
+	}
```
marceljk: can be simplified
```suggestion
	configPatch.Waf = &cdnSdk.WafConfigPatch{
		Mode: new(cdnSdk.WAFMODE_DISABLED),
		Type: new(cdnSdk.WAFTYPE_FREE),
	}
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1453,3 +1749,17 @@ func validateCountryCode(country string) (string, error) {
...
+		return nil
+	}
+	var elements []string
+	diags := tfSet.ElementsAs(ctx, &elements, true)
+	if diags.HasError() {
+		return []string{}
+	}
+	sort.Strings(elements)
+	return elements
+}
```
marceljk: Why is it needed to sort the slice? Takes the API care of the order? If not, you can use `conversion.StringSetToSlice()` instead this function, there you shouldn't ignore the error like it was done here

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -759,10 +915,46 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
...
+		configPatch.Waf.EnabledRuleGroupIds = getWafSet(ctx, wafModel.EnabledRuleGroupIds)
+		configPatch.Waf.DisabledRuleGroupIds = getWafSet(ctx, wafModel.DisabledRuleGroupIds)
+		configPatch.Waf.LogOnlyRuleGroupIds = getWafSet(ctx, wafModel.LogOnlyRuleGroupIds)
+		configPatch.Waf.EnabledRuleCollectionIds = getWafSet(ctx, wafModel.EnabledRuleCollectionIds)
+		configPatch.Waf.DisabledRuleCollectionIds = getWafSet(ctx, wafModel.DisabledRuleCollectionIds)
+		configPatch.Waf.LogOnlyRuleCollectionIds = getWafSet(ctx, wafModel.LogOnlyRuleCollectionIds)
+
+		if !utils.IsUndefined(wafModel.ParanoiaLevel) {
+			pl := cdnSdk.WafParanoiaLevel(wafModel.ParanoiaLevel.ValueString())
+			configPatch.Waf.ParanoiaLevel = &pl
```
marceljk: can be simplified
```suggestion
			cdnConfig.Waf.ParanoiaLevel = new(cdnSdk.WafParanoiaLevel(wafModel.ParanoiaLevel.ValueString()))
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1117,6 +1309,69 @@ func mapFields(ctx context.Context, distribution *cdnSdk.Distribution, model *Mo
...
+	// Detect if we are running an Import (or Data Source Read) where prior config doesn't exist
+	isImport := utils.IsUndefined(model.Config)
+
+	// Helper to unconditionally map string fields
+	mapWafString := func(apiVal *string) types.String {
+		if apiVal != nil {
+			return types.StringValue(*apiVal)
+		}
+		return types.StringNull()
+	}
```
marceljk: you can use the function `types.StringPointerValue()` from the terraform framework itself instead of this function

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1117,6 +1309,69 @@ func mapFields(ctx context.Context, distribution *cdnSdk.Distribution, model *Mo
...
+	// Helper to unconditionally map set fields
+	mapWafSet := func(apiList []string) types.Set {
+		if apiList != nil {
+			setVal, diags := types.SetValueFrom(ctx, types.StringType, apiList)
+			if !diags.HasError() {
+				return setVal
+			}
+		}
+		return types.SetNull(types.StringType)
+	}
```
marceljk: same like mentioned in the datasource, replace it by a generic utils function. You can use the same function in the datasource and resource

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1117,6 +1309,69 @@ func mapFields(ctx context.Context, distribution *cdnSdk.Distribution, model *Mo
...
+
+	var pl *string
+	if distribution.Config.Waf.ParanoiaLevel != nil {
+		plVal := string(*distribution.Config.Waf.ParanoiaLevel)
+		pl = &plVal
```
marceljk: can be simplified
```suggestion
		pl = new(string(*distribution.Config.Waf.ParanoiaLevel))
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1117,6 +1309,69 @@ func mapFields(ctx context.Context, distribution *cdnSdk.Distribution, model *Mo
...
+	wafObjAttrs["disabled_rule_group_ids"] = mapWafSet(distribution.Config.Waf.DisabledRuleGroupIds)
+	wafObjAttrs["log_only_rule_group_ids"] = mapWafSet(distribution.Config.Waf.LogOnlyRuleGroupIds)
+	wafObjAttrs["enabled_rule_collection_ids"] = mapWafSet(distribution.Config.Waf.EnabledRuleCollectionIds)
+	wafObjAttrs["disabled_rule_collection_ids"] = mapWafSet(distribution.Config.Waf.DisabledRuleCollectionIds)
+	wafObjAttrs["log_only_rule_collection_ids"] = mapWafSet(distribution.Config.Waf.LogOnlyRuleCollectionIds)
+
+	// Determine if WAF should be entirely excluded to prevent drift.
+	// The API can return an empty string for fully unconfigured backends.
+	isWafDisabled := (distribution.Config.Waf.Mode == cdnSdk.WAFMODE_DISABLED || distribution.Config.Waf.Mode == "") &&
+		(distribution.Config.Waf.Type == cdnSdk.WAFTYPE_FREE || distribution.Config.Waf.Type == "")
```
marceljk: Type and Mode will never be an empty string. It's because these fields are declared as enums in the CDN api spec and our SDK has them as enums generated. If a value returns, which isn't a value of the defined enums, it fallbacks to unknown_default_open_api

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -231,9 +245,14 @@ func TestAccCDNDistributionHttp(t *testing.T) {
...
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"domains"}, // we added a domain in the meantime...
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"domains",
+					"config.waf.allowed_http_methods",
+					"config.waf.allowed_http_versions",
+					"config.waf.allowed_request_content_types",
+					"config.waf.paranoia_level",
```
marceljk: these fields can be removed from the ignore list, because they are actually part of the import

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -190,6 +198,12 @@ func TestAccCDNDistributionHttp(t *testing.T) {
 						"DE",
 					),
 					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.optimizer.enabled", testutil.ConvertConfigVariable(testConfigVarsHttp["optimizer"])),
+					// WAF Checks
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.mode", testutil.ConvertConfigVariable(testConfigVarsHttp["waf_mode"])),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.type", testutil.ConvertConfigVariable(testConfigVarsHttp["waf_type"])),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.enabled_rule_ids.#", "1"),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.enabled_rule_ids.0", "@builtin/crs/request/941120"),
+
```
marceljk, line 270: You can add here a lot more checks. You can check here if every waf field is set

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -292,6 +311,13 @@ func TestAccCDNDistributionHttp(t *testing.T) {
 					resource.TestCheckResourceAttr("data.stackit_cdn_distribution.distribution", "config.blocked_countries.#", "1"),
 					resource.TestCheckResourceAttr("data.stackit_cdn_distribution.distribution", "config.blocked_countries.0", "CU"),
 					resource.TestCheckResourceAttr("data.stackit_cdn_distribution.distribution", "config.optimizer.enabled", testutil.ConvertConfigVariable(testConfigVarsHttp["optimizer"])),
+
+					// WAF Checks inside Data Source
+					resource.TestCheckResourceAttr("data.stackit_cdn_distribution.distribution", "config.waf.mode", testutil.ConvertConfigVariable(testConfigVarsHttp["waf_mode"])),
+					resource.TestCheckResourceAttr("data.stackit_cdn_distribution.distribution", "config.waf.type", testutil.ConvertConfigVariable(testConfigVarsHttp["waf_type"])),
+					resource.TestCheckResourceAttr("data.stackit_cdn_distribution.distribution", "config.waf.enabled_rule_ids.#", "1"),
+					resource.TestCheckResourceAttr("data.stackit_cdn_distribution.distribution", "config.waf.enabled_rule_ids.0", "@builtin/crs/request/941120"),
+
```
marceljk: same here, extend it with the same checks as above

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -331,6 +357,14 @@ func TestAccCDNDistributionHttp(t *testing.T) {
...
 					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.blocked_countries.0", "CU"),
 					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.optimizer.enabled", testutil.ConvertConfigVariable(testConfigVarsHttp["optimizer"])),
+
+					// Checking WAF Mutated Configurations
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.mode", "ENABLED"),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.type", testutil.ConvertConfigVariable(testConfigVarsHttp["waf_type"])),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.enabled_rule_ids.#", "2"),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.enabled_rule_ids.0", "@builtin/crs/request/941120"),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.enabled_rule_ids.1", "@builtin/crs/request/941130"),
+
```
marceljk: same here, extend it with the same checks as above

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -759,10 +915,46 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
...
+		configPatch.Waf.LogOnlyRuleGroupIds = getWafSet(ctx, wafModel.LogOnlyRuleGroupIds)
+		configPatch.Waf.EnabledRuleCollectionIds = getWafSet(ctx, wafModel.EnabledRuleCollectionIds)
+		configPatch.Waf.DisabledRuleCollectionIds = getWafSet(ctx, wafModel.DisabledRuleCollectionIds)
+		configPatch.Waf.LogOnlyRuleCollectionIds = getWafSet(ctx, wafModel.LogOnlyRuleCollectionIds)
+
+		if !utils.IsUndefined(wafModel.ParanoiaLevel) {
+			pl := cdnSdk.WafParanoiaLevel(wafModel.ParanoiaLevel.ValueString())
+			configPatch.Waf.ParanoiaLevel = &pl
+		}
+	}
```
marceljk, line 955: in convertConfig there are some parts duplicated. Maybe this can be summarized

stackit/internal/services/cdn/cdn_acc_test.go
matheuspolitano: Done

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: i've removed the sort

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: Yep, before we haven't validation and when add the validation i forget to get rid of it

stackit/internal/services/cdn/distribution/resource.go

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: done

stackit/internal/services/cdn/distribution/datasource.go
matheuspolitano: i move the mapWafSet out the function scope to share between datasource and resource.

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: fix

stackit/internal/services/cdn/cdn_acc_test.go
matheuspolitano: Done

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: ignored

stackit/internal/services/cdn/cdn_acc_test.go
matheuspolitano: Done

stackit/internal/services/cdn/cdn_acc_test.go
matheuspolitano: doNE

docs/data-sources/cdn_distribution.md
Fyusel: I cannot access internal documentation links. I receive a 403 (Forbidden).

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -668,18 +829,18 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
 	ctx = tflog.SetField(ctx, "project_id", projectId)
 	ctx = tflog.SetField(ctx, "distribution_id", distributionId)
 
-	configModel := distributionConfig{}
-	diags = model.Config.As(ctx, &configModel, basetypes.ObjectAsOptions{
+	configmodel := distributionConfig{}
```
Fyusel:
```suggestion
	configModel := distributionConfig{}
```

docs/data-sources/cdn_distribution.md
matheuspolitano: You are right! Sorry i've mixed up the internal-docs and docs. It's already fixed:)

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: DOne :)

stackit/internal/services/cdn/distribution/resource.go
marceljk: I meant here to remove the whole mapWafString function and use everywhere `types.StringPointerValue()` instead. It does the same thing but keeps the code clean because this is a function of the framework
```suggestion
```

stackit/internal/services/cdn/distribution/resource.go
marceljk: same here, remove this function and use instead everywhere `conversion.StringSetToSlice()`
```suggestion
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1453,3 +1734,27 @@ func validateCountryCode(country string) (string, error) {
...
+// Helper to unconditionally map set fields
+func mustMapStringSet(ctx context.Context, apiList []string) types.Set {
+	if apiList != nil {
+		setVal, diags := types.SetValueFrom(ctx, types.StringType, apiList)
+		if !diags.HasError() {
+			return setVal
+		}
+	}
+	return types.SetNull(types.StringType)
+}
```
marceljk: Move this to the conversion package and name it to something like `SliceToStringSet` and avoid context if possible and don't ignore the error.

Could be something like this:

```golang
func SliceToStringSet(list []string) (types.Set, error) {
	if list == nil {
		return types.SetNull(types.StringType), nil
	}
	set := make([]attr.Value, len(list))
	for idx, v := range list {
		stringValue := types.StringValue(v)
		set[idx] = stringValue
	}
	result, diags := types.SetValue(types.StringType, set)
	if diags.HasError() {
		return types.SetNull(types.StringType), fmt.Errorf("converting to SetValue: %v", diags.Errors())
	}
	return result, nil
}
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1117,6 +1306,57 @@ func mapFields(ctx context.Context, distribution *cdnSdk.Distribution, model *Mo
...
+
+	// Determine if WAF should be entirely excluded to prevent drift.
+	// The API can return an empty string for fully unconfigured backends.
+	isWafDisabled := (distribution.Config.Waf.Mode == cdnSdk.WAFMODE_DISABLED) &&
+		(distribution.Config.Waf.Type == cdnSdk.WAFTYPE_FREE)
+
+	var wafVal attr.Value
+	if isWafDisabled && (isImport || utils.IsUndefined(oldConfig.Waf)) {
+		wafVal = types.ObjectNull(wafTypes)
+	} else {
```
marceljk: I'm against this workaround here. The datasource shows always waf with the attributes and the resource should behave the same, even if it's not configured by the customer

stackit/internal/services/cdn/cdn_acc_test.go
```diff
@@ -190,6 +225,40 @@ func TestAccCDNDistributionHttp(t *testing.T) {
 						"DE",
 					),
 					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.optimizer.enabled", testutil.ConvertConfigVariable(testConfigVarsHttp["optimizer"])),
+					// WAF Checks
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.mode", testutil.ConvertConfigVariable(testConfigVarsHttp["waf_mode"])),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.type", testutil.ConvertConfigVariable(testConfigVarsHttp["waf_type"])),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.allowed_http_methods.#", "1"),
+					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.allowed_http_methods.0", "GET"),
```
marceljk: Don't set the values here by hand. Use them from you config.
```suggestion
					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.allowed_http_methods.#", "1"),
					resource.TestCheckResourceAttr("stackit_cdn_distribution.distribution", "config.waf.allowed_http_methods.0", testConfigVarsHttp["waf_allowed_http_methods"]),
```

Change it also for all other checks, even in the Read via datasource and update step

matheuspolitano: I've create var prevent manual hardcoding

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: i made them optional with default

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: i crated a conversion function for it

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: now im using

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: i've created in conversion to avoid check multiples time the error

marceljk (dismissed):

Fyusel:
@matheuspolitano The tests are currently failing. Afterwards I would also approve and this can be merged

Fyusel (approved):

marceljk (approved):

PR #1375, opened 2026-04-14
chore(serverbackup): switch to new SDK structure

GokceGK:
## Description

relates to STACKITTPR-574

## Checklist

- [x] Issue was linked above


stackit/internal/services/serverbackup/schedule/resource.go
```diff
@@ -466,18 +466,18 @@ func mapFields(ctx context.Context, schedule *serverbackup.BackupSchedule, model
 	if model == nil {
 		return fmt.Errorf("model input is nil")
 	}
-	if schedule.Id == nil {
+	if schedule.Id == 0 {
 		return fmt.Errorf("response id is nil")
```
cgoetz-inovex, line 470: Error message does not match check.

Also not sure if we should just remove this check. API docs sadly only specify `integer` for `schedule.Id` but not what valid values are, so I'd accept 0 here as valid. The old implementation would have accepted 0.

GokceGK: yes, you are right. I think I will remove this check.

GokceGK: I have removed this check as the value can be 0 .

cgoetz-inovex (approved):

PR #1397, opened 2026-04-20
chore(ske): switch to new SDK structure

marceljk:
## Description

relates to STACKITTPR-579

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/ske/cluster/resource.go
+++ b/stackit/internal/services/ske/cluster/resource.go
@@ -13,6 +13,7 @@ import (
 	skeUtils "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/ske/utils"
 	stringplanmodifierUtils "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils/planmodifiers/stringplanmodifier"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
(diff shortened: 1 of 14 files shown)


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1705,7 +1677,7 @@ func mapNetwork(cl *ske.Cluster, m *Model) error {
 	// If we parse that object into the terraform model, it will produce an inconsistent result after apply error
 
 	emptyNetwork := &ske.Network{}
-	if *cl.Network == *emptyNetwork && m.Network.IsNull() {
+	if cmp.Equal(cl.Network, emptyNetwork, utils.CmpIgnoreAdditionalProperties()) && m.Network.IsNull() {
```
rubenhoenle: From the [docs](https://pkg.go.dev/github.com/google/go-cmp/cmp) of the go-cmp project:

> It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values.

So I don't think this is the way to go. Reflection also doesn't seem appropriate.

So after thinking about it I see two possible ways:

- We can implement some `isEmptyNetwork()` util function for this check.
- We can implement some Network equals function two compare two networks.

Actually I prefer the `isEmptyNetwork` implementation as it allows pretty unit testing for our use-case.

rubenhoenle: Additional note after expanding the diff: After all, this situation is the result of some workaround in the Terraform provider codebase for some weird API design:

https://github.com/stackitcloud/terraform-provider-stackit/blob/0d7331bbb0bc764b372a5d44ffd2b05fb0c56550/stackit/internal/services/ske/cluster/resource.go#L1704-L1705

So this whole thing is a very niche use-case and therefore I even more prefer the `isEmptyNetwork` implementation.

So maybe worth adding a ticket to our techdebt collector for this one 😅

marceljk: replaced the `cmp.Equal` with a `IsEmptyNetwork()` and `IsEmptyExtension()` function in https://github.com/stackitcloud/terraform-provider-stackit/pull/1397/commits/57a6b768efc28df466dc8e6105ccee19f11da315

GokceGK (approved):

PR #1408, opened 2026-04-21
fix(server): update network example to use "ipv4_nameservers" instead of "nameservers"

DimitrijKimel:
## Description

The stackit_servers network example uses argument "nameservers", but stackit_network requires "ipv4_nameservers" instead

image

https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/resources/network
image

### DIFF
--- a/docs/resources/server.md
+++ b/docs/resources/server.md
@@ -73,7 +73,7 @@ description: |-
   resource "stackit_network" "network" {
     project_id         = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
     name               = "example-network"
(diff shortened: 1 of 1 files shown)


marceljk:
Hi @DimitrijKimel,

nice catch and thanks for your PR!

Our docs are generated and our pipeline checks also if it's up to date. Can you rename it here https://github.com/stackitcloud/terraform-provider-stackit/blob/acd75e450194f1b35ab3ae731f03c1e125cbcac9/stackit/internal/services/iaas/server/const.go#L75

and run `$ make generate-docs`, in order to fix it?
Thanks

DimitrijKimel:
Thank you for pointing that out. I will do that asap

PR #1444, opened 2026-05-06
feat(sfs): add snapshot_policy_id to sfs_resource_pool

marceljk:
## Description

relates to STACKITTPR-629

## Checklist

- [ ] Issue was linked above


stackit/internal/services/sfs/resourcepool/resource.go
```diff
@@ -200,6 +220,39 @@ func (r *resourcePoolResource) Schema(_ context.Context, _ resource.SchemaReques
 				Computed:    true,
 				Default:     booldefault.StaticBool(false),
 			},
+			"snapshot_policy": schema.SingleNestedAttribute{
+				Description: `Name of the snapshot policy.`,
+				Computed:    true,
+				Optional:    true,
+				Default: objectdefault.StaticValue(
+					types.ObjectValueMust(snapshotPolicyTypes, map[string]attr.Value{
+						"id":   types.StringValue(defaultSnapshotPolicyId),
```
rubenhoenle: I would expect the API to return an error for an empty string value as a snapshot policy ID.

At least I didn't get any snapshot policy with id "" from https://docs.api.stackit.cloud/documentation/sfs/version/v1#operation/ListSnapshotPolicies

rubenhoenle: API design is the one thing. But why to we send an empty string? I don't quite get it tbh 😅

marceljk: The snapshot policy ID is in the PATCH request optional, but not nullable. So to remove the snapshot policy from an existing resource pool, we need to set it to an empty string. The empty string, is also the default "snapshot_policy_id", when no id was set within the creation.

I set it on default to an empty string, to automatically remove the snapshot_policy_id, when someone removes the object in their terraform config. Otherwise terraform wouldn't trigger an update

rubenhoenle: So the empty string thing is a workaround for an API issue?

marceljk: removed the workaround and also the API issue is fixed

rubenhoenle (approved):

PR #1509, opened 2026-06-05
chore: Add a note about region to S3 guide

JorTurFer:
## Description

Working with multiple regions within the same terraform execution, I found that the underlaying AWS client caches the connection by the region, and mixing region (even using multiple providers) results in a calls mixing the region credentials when the resource doesn't set region.

## Checklist

### DIFF
--- a/templates/guides/aws_provider_s3_stackit.md.tmpl
+++ b/templates/guides/aws_provider_s3_stackit.md.tmpl
@@ -7,6 +7,9 @@ page_title: "Using AWS Provider for STACKIT Object Storage (S3 compatible)"
 
 This guide outlines the process of utilizing the [AWS Terraform Provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs) alongside the STACKIT provider to create and manage STACKIT Object Storage (S3 compatible) resources.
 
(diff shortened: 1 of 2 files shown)


templates/guides/aws_provider_s3_stackit.md.tmpl
```diff
@@ -7,6 +7,9 @@ page_title: "Using AWS Provider for STACKIT Object Storage (S3 compatible)"
 
 This guide outlines the process of utilizing the [AWS Terraform Provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs) alongside the STACKIT provider to create and manage STACKIT Object Storage (S3 compatible) resources.
 
+>Note: AWS Provider uses the region as cache key to manage the connection efficiently. Therefore, the region must be set to the desired one in all the resources explicitly to ensure proper functionality when multiple regions are used.
```
GokceGK:
```suggestion
> Note: AWS Provider uses the `region` as cache key to manage the connection efficiently. Therefore, the `region` must be set to the desired one in all the resources explicitly to ensure proper functionality when multiple regions are used.
```

Manuelvaas (approved):

GokceGK (approved):

PR #1522, opened 2026-06-16
feat(ci): re-enable linting rule to prevent usage of deprecated code

rubenhoenle:
## Description

can be re-enabled now since the migration to the new SDK layer is done

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/conversion/conversion.go
+++ b/stackit/internal/conversion/conversion.go
@@ -161,38 +161,27 @@ func StringListToPointer(list basetypes.ListValue) (*[]string, error) {
 // StringListToSlice converts basetypes.ListValue to a list of strings.
 // It returns nil if the value is null or unknown.
 func StringListToSlice(list basetypes.ListValue) ([]string, error) {
+	return StringListToEnumSlice[string](list)
+}
+
+// StringListToEnumSlice converts basetypes.ListValue to a list of enums.
+// It returns nil if the value is null or unknown.
+func StringListToEnumSlice[T ~string](list basetypes.ListValue) ([]T, error) {
 	if list.IsNull() || list.IsUnknown() {
 		return nil, nil
 	}
 
 	// Instantiate an empty slice to ensure the slice is not nil
-	listStr := []string{}
+	listStr := []T{}
 	for i, el := range list.Elements() {
 		elStr, ok := el.(types.String)
 		if !ok {
 			return nil, fmt.Errorf("element %d is not a string", i)
 		}
-		listStr = append(listStr, elStr.ValueString())
+		listStr = append(listStr, T(elStr.ValueString()))
 	}
 
 	return listStr, nil
-}
-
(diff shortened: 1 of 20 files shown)


stackit/internal/conversion/conversion.go
```diff
@@ -170,6 +178,23 @@ func StringListToSlice(list basetypes.ListValue) ([]string, error) {
...
+	}
+
+	if s == nil {
+		return nil, nil
+	}
+
+	return utils.Map(s, func(s string) T {
+		return T(s)
+	}), nil
+}
```
marceljk: Small optimization suggestion: I think it would be nice if we can avoid to iterate twice through the same list. At the moment we iterate once it in the call of `StringListToSlice` and a second time in the `utils.Map`.

We could change it, so that all the logic is in `StringListToEnumSlice` and it takes care of the correct type. And in the `StringListToSlice` function we can just call this function, with string as type.

Something like this:
```go
// StringListToSlice converts basetypes.ListValue to a list of strings.
// It returns nil if the value is null or unknown.
func StringListToSlice(list basetypes.ListValue) ([]string, error) {
	return StringListToEnumSlice[string](list)
}

// StringListToSlice converts basetypes.ListValue to a list of enums.
// It returns nil if the value is null or unknown.
func StringListToEnumSlice[T ~string](list basetypes.ListValue) ([]T, error) {
	if list.IsNull() || list.IsUnknown() {
		return nil, nil
	}

	// Instantiate an empty slice to ensure the slice is not nil
	listStr := []T{}
	for i, el := range list.Elements() {
		elStr, ok := el.(types.String)
		if !ok {
			return nil, fmt.Errorf("element %d is not a string", i)
		}
		listStr = append(listStr, T(elStr.ValueString()))
	}

	return listStr, nil
}
```

stackit/internal/conversion/conversion_test.go
```diff
@@ -635,3 +636,155 @@ func TestTerraformStringSetToList(t *testing.T) {
...
+	type args struct {
+		list basetypes.ListValue
+	}
+	type testCase[T interface{ ~string }] struct {
+		name    string
+		args    args
+		want    []T
+		wantErr bool
+	}
+	tests := []testCase[opensearch.InstanceParametersTlsProtocolsInner]{
```
marceljk: The attribute which uses this enum was recently deprecated, so I would suggest to use a different enum. 
https://github.com/stackitcloud/stackit-sdk-go/blob/e26a28333b0732e46ac8d807d5b9a732b36b20e1/services/opensearch/v1api/model_instance_parameters.go#L46-L47

https://github.com/stackitcloud/stackit-sdk-go/blob/e26a28333b0732e46ac8d807d5b9a732b36b20e1/services/opensearch/CHANGELOG.md?plain=1#L8

marceljk (approved):

PR #1532, opened 2026-06-22
Support Azure DevOps service connection

### LINKED ISSUE #1321: Struggling with setting up Workload Identity Federation for my Azure DevOps Pipeline
Hi, as the title says, I'm currently trying to setup Workload Identity Federation (WIF) in my Azure DevOps Pipeline in order to move away from long lived service account keys that I currently supply with the `STACKIT_SERVICE_ACCOUNT_KEY ` environment variable.

I followed the guide [here](https://registry.terraform.io/providers/stackitcloud/stackit/latest/docs/guides/workload_identity_federation#azure-devops-assertions) and opted for the method with the `STACKIT_SERVICE_ACCOUNT_EMAIL` and `STACKIT_USE_OIDC` environment variables but after the setup, I'm getting this error:

```
╷
│ Error: Error configuring provider
│ 
│   with provider["registry.opentofu.org/stackitcloud/stackit"],
│   on providers.tf line 1, in provider "stackit":
│    1: provider "stackit" {
│ 
│ Setting up authentication: configuring no auth client: error initializing
│ client: error reading federated token file - open
│ /var/run/secrets/stackit.cloud/serviceaccount/token: no such file or
│ directory
╵
```
Side note: The error messages are kind of cryptic. For example, in my first pipeline run, I forgot to set the mail properly and got this error.
```
╷
│ Error: Error configuring provider
│ 
│   with provider["registry.opentofu.org/stackitcloud/stackit"],
│   on providers.tf line 1, in provider "stackit":
│    1: provider "stackit" {
│ 
│ Setting up authentication: configuring no auth client: error initializing
│ client: client ID cannot be empty
╵

```
Side note no.2: Something I noticed is that the provider has no authentication fall-through mechanic when supplying `STACKIT_USE_OIDC`. If this fails, there is no attempt to check the service account key methods.

---

Back to the first error. I'm not sure if I correctly set up Federated Identity Providers in the stackit portal. When adding a new provider, the form has one entry by default:
aud → equals → sts.accounts.stackit.cloud 

It was not clear to me whether I should keep it  or not and just added the other entry:
aud->equals->api://AzureADTokenExchange

Resulting in having both entries. For testing, I removed the stackit one but that resulted in the same error.

The second part that I might have misconfigured is the `sub` entry. In the docs it says:
```
Using a hypothetical pipeline named terraform-ado-oidc inside the project '[https://myorg.azure.com/project-abc`](https://myorg.azure.com/project-abc%60) as example ....
....
sub->equals->p://myorg/project-abc/terraform-ado-oidc # This is the pipeline where the process is running
```
I assume, `myorg` is the part after https://dev.azure.com/ and before the next '/' from `$(System.CollectionUri)` or `$(System.TeamFoundationCollectionUri)`.
I assume, `project-abc` is `$(System.TeamProject)`. In practice, when using a normal git repo url as example `https://dev.azure.com/<myorg>/<project-abc>/_git/<repo-name>`
In theory, there is also `$(System.TeamProjectId)`.

But for the pipeline name, `terraform-ado-oidc` in the example, this is not obvious. I can have two pipelines with the same name under different folders in Azure DevOps. They are under different paths but have the same name.

Image

Both of the return `cluster`, when printing `$(Build.DefinitionName)` in them.

That's why I think it is imprecise to tell the users to use the pipeline name. I can differentiate them via their ID `$(System.DefinitionId)` when printing it in the pipeline or in the url (definitionId=012345), when on the pipeline runs page. Further details can be queried under `https://dev.azure.com/<orgname>/<projectname>/_apis/pipelines/<DefinitionId>?api-version=7.1`. In that response, there is also the `folder` attribute. In my case above, `\\k8s` and for the other one `\\test`.

I tried changing the pipeline name in the `sub` assertion to the `$(System.DefinitionId)` value but I still get the same error.

I also tried:
```
sub->equals->p://<myorg>/<myproject>/\\test\cluster 
```

but that also didn't work.

Can you help me?

JorTurFer:
## Description

During next days, I'll publish a new [Azure DevOps task](https://marketplace.visualstudio.com/items?itemName=SchwarzDigits.stackit-service-connection) and service connection to support STACKIT Service Accounts nativelly within [Azure DevOps](https://github.com/stackitcloud/azure-devops-stackit-service-connection-extension). This PR adds support for native a service connection.

As soon as the new task is public, I'll update the PR adding more info to the guide, but in terms of code, this is all the needed changes

Fixes https://github.com/stackitcloud/terraform-provider-stackit/issues/1321

## Checklist

- [ ] Issue was linked above


JorTurFer:
not stale

JorTurFer:
Azure DevOps task is live 🚀  
https://github.com/stackitcloud/azure-devops-stackit-service-connection-extension
https://marketplace.visualstudio.com/items?itemName=SchwarzDigits.stackit-service-connection

We can continue with this PR when you wish :)

JorTurFer:
not stale

JorTurFer:
not stale

marceljk:
Code looks good to me. Before approving and merging it, I would like to test it. I had to request your new task for our organization. As soon this is approved and ready to use, I'll continue with testing this PR.

JorTurFer:
> Code looks good to me. Before approving and merging it, I would like to test it. I had to request your new task for our organization. As soon this is approved and ready to use, I'll continue with testing this PR.

I can't intall it in the org, but once it's installed, I can help with the config. Just ping me if you need help

JorTurFer:
not stale

docs/guides/workload_identity_federation.md
```diff
@@ -50,8 +49,7 @@ execute terraform on the main branch, we will configure the service account "Fed
     - Via Pipeline: Add a script step echo $(System.CollectionId) to print it during a run.
 - **Assertions**:
   - **aud**->equals->api://AzureADTokenExchange # Mandatory value
-  - **sub**->equals->p://myorg/project-abc/terraform-ado-oidc # This is the pipeline where the process is running
-  - **rpo_ref**->equals->refs/heads/main # This is the branch where the pipeline will run
+  - **sub**->equals->sc://myorg/project-abc/stackit-service-connection # This is the service connection that is being used to authenticate with STACKIT.
```
marceljk: I think it would be nice if we document both options. So the assertions, which can be set without your service connection extension and these which can be set your service connection extension.

As far as I saw, the assertion fails, when rpo_ref is set and the service connection extension. Is this a known limitation? Not sure if it makes sense, because the list with assertions is based on this issue [#1321](https://www.github.com/stackitcloud/terraform-provider-stackit/issues/1321) not complete, but maybe we could add hint that rpo_ref doesn't work with service connection.

Beside that comments, I think it would be helpful, to define the example `sub` for service connection a bit different. When I tried to configure it on my own, I misconfigured it the whole time because I set a wrong value for the last part `stackit-service-connection`. Here it would be maybe helpful, to choose a name like `my-stackit-service-connection` which makes it more clear, that the configured service connection name must be set. But maybe it's also only an issue I had, because I'm not that familiar with ADO and WIF / service connection.

docs/guides/workload_identity_federation.md
```diff
@@ -160,15 +161,18 @@ pool:
...
 
 jobs:
   - job: demo-job
     displayName: "Workload Identity Federation with STACKIT"
     steps:
       - checkout: self
 
+      - task: StackitAuthenticate@1
+        inputs:
+          serviceConnection: "My STACKIT Connection (WIF)"
```
marceljk: I would also suggest, to rename here the serviceConnection, to the same value like in the assertion above, e.g. `my-stackit-service-connection`. This makes it probably easier for first time users with service connection and WIF. We could add here a comment like `# Name of your configured project service connection`, because I had here also some issues in the beginning, that I first didn't configure the service connection in my project and was confused, why I can't configure here a valid value. But as said, I was using service connection for the first time and maybe that's only an issue I had

marceljk (commented):
I finally could try your extensions and tried to configure it based on the docs. Left a few comments, where I had some configuration issues

docs/guides/workload_identity_federation.md
JorTurFer: The token is different when you request it for a pipeline or a service connection. We can keep both options but I think that documenting both will be more complex and the recommended way with ADO is via service connection, it's the standard.

about 'stackit-service-connection', what do you think if I rename it to 'YourStackitServiceConnectionName'? Here and for the other comment

JorTurFer: `sc://MyOrg/Project-ABC/MyStackitServiceConnectionName`

marceljk: Yes, both names are fine for me.

> `sc://MyOrg/Project-ABC/MyStackitServiceConnectionName`

I think this one is even better.

marceljk: If the service connection is the way to go and is recommended, then I'm also fine with only documenting this

JorTurFer: done!

docs/guides/workload_identity_federation.md
JorTurFer: done!

marceljk (dismissed):
Thanks for your contribution! Looks good to me and already posted it in our internal channel for a second review

JorTurFer:
not stale

stackit/provider.go
```diff
@@ -634,7 +640,7 @@ func (p *Provider) Configure(ctx context.Context, req provider.ConfigureRequest,
 		oidcReqToken = utils.GetEnvStringOrDefault(providerConfig.OIDCTokenRequestToken, "SYSTEM_ACCESSTOKEN", "")
 		// This can be set to the ID of the service connection to restrict the token exchange to that connection, not supported by default to avoid additional configuration
 		// for users that don't need it, can be added as an additional provider config parameter in the future if there is demand
```
Fyusel: This comment needs an update

JorTurFer: nice catch! solved :)

Fyusel (approved):

marceljk (approved):

PR #1543, opened 2026-06-26
fix(iaas): IaaS acceptance tests fixes

Fyusel:
Create a project to network interface and server acceptance tests. ipv4_nameservers from List to Set. Set is an unordered list and detects changes based on a hash. Update image_v2 acceptance test.

## Description

STACKITTPR-673

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/iaas/iaas_acc_test.go
+++ b/stackit/internal/services/iaas/iaas_acc_test.go
@@ -10,11 +10,12 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/stackitcloud/stackit-sdk-go/services/iaas/v2api/wait"
+	resourcemanager "github.com/stackitcloud/stackit-sdk-go/services/resourcemanager/v0api"
+	resourcemanager_wait "github.com/stackitcloud/stackit-sdk-go/services/resourcemanager/v0api/wait"
 
 	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
@@ -118,11 +119,13 @@ const (
 // SERVER - MIN
 
 var testConfigServerVarsMin = config.Variables{
-	"project_id":   config.StringVariable(testutil.ProjectId),
-	"name":         config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
-	"network_name": config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
-	"machine_type": config.StringVariable("t1.1"),
-	"image_id":     config.StringVariable("fb5b3fa8-5e20-478a-929a-2b7da1676b18"),
+	"server_name":         config.StringVariable(fmt.Sprintf("tf-acc-server-min%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"network_name":        config.StringVariable(fmt.Sprintf("tf-acc-server-min-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"machine_type":        config.StringVariable("t1.1"),
+	"image_id":            config.StringVariable("fb5b3fa8-5e20-478a-929a-2b7da1676b18"),
+	"name":                config.StringVariable(fmt.Sprintf("tfe2e-project-server-min-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"owner_email":         config.StringVariable(testutil.TestProjectServiceAccountEmail),
+	"parent_container_id": config.StringVariable(testutil.TestProjectParentContainerID),
 }
 
 var testConfigServerVarsMinUpdated = func() config.Variables {
@@ -136,8 +139,9 @@ var testConfigServerVarsMinUpdated = func() config.Variables {
 // SERVER - MAX
 
 var testConfigServerVarsMax = config.Variables{
-	"project_id":           config.StringVariable(testutil.ProjectId),
-	"name":                 config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"network_name":         config.StringVariable(fmt.Sprintf("tf-acc-server-max-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"server_name":          config.StringVariable(fmt.Sprintf("tf-acc-server-max-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"project_name":         config.StringVariable(fmt.Sprintf("tfe2e-project-server-max-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
 	"name_not_updated":     config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
 	"machine_type":         config.StringVariable("t1.1"),
 	"image_id":             config.StringVariable("fb5b3fa8-5e20-478a-929a-2b7da1676b18"),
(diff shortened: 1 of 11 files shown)


stackit/internal/utils/utils_test.go
```diff
@@ -133,6 +133,57 @@ func TestListValuetoStrSlice(t *testing.T) {
...
+			description: "values ok",
+			input: types.SetValueMust(types.StringType, []attr.Value{
+				types.StringValue("a"),
+				types.StringValue("b"),
+				types.StringValue("c"),
+			}),
+			expected: []string{"a", "b", "c"},
+			isValid:  true,
+		},
+		{
```
marceljk: Can you add here a test case for a null set? Want to have it here, to ensure the implementation doesn't change. Because if during an update the set was removed from the config, we usually want to send an explicit empty slice to the API. Otherwise the API wouldn't recognize during the patch call, that the value should be removed and Terraform results in a state drift.
```suggestion
		},
		{
			description: "null set",
			input:       types.SetNull(types.StringType),
			expected:    []string{},
			isValid:     true,
		},
		{
```

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -118,11 +120,13 @@ const (
...
-	"network_name": config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
-	"machine_type": config.StringVariable("t1.1"),
-	"image_id":     config.StringVariable("fb5b3fa8-5e20-478a-929a-2b7da1676b18"),
+	"server_name":         config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"network_name":        config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"machine_type":        config.StringVariable("t1.1"),
+	"image_id":            config.StringVariable("fb5b3fa8-5e20-478a-929a-2b7da1676b18"),
+	"name":                config.StringVariable(fmt.Sprintf("tfe2e-project-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"owner_email":         config.StringVariable(testutil.TestProjectServiceAccountEmail),
+	"parent_container_id": config.StringVariable(testutil.TestProjectParentContainerID),
```
marceljk, line 128: Can your extend our table, that the parent container id is now also for IaaS required?
https://github.com/stackitcloud/terraform-provider-stackit/blob/4e42f8cbdfa26fee29aaf9d116f66e682c1d7b33/README.md?plain=1#L235

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -118,11 +120,13 @@ const (
...
-	"project_id":   config.StringVariable(testutil.ProjectId),
-	"name":         config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
-	"network_name": config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
-	"machine_type": config.StringVariable("t1.1"),
-	"image_id":     config.StringVariable("fb5b3fa8-5e20-478a-929a-2b7da1676b18"),
+	"server_name":         config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"network_name":        config.StringVariable(fmt.Sprintf("tf-acc-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
+	"machine_type":        config.StringVariable("t1.1"),
+	"image_id":            config.StringVariable("fb5b3fa8-5e20-478a-929a-2b7da1676b18"),
+	"name":                config.StringVariable(fmt.Sprintf("tfe2e-project-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
```
marceljk: Personal preference: I think it would be nice, if the project name contains the information, to which test is belongs. Maybe something like `server-min`. This makes it easier to find the related test in the portal, in case the test stops without cleaning up.

```suggestion
	"name":                config.StringVariable(fmt.Sprintf("tfe2e-project-server-min-%s", acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))),
```

But we should to take care, that the name doesn't reach the limit of 39 chars

marceljk: Or as alternative, we print the project name within the acc test, like we have it with the network name.

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -2199,8 +2210,8 @@ func TestAccServerMin(t *testing.T) {
 				Config:          fmt.Sprintf("%s\n%s", testutil.NewConfigBuilder().BuildProviderConfig(), resourceServerMinConfig),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					// Server
-					resource.TestCheckResourceAttr("stackit_server.server", "project_id", testutil.ConvertConfigVariable(testConfigServerVarsMin["project_id"])),
-					resource.TestCheckResourceAttr("stackit_server.server", "name", testutil.ConvertConfigVariable(testConfigServerVarsMin["name"])),
+					resource.TestCheckResourceAttrSet("stackit_server.server", "project_id"),
```
marceljk: I think it would be better, to don't only check if the project_id is set, but also if it's set to the correct value.
```suggestion
					resource.TestCheckResourceAttrPair(
						"stackit_resourcemanager_project.example", "project_id",
						"stackit_server.server", "project_id",
					),
```

marceljk: same applies to the other occurrences

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -5453,6 +5544,7 @@ func testAccCheckDestroy(s *terraform.State) error {
 		testAccCheckNetworkAreaDestroy,
 		testAccCheckRoutingTableDestroy,
 		testAccCheckRoutingTableRouteDestroy,
+		testAccCheckResourceManagerProjectsDestroy,
 	}
 	var errs []error
```
marceljk, line 5759: I think we should change two small things in this testAccCheckDestroy function to prevent errors:

1. I would suggest to remove the waitgroup stuff here and don't call the checkFunctions in goroutines. Most of the resources must be deleted before the next one can be removed (e.g. network interface first, before deleting network). With the goroutine it all happens simultaneously, which is more likely to fail.

2. We should change the order of the checkFunctions, to take care, that everything will be cleaned up in the correct order. For example first server, then volumes, network interfaces, networks, (then probably everything else) and in the end resource manager, routingTableRoutes, routingTable, networkAreaRegion, networkArea. Not sure if I missed something, but I think this should be almost the correct order.

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -5697,28 +5775,68 @@ func testAccCheckServerDestroy(s *terraform.State) error {
...
+	}
 	return nil
 }
 
+func isIgnorableDestroyError(err error) bool {
+	var oapiErr *oapierror.GenericOpenAPIError
+	if errors.As(err, &oapiErr) {
+		return oapiErr.StatusCode == http.StatusForbidden ||
+			oapiErr.StatusCode == http.StatusNotFound ||
+			oapiErr.StatusCode == http.StatusBadRequest
```
marceljk: Should a BadRequest really be ignored? Usually this means that the sent payload was invalid

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -5966,3 +6084,57 @@ func testAccCheckRoutingTableRouteDestroy(s *terraform.State) error {
...
+		if err != nil {
+			return fmt.Errorf("destroying project %s during CheckDestroy: %w", items[i].ContainerId, err)
+		}
+
+		_, err = resourcemanager_wait.DeleteProjectWaitHandler(ctx, client.DefaultAPI, items[i].ContainerId).WaitWithContext(ctx)
+		if err != nil {
+			return fmt.Errorf("destroying project %s during CheckDestroy: waiting for deletion %w", items[i].ContainerId, err)
+		}
+	}
+	return nil
```
marceljk: I would implement the testDestroy function a bit different. I know that the current implementation was done for a lot of services and also here in IaaS, but I think we can change a few things to make this more stable.

In generel, I think we shouldn't read here any IDs or informations from the testutil vars. Because it could be the case, that we the project got a different parent ID, e.g. from a folder which is part of the test. So we should read everything out of the state.

I also would remove the list endpoint, because theoretically every project could have a different parent ID and then this doesn't work anymore. So instead I would just do a GET request, check the status code (and maybe the state of the resource) and then call the Delete endpoint. 

I already wrote this for an upcoming resource in our internal repo. I'll send you a link to it in our internal chat.

stackit/internal/utils/utils_test.go
Fyusel: Removed since the Set is not used anymore.

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -5440,35 +5717,29 @@ func TestAccRoutingTableRouteMax(t *testing.T) {
...
 		testAccCheckIaaSPublicIpDestroy,
 		testAccCheckIaaSKeyPairDestroy,
 		testAccCheckIaaSImageDestroy,
-		testAccCheckNetworkDestroy,
 		testAccCheckNetworkInterfaceDestroy,
+		testAccCheckNetworkDestroy,
 		testAccCheckNetworkAreaRegionDestroy,
 		testAccCheckNetworkAreaDestroy,
 		testAccCheckRoutingTableDestroy,
 		testAccCheckRoutingTableRouteDestroy,
```
marceljk: these two should be swapped in the order, because the route is a subresource of the routing table

stackit/internal/services/iaas/iaas_acc_test.go
```diff
@@ -5966,3 +6262,84 @@ func testAccCheckRoutingTableRouteDestroy(s *terraform.State) error {
...
+			}
+		}
+
+		// check if the nameservers are present
+		for _, expected := range expectedIPs {
+			if !importedIPs[expected] {
+				return fmt.Errorf("bug in import: expected nameserver %s was lost. Found: %v", expected, importedIPs)
+			}
+		}
+
```
marceljk, line 6371: I would add here also a length compare check. So if the importedIPs have the same amount of IPs like importedIPs. Because theoretically, there can be more importedIPs without recognizing it

marceljk (dismissed):

marceljk (approved):

PR #1575, opened 2026-07-09
feat: integrate vpc in network

cah-hbaum:
Integrate the vpc data into the network resource.

## Description

relates to #1234

## Checklist

- [ ] Issue was linked above


stackit/internal/services/iaas/network/resource.go
```diff
@@ -404,6 +405,18 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 					boolplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"vpc_id": schema.StringAttribute{
+				Description: "The ID of the VPC the network is associated with.",
+				Optional:    true,
+				Computed:    true,
```
rubenhoenle: Why is this set to computed? 😅

rubenhoenle (changes requested):
Rejecting because

- implementation of acc tests is missing
- only the schema is added, but you also have to send the values to the API and write the API response back to the Terraform state

stackit/internal/services/iaas/network/resource.go
cah-hbaum: Removed the `Computed` part (youre right that its not required here)

cgoetz-inovex:
superseded by https://github.com/stackitcloud/terraform-provider-stackit/pull/1616

PR #1576, opened 2026-07-09
feat(dremio): stop returning oauth client secret on get/list dremio(s)

r-savu:
## Description

<!-- **Please link some issue here describing what you are trying to achieve.**
Previously the get dremio and list dremio endpoints were returning the full IDP config, including the client secret.

The client secert should be known by the IDP and the instance itself and not be returned by "routine" get/list requests.
The change relates to the internal ticket STACKITDER-894, introducing a breaking change in the types generated in the SDK. 
This PR addresses the change in the SDK.

## Checklist

- [x] Issue was linked above


stackit/internal/services/dremio/instance/resource.go
```diff
@@ -717,7 +717,7 @@ func mapAuthentication(instanceResp *dremioSdk.DremioResponse, model *Model) err
 		authModel.AzureAD = &AzureADModel{
 			AuthorityUrl: types.StringValue(azureADResp.AuthorityUrl),
 			ClientId:     types.StringValue(azureADResp.ClientId),
-			ClientSecret: types.StringValue(azureADResp.ClientSecret),
+			ClientSecret: types.StringNull(),
```
Fyusel: The acc tests are failing right now and also a manual test is failing.
This will not work since the client secret is required in the schema and is set after a creation to null which results in a state drift.

Apply fails with:
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to stackit_dremio_instance.example, provider
│ "provider[\"registry.terraform.io/stackitcloud/stackit\"]" produced an unexpected new value: .authentication:
│ inconsistent values for sensitive attribute.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

stackit/internal/services/dremio/instance/resource.go
```diff
@@ -708,33 +708,39 @@ func mapAuthentication(instanceResp *dremioSdk.DremioResponse, model *Model) err
...
+		if model.Authentication.AzureAD == nil {
+			model.Authentication.AzureAD = new(AzureADModel)
 		}
+
+		azureADResp := instanceResp.Authentication.Azuread
+		azureADModel := model.Authentication.AzureAD
+
+		azureADModel.AuthorityUrl = types.StringValue(azureADResp.AuthorityUrl)
+		azureADModel.ClientId = types.StringValue(azureADResp.ClientId)
+		azureADModel.RedirectUrl = types.StringValue(*azureADResp.RedirectUrl)
```
Fyusel: Potential nil pointer exception
->use types.StringPointerValue()

Fyusel (commented):
+ linter issues :-)

stackit/internal/services/dremio/instance/resource.go
```diff
@@ -703,37 +707,34 @@ func mapModelFields(instanceResp *dremioSdk.DremioResponse, model *Model, region
...
-func mapAuthentication(instanceResp *dremioSdk.DremioResponse, model *Model) error {
-	if model == nil {
-		return fmt.Errorf("model input is nil")
-	}
-
-	authModel := AuthenticationModel{
-		Type: types.StringValue(string(instanceResp.Authentication.Type)),
-	}
+func mapAuthentication(instanceResp *dremioSdk.DremioResponse, auth *AuthenticationModel) error {
+	auth.Type = types.StringValue(string(instanceResp.Authentication.Type))
```
Fyusel, line 718: nil checks of instanceResp and auth are missing

stackit/internal/services/dremio/instance/datasource.go
```diff
@@ -64,6 +99,74 @@ func (d *instanceDataSource) Configure(ctx context.Context, req datasource.Confi
...
+	err = mapDatasourceAuthentication(instanceResp, model.Authentication)
+	if err != nil {
+		return fmt.Errorf("failed to map Authentication fields")
+	}
+
+	return nil
+}
+
+func mapDatasourceAuthentication(instanceResp *dremioSdk.DremioResponse, auth *AuthenticationDatasourceModel) error {
+	auth.Type = types.StringValue(string(instanceResp.Authentication.Type))
```
Fyusel, line 134: nil checks of instanceResp and auth are missing

Fyusel:
Please run: make generate-docs

stackit/internal/services/dremio/instance/resource.go
r-savu: the nil check and pointer instantiation is done right before the function call, inside `mapFields`. 

My reasoning is that the presence/absence of an auth config block is a property of the parent struct, and the scope of `mapAuthentication` is limited to only modify the contents of the given pointer, without access to fields from the parent struct.

stackit/internal/services/dremio/instance/datasource.go
r-savu: same as above

stackit/internal/services/dremio/instance/resource.go
Fyusel: I see your point that this is currently only called in the `mapFields()` function and those values are checked there already but if this function is used in the future elsewhere / refactoring (or in unit tests) someone might forgot to add the nil checks and then it will crash which needs to prevented.

I prefer the defensive style here to prevent issues in the future.

stackit/internal/services/dremio/instance/datasource.go
Fyusel: same as above

Fyusel (approved):

rubenhoenle (approved):

PR #1578, opened 2026-07-10
feat(ske): support audit log configuration

tobias-pfaffelmoser-ske:
## Description

Adds support for configuring SKE cluster audit log forwarding via a new optional
`audit` block on the `stackit_ske_cluster` resource (and as a read-only attribute
on the data source):

```hcl
resource "stackit_ske_cluster" "example" {
  # ...
  audit = {
    enabled = true
  }
}
```

The audit.enabled flag is passed through to the SKE API (ske.Audit) on create/update and mapped back into state on read. This feature is in private preview and can only be enabled for accounts/projects that have been enabled for audit log forwarding to a Telemetry Router.

This also bumps the stackit-sdk-go/services/ske dependency to v1.19.0, which introduces the audit field.

## Checklist

- [ ] Issue was linked above


stackit/internal/services/ske/cluster/resource.go
```diff
@@ -412,6 +423,8 @@ var descriptions = map[string]string{
 	"access_idp":          "Configure IDP",
 	"access_idp_enabled":  "Enable IDP integration for the cluster.",
 	"access_idp_type":     "The IDP type. Possible values: 'stackit'.",
+	"audit":               "Cluster audit log forwarding configuration. This feature is in private preview. Enabling audit logging is only possible for enabled accounts.",
```
relusc:
```suggestion
	"audit":               "Cluster audit log forwarding configuration.",
```

This can be shortened as it is not part of any public documentation, release notes or similar

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1943,6 +1995,47 @@ func getMaintenanceTimes(ctx context.Context, cl *ske.Cluster, m *Model) (startT
...
+		}
+
+		auditModel := audit{}
+		diags := m.Audit.As(ctx, &auditModel, basetypes.ObjectAsOptions{})
+		if diags.HasError() {
+			return fmt.Errorf("converting audit object: %v", diags.Errors())
+		}
+
+		// The user explicitly configured `audit = { enabled = false }`. The API does
+		// not echo a disabled audit config back, so we keep the configured value to
```
MichaelEischer: If nothing is configured for a cluster, then the API always echoes `audit = {enabled = false}` in regions where the feature is available.

stackit/internal/services/ske/cluster/resource.go
tobias-pfaffelmoser-ske: Thx! Adapted.

stackit/internal/services/ske/cluster/resource.go
MichaelEischer: I'm not sure what this means for the code exactly. But we can probably drop the if below. For regions without the feature, we actually want the `Otherwise` case below as users there shouldn't enable the feature in terraform anyways.

tobias-pfaffelmoser-ske: Makes sense. Simplified.

stackit/internal/services/ske/cluster/resource.go
```diff
@@ -1943,6 +1995,25 @@ func getMaintenanceTimes(ctx context.Context, cl *ske.Cluster, m *Model) (startT
 	return startTime, endTime, nil
 }
 
+func mapAudit(cl *ske.Cluster, m *Model) error {
+	// A missing audit block only occurs in regions where the feature is
```
rubenhoenle, line 2047: I don't fully get this, can't the API just include `false` in the response in that case?

```json
{
    ...
    "audit": {
        "enabled": false
    }
}
```

Else you might have users which will run into state drifts: When Terraform users set `audit.enabled` to `false` in their Terraform configuration, Terraform will also send `false` to the API. From my current understanding the API then just won't return the `audit` object at all in the response. This will lead to Terraform setting `audit.enabled` to `null` here which will then result in a state drift error.

tobias-pfaffelmoser-ske: Good point. 

Just checked with the team: when the **feature is available** in the respective region, the API **always** echoes the audit object back, including `{ "enabled": false }`. So setting `audit.enabled = false` returns `enabled: false` and maps cleanly back into state  - so no drift expected in this scenario.

When the **feature is not available** in the respective region, the API omits the `audit` object entirely. That is the situation where `cl.Audit == nil`, and in that case, audit shouldn't be configured on the cluster in the first place. Mapping it to null might make sense: if a user tries to configure `audit` in a region without the feature, the resulting diff could be the desired signal that the configuration has no effect at all. WDYT?

rubenhoenle: I'm just curious, why would the feature be not available in a region at all? Don't we roll out the same service version to all regions?

rubenhoenle: > the resulting diff could be the desired signal that the configuration has no effect at all. WDYT?

And no, that's nothing we should aim for. State drifts are pain for our users and we shouldn't expose such problems to our users.... What brings me back to the question above :wink:

h3adex: > I'm just curious, why would the feature be not available in a region at all? Don't we roll out the same service version to all regions?

e.g. in eu02 ske is rolled out but telemetry router/link is still WIP. This will be also the case for next regions

tobias-pfaffelmoser-ske: As checked together with @MichaelEischer today:
- SKE API change for audit log feature rolled out to all public regions -> API returns non-null audit object in any scenario
- If audit feature generally not available/activated (at e.g. eu02), API also returns audit object with `{ "enabled": false }`. However, sending `{ "enabled": true }` to API results in returned error in respective region.

rubenhoenle (changes requested):
As written in the PR template checklist, the Acceptance tests / E2E tests need to be adjusted to include your new field: [ske_acc_test.go](https://github.com/stackitcloud/terraform-provider-stackit/blob/e45f2f5aa8ca8770ee2c4ca6fc922dc5ecaaea63/stackit/internal/services/ske/ske_acc_test.go)

**Min test**

In the min test only required fields must be set in the test configuration.

Since your field is optional you only have to add checks that the new field has the correct default value.

https://github.com/stackitcloud/terraform-provider-stackit/blob/e45f2f5aa8ca8770ee2c4ca6fc922dc5ecaaea63/stackit/internal/services/ske/ske_acc_test.go#L120-L265

**Max test**

For the Max test you have to add your field to the Terraform config: 

https://github.com/stackitcloud/terraform-provider-stackit/blob/e45f2f5aa8ca8770ee2c4ca6fc922dc5ecaaea63/stackit/internal/services/ske/testdata/resource-max.tf#L41-L109

Afterwards add a new variable for your field:

https://github.com/stackitcloud/terraform-provider-stackit/blob/e45f2f5aa8ca8770ee2c4ca6fc922dc5ecaaea63/stackit/internal/services/ske/ske_acc_test.go#L57-L97 

Also update the field to test not only the create operation but also the update operation:

https://github.com/stackitcloud/terraform-provider-stackit/blob/e45f2f5aa8ca8770ee2c4ca6fc922dc5ecaaea63/stackit/internal/services/ske/ske_acc_test.go#L110-L118

Then add the checks for your new field to the text steps:

https://github.com/stackitcloud/terraform-provider-stackit/blob/e45f2f5aa8ca8770ee2c4ca6fc922dc5ecaaea63/stackit/internal/services/ske/ske_acc_test.go#L267-L509

stackit/internal/services/ske/ske_acc_test.go
rubenhoenle, line 201: Please also check the output of the datasource here (min test)

rubenhoenle, line 258: Please also check it here after the update (min test)

rubenhoenle (commented):
Looks good besides that ✔️

stackit/internal/services/ske/ske_acc_test.go
tobias-pfaffelmoser-ske: Done

tobias-pfaffelmoser-ske: Done

rubenhoenle:
@tobias-pfaffelmoser-ske could you please resolve the conflicts so we can merge this? :sweat_smile:

rubenhoenle (dismissed):
I rebased the PR and solved the conflicts. Please take care of this next time on your own @tobias-pfaffelmoser-ske .

stackit/internal/services/ske/cluster/resource_test.go
```diff
@@ -2442,6 +2445,64 @@ func TestToNetworkPayload(t *testing.T) {
 	}
 }
 
+func TestToAuditPayload(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name    string
+		input   types.Object
+		want    *ske.Audit
+		wantErr bool
```
cgoetz-inovex:
```suggestion
```
`wantErr` is never `true`

stackit/internal/services/ske/cluster/resource_test.go
```diff
@@ -2442,6 +2445,64 @@ func TestToNetworkPayload(t *testing.T) {
...
+			m := &Model{
+				Audit: tt.input,
+			}
+			got, err := toAuditPayload(t.Context(), m)
+			if err != nil && !tt.wantErr {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if err == nil && tt.wantErr {
+				t.Fatalf("expected error, but got none")
+			}
```
cgoetz-inovex:
```suggestion
			if err != nil {
				t.Fatalf("unexpected error: %v", err)
			}
```

stackit/internal/services/ske/cluster/resource_test.go
```diff
@@ -2817,6 +2878,75 @@ func TestValidateConfig(t *testing.T) {
...
 }
 
+func TestMapAudit(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name       string
+		input      *ske.Audit
+		stateAudit types.Object
+		want       types.Object
+		wantErr    bool
```
cgoetz-inovex:
```suggestion
```
same as above

stackit/internal/services/ske/cluster/resource_test.go
```diff
@@ -2817,6 +2878,75 @@ func TestValidateConfig(t *testing.T) {
...
+				Audit: tt.input,
+			}
+
+			err := mapAudit(cluster, m)
+			if !tt.wantErr && err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if tt.wantErr && err == nil {
+				t.Fatalf("expected error, but got none")
+			}
```
cgoetz-inovex:
```suggestion
			if err != nil {
				t.Fatalf("unexpected error: %v", err)
			}
```

cgoetz-inovex (dismissed):
some minor cosmetic suggestions, but fine to be merged from my side, if time is an issue.
thx for your contrib

stackit/internal/services/ske/cluster/resource_test.go
tobias-pfaffelmoser-ske: Adapted as suggested

stackit/internal/services/ske/cluster/resource_test.go
tobias-pfaffelmoser-ske: Adapted as suggested

stackit/internal/services/ske/cluster/resource_test.go
tobias-pfaffelmoser-ske: Adapted as suggested

stackit/internal/services/ske/cluster/resource_test.go
tobias-pfaffelmoser-ske: Adapted as suggested

tobias-pfaffelmoser-ske:
> I rebased the PR and solved the conflicts. Please take care of this next time on your own @tobias-pfaffelmoser-ske .

Yes - sure. Wasn't able to handle last week due to vacation-absence. Just rebased the PR again.

tobias-pfaffelmoser-ske:
> some minor cosmetic suggestions, but fine to be merged from my side, if time is an issue. thx for your contrib

@cgoetz-inovex Thx! I addressed your comments. Please re-check. Otherwise, fine to be merged from our side ...

cgoetz-inovex (approved):

cgoetz-inovex:
@tobias-pfaffelmoser-ske, thanks for your contribution

MarcAd1234:
Hey, is there a reason why this was reverted? #1693

marceljk:
Hey @MarcAd1234,

we had some issues in the ske cluster resource, which we wanted to fix first, to have again a working resource (#1673).

The fix is now merged. I'm testing now, if the integration of audit log in the ske cluster still works without any unexpected side effects. If this is working, we will revert the reverted commit and create a new Terraform provider release.

Best regards,
Marcel from the STACKIT Developer Tools Team

PR #1594, opened 2026-07-17
feat(waf): Onboard Managed Rule Set

Manuelvaas:
## Description

relates to STACKITTPR-716

## Checklist

- [x] Issue was linked above


stackit/internal/services/albwaf/managed_rule_set/datasource.go
```diff
@@ -0,0 +1,186 @@
...
+	ctx = tflog.SetField(ctx, "name", name)
+	ctx = tflog.SetField(ctx, "region", region)
+
+	managedRuleSetResp, err := d.client.DefaultAPI.GetManagedRuleSet(ctx, projectId, region, name).Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		if errors.As(err, &oapiErr) && oapiErr.StatusCode == http.StatusNotFound {
+			resp.State.RemoveResource(ctx)
+			return
+		}
```
SerseusWasTaken, line 182: We should add a explicit error message here that the resource was not found, otherwise we get a `This is a bug in the provider, which should be reported in the provider's own issue tracker.` message which is not the case

stackit/internal/services/albwaf/managed_rule_set/resource.go
```diff
@@ -0,0 +1,500 @@
...
+	r.client = apiClient
+	tflog.Info(ctx, "ALB WAF client configured")
+}
+
+func (r *managedRuleSetResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_alb_waf_managed_rule_set"
+}
+
+func (r *managedRuleSetResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
```
SerseusWasTaken, line 140: It looks like the datasource and resource have the same schema, so we could define a map which defines all descriptions of the field once and use it in both files, like the git instance does for example. What do you think?

stackit/internal/services/albwaf/managed_rule_set/resource.go
```diff
@@ -0,0 +1,500 @@
...
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating ALB WAF Managed Rule Set", fmt.Sprintf("Calling API: %v", err))
+		return
+	}
+
+	ctx = core.LogResponse(ctx)
+
+	if createResp.Name == nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating ALB WAF Managed Rule Set", "Got empty Managed Rule Set name")
+		return
+	}
```
SerseusWasTaken, line 327: Are we under the assumption here that no rule set was actually created when we get a empty name as a response? Since it might be that we actually do create a resource but return an error here nonetheless.

stackit/internal/services/albwaf/managed_rule_set/resource.go
```diff
@@ -0,0 +1,500 @@
...
+	diags = resp.State.Set(ctx, model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	tflog.Info(ctx, "ALB WAF Managed Rule Set created")
+}
+
+func (r *managedRuleSetResource) Update(ctx context.Context, _ resource.UpdateRequest, resp *resource.UpdateResponse) { // nolint:gocritic // function signature required by Terraform
+	core.LogAndAddError(ctx, &resp.Diagnostics, "Ressource not updatable", "alb Managed Rule Set is not updatable")
```
SerseusWasTaken, line 354: There is a PATCH endpoint for the MRS: https://docs.api.stackit.cloud/documentation/alb-waf/version/v1beta#tag/ManagedRuleSet/operation/WAFService_PatchManagedRuleSet 
Wouldn't that be what we want to call here? Or does the API not do anything at the moment?

stackit/internal/services/albwaf/managed_rule_set/resource.go
```diff
@@ -0,0 +1,500 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	ctx = core.InitProviderContext(ctx)
+
+	projectId := model.ProjectId.ValueString()
+	region := r.providerData.GetRegionWithOverride(model.Region)
+	ctx = tflog.SetField(ctx, "project_id", projectId)
+	ctx = tflog.SetField(ctx, "region", region)
```
SerseusWasTaken, line 307: We could add the name to the tflogs from the model here as well

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/hashicorp/terraform-plugin-testing v1.16.0
 	github.com/stackitcloud/stackit-sdk-go/core v0.26.0
 	github.com/stackitcloud/stackit-sdk-go/services/alb v0.16.0
+	github.com/stackitcloud/stackit-sdk-go/services/albwaf v0.9.0
```
SerseusWasTaken: Nitpick: version 0.10.0 is already available

SerseusWasTaken (commented):
The acc test fails for me with: 
```
TF_ACC=1 go test -timeout=20m -v stackit/internal/services/albwaf/albwaf_acc_test.go 
=== RUN   TestAccManagedRuleSet
    albwaf_acc_test.go:43: Step 1/4 error: Check failed: Check 2/5 error: stackit_alb_waf_managed_rule_set.managed_rule_set: Attribute 'region' expected "EU01", got "eu01"
    panic.go:694: Error running post-test destroy, there may be dangling resources: getting resp: 400 Bad Request, status code 400, Body: {"code":3,"message":"one or more fields are invalid","details":[{"@type":"type.googleapis.com/google.rpc.BadRequest.FieldViolation","field":"region","description":"Expected region t
o be \"eu01\", but got \"EU01\"."}]}
--- FAIL: TestAccManagedRuleSet (1.44s)
FAIL
FAIL    command-line-arguments  2.171s
FAIL
```
But i know some other tests don't run for me but which do for others, so maybe this is some weird issue with my machine again 🤔

stackit/internal/services/albwaf/managed_rule_set/resource.go
Manuelvaas: We had a discussion on that and decided that we will not bring the update endpoint yet. Feel free to reach out to me for further information.

stackit/internal/services/albwaf/managed_rule_set/resource.go
Manuelvaas: The response of the name is required in this context.

SerseusWasTaken (dismissed):

SerseusWasTaken (approved):

PR #1601, opened 2026-07-21
feat(cdn) - Feature parity

matheuspolitano:
## Description

(https://jira.schwarz/browse/STACKITCDN-1451)

## Checklist

- [x] Issue was linked above


cgoetz-inovex:
Hi @matheuspolitano,
thanks for your contribution, I've created an internal issue to review and test this PR.

stackit/internal/services/cdn/distribution/datasource_test.go
```diff
@@ -66,6 +66,9 @@ func TestMapDataSourceFields(t *testing.T) {
 		"backend":                backend,
 		"regions":                regionsFixture,
 		"blocked_countries":      blockedCountriesFixture,
+		"blocked_ips":            types.ListValueMust(types.StringType, []attr.Value{}),
```
Manuelvaas, line 69: Could you add some tests where the values are not null

stackit/internal/services/cdn/distribution/resource_test.go
```diff
@@ -13,6 +13,19 @@ import (
 	cdnSdk "github.com/stackitcloud/stackit-sdk-go/services/cdn/v1api"
 )
 
+func createTestConfig(vals map[string]attr.Value) types.Object {
```
Manuelvaas, line 16: Please add tests here as well to test the values. The createTestConfig is kind of special. Maybe you could use a fixture config like in other tests instead of this?

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -714,6 +723,23 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
 						// API response (empty list).
 						Default: listdefault.StaticValue(types.ListValueMust(types.StringType, []attr.Value{})),
 					},
+					"blocked_ips": schema.ListAttribute{
+						Optional:    true,
+						Computed:    true, // Required when using Default
+						Description: schemaDescriptions["config_blocked_ips"],
+						ElementType: types.StringType,
+						Default:     listdefault.StaticValue(types.ListValueMust(types.StringType, []attr.Value{})),
```
Manuelvaas: We try to avoid using defaults because then the value gets actively set by Terraform, and you cannot use a server-side default or change this later. Feel free to reach out to me if you have further questions regarding this.

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -714,6 +723,23 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
+					},
+					"default_cache_duration": schema.StringAttribute{
+						Optional:    true,
+						Computed:    true,
```
Manuelvaas, line 735: Is this value changed by the service?

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -714,6 +723,23 @@ func (r *distributionResource) Schema(_ context.Context, _ resource.SchemaReques
...
+						Description: schemaDescriptions["config_default_cache_duration"],
+					},
+					"monthly_limit_bytes": schema.Int64Attribute{
+						Optional:    true,
+						Computed:    true,
```
Manuelvaas, line 740: Is this value changed by the service?

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -993,6 +1026,12 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
 	if !utils.IsUndefined(configModel.StripResponseCookies) {
 		configPatch.StripResponseCookies = configModel.StripResponseCookies.ValueBoolPointer()
 	}
+	if !utils.IsUndefined(configModel.DefaultCacheDuration) {
```
Manuelvaas, line 1033: Interesting that there is a toCreatePayload but no toUpdatePayload function 🤔 If you like, you could introduce that, but it's not a must-have because it was already that way. But would be much cleaner and better testable.

Manuelvaas (changes requested):

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: If we don't define a schema Default, Terraform plans the omitted attribute as null. But when the API returns [], mapFields writes [] into the state. Because null (plan) does not match [] (state), Terraform throws the "Provider produced inconsistent result after apply" error.  Adding Default: listdefault.StaticValue(...) forces the planned value to be [] in the plan before apply, making it match the returned state perfectly and preventing the error.

matheuspolitano: Also this value will never change so is safe

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: Yep

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: Yep

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: We have a plan to refactor in the future

stackit/internal/services/cdn/distribution/resource_test.go
matheuspolitano: Its done

stackit/internal/services/cdn/distribution/datasource_test.go
matheuspolitano: Done

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: After your dm i've removed the default field

Manuelvaas (dismissed):

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -993,6 +1030,12 @@ func (r *distributionResource) Update(ctx context.Context, req resource.UpdateRe
...
+		configPatch.DefaultCacheDuration = *cdnSdk.NewNullableString(conversion.StringValueToPointer(configModel.DefaultCacheDuration))
+	}
+	if !utils.IsUndefined(configModel.MonthlyLimitBytes) {
+		configPatch.MonthlyLimitBytes = *cdnSdk.NewNullableInt64(conversion.Int64ValueToPointer(configModel.MonthlyLimitBytes))
```
cgoetz-inovex, line 1037: example config:
```hcl
   resource "stackit_cdn_distribution" "example-distribution" {
     project_id = var.project_id

     config = {
       backend = {
         type       = "http"
         origin_url = "https://example.com"
       }

       regions = ["EU"]

       default_cache_duration = "PT2H"
       monthly_limit_bytes    = 104857600
     }
   }
```
run `terraform apply` then remove `default_cache_duration` and `monthly_limit_bytes`.
Then run `apply` again.

snippet from state:
```
        default_cache_duration = "PT2H"
        forward_host_header    = false
        monthly_limit_bytes    = 104857600
```

I'd expect that the state contains the same values as when a distribution is created without these two values in the config:

```
default_cache_duration = "P1D"
```
and `monthly_limit_bytes` missing completely from state.

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1446,10 +1489,40 @@ func mapFields(ctx context.Context, distribution *cdnSdk.Distribution, model *Mo
...
+	if diags.HasError() {
+		return core.DiagsToError(diags)
+	}
+
+	var defaultCacheDuration types.String
+	if distribution.Config.DefaultCacheDuration.IsSet() {
+		defaultCacheDuration = types.StringPointerValue(distribution.Config.DefaultCacheDuration.Get())
+	} else {
+		defaultCacheDuration = types.StringNull()
+	}
```
cgoetz-inovex: Nitpick: could be simplified
```suggestion
	defaultCacheDuration := types.StringPointerValue(distribution.Config.DefultCacheDuration.Get())
```
These NullableStringTypes are a bit unintuitive, `Get()` is always save to call, it will just return nil. When passed to `types.StringPointerValue()` this would result in a `StringNull`:
```go
func TestNullableStringToTerraformString(t *testing.T) {
	uninitialized := cdnSdk.NullableString{}

	setNil := cdnSdk.NullableString{}
	setNil.Set(nil)

	uninitializedValue := types.StringPointerValue(uninitialized.Get())
	setNilValue := types.StringPointerValue(setNil.Get())

	if diff := cmp.Diff(types.StringNull(), uninitializedValue); diff != "" {
		t.Fatalf("uninitialized NullableString converted unexpectedly (-want +got):\n%s", diff)
	}
	if diff := cmp.Diff(uninitializedValue, setNilValue); diff != "" {
		t.Fatalf("uninitialized and nil NullableString conversions differ (-uninitialized +nil):\n%s", diff)
	}
}
```

stackit/internal/services/cdn/distribution/resource.go
```diff
@@ -1446,10 +1489,40 @@ func mapFields(ctx context.Context, distribution *cdnSdk.Distribution, model *Mo
...
+
+	var monthlyLimitBytes types.Int64
+	if distribution.Config.MonthlyLimitBytes.IsSet() {
+		monthlyLimitBytes = types.Int64PointerValue(distribution.Config.MonthlyLimitBytes.Get())
+	} else {
+		monthlyLimitBytes = types.Int64Null()
+	}
```
cgoetz-inovex: same as above

stackit/internal/services/cdn/distribution/datasource.go
```diff
@@ -678,11 +694,41 @@ func mapDataSourceFields(ctx context.Context, distribution *cdnSdk.Distribution,
...
+	} else {
+		defaultCacheDuration = types.StringNull()
+	}
+
+	var monthlyLimitBytes types.Int64
+	if distribution.Config.MonthlyLimitBytes.IsSet() {
+		monthlyLimitBytes = types.Int64PointerValue(distribution.Config.MonthlyLimitBytes.Get())
+	} else {
+		monthlyLimitBytes = types.Int64Null()
+	}
```
cgoetz-inovex, line 722: see resource for nitpick suggestion, but these aren't a merge blocker from my side

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: This is computed value if you remove will keep the lastest value we are reflecting our api behavior

stackit/internal/services/cdn/distribution/datasource_test.go
cgoetz-inovex: I think the commit where this was done is missing from the PR

stackit/internal/services/cdn/distribution/resource.go
matheuspolitano: I added this information in our docs

stackit/internal/services/cdn/distribution/resource.go

stackit/internal/services/cdn/distribution/datasource.go
matheuspolitano: Done

stackit/internal/services/cdn/distribution/resource.go

stackit/internal/services/cdn/distribution/datasource_test.go
cgoetz-inovex: resolving, this is covered in acc tests

cgoetz-inovex (approved):

PR #1604, opened 2026-07-22
Feat/sqlserverflex encryption support

SerseusWasTaken:
## Description
Adds encryption support and two missing read-only fields

relates to STACKITTPR-756

## Checklist

- [ ] Issue was linked above


stackit/internal/services/sqlserverflex/instance/datasource.go
```diff
@@ -115,6 +122,29 @@ func (r *instanceDataSource) Schema(_ context.Context, _ datasource.SchemaReques
 				Description: descriptions["backup_schedule"],
 				Computed:    true,
 			},
+			"encryption": schema.SingleNestedAttribute{
+				Description: descriptions["encryption"],
+				Optional:    true,
```
marceljk: should be only computed in the datasource
```suggestion
```

stackit/internal/services/sqlserverflex/instance/resource.go
```diff
@@ -1001,6 +1098,23 @@ func mapFields(ctx context.Context, resp *sqlserverflex.GetInstanceResponse, mod
...
+			"kek_key_version": types.StringValue(resp.Encryption.KekKeyVersion),
+			"service_account": types.StringValue(resp.Encryption.ServiceAccount),
+		}
+		encryptionObject, diags = types.ObjectValue(encryptionTypes, encryptionValues)
+		if diags.HasError() {
+			return fmt.Errorf("creating encryption: %w", core.DiagsToError(diags))
+		}
+	} else {
+		encryptionObject = types.ObjectNull(encryptionTypes)
+	}
```
marceljk: nitpick, you can just instantiate the var `encryptionObject` as null object, then you don't need the else block
```suggestion
	encryptionObject := types.ObjectNull(encryptionTypes)
	if resp.Encryption != nil {
		encryptionValues = map[string]attr.Value{
			"kek_key_id":      types.StringValue(resp.Encryption.KekKeyId),
			"kek_keyring_id":  types.StringValue(resp.Encryption.KekKeyRingId),
			"kek_key_version": types.StringValue(resp.Encryption.KekKeyVersion),
			"service_account": types.StringValue(resp.Encryption.ServiceAccount),
		}
		encryptionObject, diags = types.ObjectValue(encryptionTypes, encryptionValues)
		if diags.HasError() {
			return fmt.Errorf("creating encryption: %w", core.DiagsToError(diags))
		}
	}
```

stackit/internal/services/sqlserverflex/instance/resource.go
```diff
@@ -381,6 +409,52 @@ func (r *instanceResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"encryption": schema.SingleNestedAttribute{
+				Description: descriptions["encryption"],
+				Optional:    true,
+				Computed:    true,
```
marceljk: Should also work without computed. At least for postgresflex it's working for me in my current branch without any issue. Didn't test it yet for sqlserverflex
```suggestion
```

stackit/internal/services/sqlserverflex/instance/resource.go
```diff
@@ -1023,14 +1137,33 @@ func mapFields(ctx context.Context, resp *sqlserverflex.GetInstanceResponse, mod
...
+		}
+		if !encryption.KekKeyRingId.IsNull() || !encryption.KekKeyRingId.IsUnknown() {
+			encryptionPayload.KekKeyRingId = encryption.KekKeyRingId.ValueString()
+		}
+		if !encryption.KekKeyVersion.IsNull() || !encryption.KekKeyVersion.IsUnknown() {
+			encryptionPayload.KekKeyVersion = encryption.KekKeyVersion.ValueString()
+		}
+		if !encryption.ServiceAccount.IsNull() || !encryption.ServiceAccount.IsUnknown() {
+			encryptionPayload.ServiceAccount = encryption.ServiceAccount.ValueString()
+		}
```
marceljk: all these attributes are marked as required in the object, so we can assume here that they are set and don't need to do here additional checks. The call to `ValueString()` is anyways safe and returns always an empty string, if nothing was set by the user.
Can be simplified to the following:
```suggestion
		encryptionPayload = &sqlserverflex.InstanceEncryption{
			KekKeyId:       encryption.KekKeyId.ValueString(),
			KekKeyRingId:   encryption.KekKeyRingId.ValueString(),
			KekKeyVersion:  encryption.KekKeyVersion.ValueString(),
			ServiceAccount: encryption.ServiceAccount.ValueString(),
		}
```

stackit/internal/services/sqlserverflex/sqlserverflex_acc_test.go
```diff
@@ -314,6 +337,11 @@ func TestAccSQLServerFlexMaxResource(t *testing.T) {
 					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_instance.instance", "retention_days", testutil.ConvertConfigVariable(testConfigVarsMax["retention_days"])),
 					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_instance.instance", "backup_schedule", testutil.ConvertConfigVariable(testConfigVarsMax["backup_schedule"])),
 
+					resource.TestCheckResourceAttrSet("stackit_sqlserverflex_instance.instance", "encryption.kek_key_id"),
+					resource.TestCheckResourceAttrSet("stackit_sqlserverflex_instance.instance", "encryption.kek_keyring_id"),
+					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "encryption.kek_key_version", testutil.ConvertConfigVariable(testConfigVarsMax["kek_key_version"])),
+					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "encryption.service_account", testutil.ConvertConfigVariable(testConfigVarsMax["service_account_email"])),
```
marceljk: should be checked here from the data source instead of the resource

stackit/internal/services/sqlserverflex/sqlserverflex_acc_test.go
```diff
@@ -250,6 +259,15 @@ func TestAccSQLServerFlexMaxResource(t *testing.T) {
...
 				ConfigVariables: testConfigVarsMax,
 				Check: resource.ComposeAggregateTestCheckFunc(
+					// Key and Keyring
+					resource.ComposeAggregateTestCheckFunc(
+						resource.TestCheckResourceAttr("stackit_kms_keyring.keyring", "project_id", testutil.ProjectId),
+						resource.TestCheckResourceAttr("stackit_kms_keyring.keyring", "region", testutil.Region),
+						resource.TestCheckResourceAttr("stackit_kms_keyring.keyring", "display_name", testutil.ConvertConfigVariable(testConfigVarsMax["display_name"])),
+						resource.TestCheckResourceAttrSet("stackit_kms_keyring.keyring", "keyring_id"),
+						resource.TestCheckNoResourceAttr("stackit_kms_keyring.keyring", "description"),
+					),
```
marceljk: optional: the kms resources doesn't need to be checked here, because they aren't part of the resources which should be tested here.

stackit/internal/services/sqlserverflex/sqlserverflex_acc_test.go
```diff
@@ -147,7 +156,7 @@ func TestAccSQLServerFlexMinResource(t *testing.T) {
 					resource.TestCheckResourceAttrSet("data.stackit_sqlserverflex_user.user", "user_id"),
 					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_user.user", "username", testutil.ConvertConfigVariable(testConfigVarsMin["username"])),
 					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_user.user", "roles.#", "1"),
-					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_user.user", "roles.0", testutil.ConvertConfigVariable(testConfigVarsMax["role"])),
+					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_user.user", "roles.0", testutil.ConvertConfigVariable(testConfigVarsMin["role"])),
```
marceljk, line 150: good catch

stackit/internal/services/sqlserverflex/testdata/resource-max.tf
```diff
@@ -24,8 +48,14 @@ resource "stackit_sqlserverflex_instance" "instance" {
...
-  retention_days  = var.retention_days
-  version         = var.server_version
+  retention_days = var.retention_days
+  version        = var.server_version
+  encryption = {
+    kek_key_id      = stackit_kms_key.key.key_id
+    kek_keyring_id  = stackit_kms_keyring.keyring.keyring_id
+    kek_key_version = var.kek_key_version
+    service_account = var.service_account_email
+  }
```
marceljk: At the moment this feature is only in private preview and not available in every organisation. Because of this, I would suggest, to maybe split this into a whole separate max acc tests. So that we have one test with the GA features and one, where the private preview is tested

stackit/internal/services/sqlserverflex/instance/datasource.go
```diff
@@ -149,6 +179,16 @@ func (r *instanceDataSource) Schema(_ context.Context, _ datasource.SchemaReques
...
+					"instance_address": schema.ListAttribute{
+						Description: descriptions["instance_address"],
+						ElementType: types.StringType,
+						Computed:    true,
+					},
+					"router_address": schema.ListAttribute{
+						Description: descriptions["router_address"],
+						ElementType: types.StringType,
+						Computed:    true,
+					},
```
marceljk: wrong types here, should be StringAttribute instead of ListAttribute

stackit/internal/services/sqlserverflex/instance/resource.go
```diff
@@ -454,6 +528,16 @@ func (r *instanceResource) Schema(_ context.Context, _ resource.SchemaRequest, r
...
+					"instance_address": schema.ListAttribute{
+						Description: descriptions["instance_address"],
+						ElementType: types.StringType,
+						Computed:    true,
+					},
+					"router_address": schema.ListAttribute{
+						Description: descriptions["router_address"],
+						ElementType: types.StringType,
+						Computed:    true,
+					},
```
marceljk: also here wrong types, should be String instead of List

stackit/internal/services/sqlserverflex/instance/resource.go
SerseusWasTaken: Right, good point 😄

stackit/internal/services/sqlserverflex/sqlserverflex_acc_test.go
SerseusWasTaken: Is that always the case for the acceptance tests? I don't see any harm here, if anything, this makes potentially debugging the test easier

stackit/internal/services/sqlserverflex/testdata/resource-max.tf
SerseusWasTaken: Good idea, i will split the tests

stackit/internal/services/sqlserverflex/instance/datasource.go
SerseusWasTaken: Yeah my bad 🤦 i wonder how the tests didnt catch that...

stackit/internal/services/sqlserverflex/instance/resource.go
SerseusWasTaken: Actually i do have to keep one if statement since `encryption.KekKeyId.ValueString()` and such panic if encryption is nil. The if statements guarded against that. i'll introduce a single `encryption != nil` guard

stackit/internal/services/sqlserverflex/sqlserverflex_acc_test.go
marceljk: I think at the moment it's a bit mixed. If a check is here wrong, it will stop continue with the other tests. But I'm also fine with leaving it like it is

stackit/internal/services/sqlserverflex/sqlserverflex_acc_test.go
```diff
@@ -156,7 +147,7 @@ func TestAccSQLServerFlexMinResource(t *testing.T) {
 					resource.TestCheckResourceAttrSet("data.stackit_sqlserverflex_user.user", "user_id"),
 					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_user.user", "username", testutil.ConvertConfigVariable(testConfigVarsMin["username"])),
 					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_user.user", "roles.#", "1"),
-					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_user.user", "roles.0", testutil.ConvertConfigVariable(testConfigVarsMin["role"])),
+					resource.TestCheckResourceAttr("data.stackit_sqlserverflex_user.user", "roles.0", testutil.ConvertConfigVariable(testConfigVarsMax["role"])),
```
marceljk: this can be changed back to `testConfigVarsMin`

stackit/internal/services/sqlserverflex/sqlserverflexx_private_preview_acc_test.go
```diff
@@ -0,0 +1,265 @@
...
+					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "version", testutil.ConvertConfigVariable(configVarsMaxPrivatePreviewUpdated()["server_version"])),
+					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "retention_days", testutil.ConvertConfigVariable(configVarsMaxPrivatePreviewUpdated()["retention_days"])),
+					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "backup_schedule", testutil.ConvertConfigVariable(configVarsMaxPrivatePreviewUpdated()["backup_schedule"])),
+
+					resource.TestCheckResourceAttrSet("stackit_sqlserverflex_instance.instance", "encryption.kek_key_id"),
+					resource.TestCheckResourceAttrSet("stackit_sqlserverflex_instance.instance", "encryption.kek_keyring_id"),
+					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "encryption.kek_key_version", testutil.ConvertConfigVariable(testConfigVarsMax["kek_key_version"])),
+					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "encryption.service_account", testutil.ConvertConfigVariable(testConfigVarsMax["service_account_email"])),
```
marceljk: Even if the values doesn't change, it's better to use for the checks always the updated config. Otherwise it can get confusing when debugging there some issues in the future
```suggestion
					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "encryption.kek_key_version", testutil.ConvertConfigVariable(configVarsMaxPrivatePreviewUpdated()["kek_key_version"])),
					resource.TestCheckResourceAttr("stackit_sqlserverflex_instance.instance", "encryption.service_account", testutil.ConvertConfigVariable(configVarsMaxPrivatePreviewUpdated()["service_account_email"])),
```

stackit/internal/services/sqlserverflex/sqlserverflex_private_preview_acc_test.go
marceljk, line 1: naming of the test file 
```diff
- sqlserverflexx_private_preview_acc_test.go
+ sqlserverflex_private_preview_acc_test.go
```

stackit/internal/services/sqlserverflex/sqlserverflexx_private_preview_acc_test.go
```diff
@@ -0,0 +1,265 @@
...
+			// Creation
+			{
+				Config: testutil.NewConfigBuilder().BuildProviderConfig() + "\n" + resourcePrivatePreviewConfig,
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("stackit_sqlserverflex_instance.instance", plancheck.ResourceActionCreate),
+						plancheck.ExpectResourceAction("stackit_sqlserverflex_user.user", plancheck.ResourceActionCreate),
+					},
+				},
+				ConfigVariables: testConfigVarsMax,
```
marceljk:
```suggestion
				ConfigVariables: testConfigVarsMaxPrivatePreview,
```
wrong config used in the whole file

docs/resources/sqlserverflex_instance.md
```diff
@@ -39,6 +39,7 @@ resource "stackit_sqlserverflex_instance" "example" {
```
marceljk, line 42: Can you quickly update the example here. I missed that during the migration from v2 to v3 and I think it would be better to provide here an example which doesn't return any warnings.

```hcl

resource "stackit_sqlserverflex_instance" "instance" {
  project_id      = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
  name            = "example-instance"
  flavor_id       = "4.16-Single"
  backup_schedule = "0 0 * * *"
  network = {
    acl = ["XXX.XXX.XXX.X/XX", "XX.XXX.XX.X/XX"]
  }
  storage = {
    class = "premium-perf2-stackit"
    size  = 5
  }
  retention_days = 32
  version        = "2022"
}
```

marceljk (dismissed):
LGTM and works as expected :)

marceljk (approved):

PR #1607, opened 2026-07-22
feat(postgresflex): migrate postgresflex resources from v2 to v3 API

marceljk:
## Description

relates to STACKITTPR-734

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/postgresflex/database/resource.go
+++ b/stackit/internal/services/postgresflex/database/resource.go
@@ -436,6 +436,7 @@ func (r *databaseResource) Delete(ctx context.Context, req resource.DeleteReques
 	err = utils.RetryRequestWithoutResponse(ctx, r.client.DefaultAPI.DeleteDatabase(ctx, projectId, region, instanceId, databaseId).Execute, config)
 	if err != nil {
 		if oapiErr, ok := errors.AsType[*oapierror.GenericOpenAPIError](err); ok && oapiErr.StatusCode == http.StatusNotFound {
+			resp.State.RemoveResource(ctx)
 			return
 		}
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error deleting database", fmt.Sprintf("Calling API: %v", err))
--- a/stackit/internal/services/postgresflex/database/resource.go
+++ b/stackit/internal/services/postgresflex/database/resource.go
@@ -25,7 +25,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/stackitcloud/stackit-sdk-go/core/oapierror"
-	postgresflex "github.com/stackitcloud/stackit-sdk-go/services/postgresflex/v3beta1api"
+	postgresflex "github.com/stackitcloud/stackit-sdk-go/services/postgresflex/v3api"
 )
 
 // Ensure the implementation satisfies the expected interfaces.
@@ -284,6 +284,7 @@ func (r *databaseResource) Read(ctx context.Context, req resource.ReadRequest, r
 	ctx = tflog.SetField(ctx, "database_id", databaseIdStr)
 	ctx = tflog.SetField(ctx, "region", region)
 
+	// In v2 the ID was a string. This was changed in the v3 API.
 	databaseId, err := strconv.ParseInt(databaseIdStr, 10, 64)
 	if err != nil {
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading database", fmt.Sprintf("Parsing database ID: %v", err))
@@ -338,6 +339,7 @@ func (r *databaseResource) Update(ctx context.Context, req resource.UpdateReques
 	ctx = tflog.SetField(ctx, "region", region)
 	ctx = tflog.SetField(ctx, "database_id", databaseIdStr)
 
+	// In v2 the ID was a string. This was changed in the v3 API.
 	databaseId, err := strconv.ParseInt(databaseIdStr, 10, 64)
 	if err != nil {
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating database", fmt.Sprintf("Parsing database ID: %v", err))
@@ -414,6 +416,7 @@ func (r *databaseResource) Delete(ctx context.Context, req resource.DeleteReques
 	ctx = tflog.SetField(ctx, "database_id", databaseIdStr)
 	ctx = tflog.SetField(ctx, "region", region)
 
+	// In v2 the ID was a string. This was changed in the v3 API.
 	databaseId, err := strconv.ParseInt(databaseIdStr, 10, 64)
 	if err != nil {
 		core.LogAndAddError(ctx, &resp.Diagnostics, "Error deleting database", fmt.Sprintf("Parsing database ID: %v", err))
--- a/stackit/internal/services/postgresflex/database/resource.go
+++ b/stackit/internal/services/postgresflex/database/resource.go
@@ -5,7 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/conversion"
 	postgresflexUtils "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/postgresflex/utils"
@@ -23,7 +25,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/stackitcloud/stackit-sdk-go/core/oapierror"
-	postgresflex "github.com/stackitcloud/stackit-sdk-go/services/postgresflex/v2api"
+	postgresflex "github.com/stackitcloud/stackit-sdk-go/services/postgresflex/v3beta1api"
 )
 
 // Ensure the implementation satisfies the expected interfaces.
@@ -166,16 +168,10 @@ func (r *databaseResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 			"name": schema.StringAttribute{
 				Description: descriptions["name"],
 				Required:    true,
-				PlanModifiers: []planmodifier.String{
-					stringplanmodifier.RequiresReplace(),
-				},
 			},
 			"owner": schema.StringAttribute{
 				Description: descriptions["owner"],
 				Required:    true,
-				PlanModifiers: []planmodifier.String{
-					stringplanmodifier.RequiresReplace(),
-				},
 			},
 			"region": schema.StringAttribute{
 				Optional: true,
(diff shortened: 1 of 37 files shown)


stackit/internal/services/postgresflex/database/resource.go
```diff
@@ -23,7 +25,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/stackitcloud/stackit-sdk-go/core/oapierror"
-	postgresflex "github.com/stackitcloud/stackit-sdk-go/services/postgresflex/v2api"
+	postgresflex "github.com/stackitcloud/stackit-sdk-go/services/postgresflex/v3beta1api"
```
marceljk: must be updated to v3api before merge

stackit/internal/utils/retry.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if config.Delay <= 0 && config.Backoff == nil {
+		config.Delay = 100 * time.Millisecond
+	}
+
+	retryableSet := make(map[int]bool, len(config.RetryStatusCodes))
+	for _, code := range config.RetryStatusCodes {
+		retryableSet[code] = true
+	}
```
SerseusWasTaken: why use a map here? I think filling a slice with the retry status codes and later checking if a given code should be retried with `slices.Contains` is a bit simpler

stackit/internal/services/postgresflex/database/resource.go
```diff
@@ -306,9 +319,78 @@ func (r *databaseResource) Read(ctx context.Context, req resource.ReadRequest, r
...
+	}
+	err = utils.RetryRequestWithoutResponse(ctx, r.client.DefaultAPI.PartialUpdateDatabase(ctx, projectId, region, instanceId, databaseId).PartialUpdateDatabasePayload(*payload).Execute, config)
+	if err != nil {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating database", fmt.Sprintf("Calling API: %v", err))
+		return
+	}
+
+	ctx = core.LogResponse(ctx)
+
+	database, err := r.client.DefaultAPI.GetDatabase(ctx, projectId, region, instanceId, databaseId).Execute()
```
SerseusWasTaken, line 377: Is Get always available and does not need the retry method?

stackit/internal/services/postgresflex/database/resource.go
```diff
@@ -325,18 +407,35 @@ func (r *databaseResource) Delete(ctx context.Context, req resource.DeleteReques
...
+		},
+		RetryStatusCodes: []int{
+			http.StatusLocked,
+		},
+	}
+	err = utils.RetryRequestWithoutResponse(ctx, r.client.DefaultAPI.DeleteDatabase(ctx, projectId, region, instanceId, databaseId).Execute, config)
+	if err != nil {
+		if oapiErr, ok := errors.AsType[*oapierror.GenericOpenAPIError](err); ok && oapiErr.StatusCode == http.StatusNotFound {
 			return
 		}
```
SerseusWasTaken, line 444:
```suggestion
		if oapiErr, ok := errors.AsType[*oapierror.GenericOpenAPIError](err); ok && oapiErr.StatusCode == http.StatusNotFound {
  resp.State.RemoveResource(ctx)
  return
}
```

stackit/internal/services/postgresflex/instance/resource.go
```diff
@@ -9,19 +9,25 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
+	sdkUtils "github.com/stackitcloud/stackit-sdk-go/core/utils"
 
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/conversion"
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/core"
+	postgresflexUtils "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/postgresflex/utils"
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils"
-	stringplanmodifierCustom "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils/planmodifiers/stringplanmodifier"
+	listplanmodifier2 "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils/planmodifiers/listplanmodifier"
```
SerseusWasTaken: Why `listplanmodifier2` and not just `listplanmodifier`?

stackit/internal/services/postgresflex/instance/resource.go
SerseusWasTaken, line 785: Maybe i'm wrong here, but shouldn't we also have to check for `flavor_id` here? As far as i can tell `Update` only works with the `flavor` field here and not with `flavor_id`. But its up to the user which one they want to use, right?

stackit/internal/services/postgresflex/instance/resource.go
```diff
@@ -592,160 +888,252 @@ func (r *instanceResource) ImportState(ctx context.Context, req resource.ImportS
...
 	var diags diag.Diagnostics
-	if instance.Acl == nil || instance.Acl.Items == nil {
+	if resp.Network.Acl == nil {
 		aclList = types.ListNull(types.StringType)
 	} else {
-		respACL := instance.Acl.Items
 		modelACL, err := utils.ListValueToStringSlice(model.ACL)
 		if err != nil {
 			return err
 		}
```
SerseusWasTaken: Nitpick:
```go
	var aclList = types.ListNull(types.StringType)
	var diags diag.Diagnostics
	if resp.Network.Acl != nil {
		modelACL, err := utils.ListValueToStringSlice(model.ACL)
		if err != nil {
			return err
		}
```
This should do the same, and we don't have to use an else block

stackit/internal/services/postgresflex/user/resource.go
```diff
@@ -434,18 +511,35 @@ func (r *userResource) Delete(ctx context.Context, req resource.DeleteRequest, r
...
+		RetryStatusCodes: []int{
+			http.StatusLocked,
+		},
+	}
+	err = utils.RetryRequestWithoutResponse(ctx, r.client.DefaultAPI.DeleteUser(ctx, projectId, region, instanceId, userId).Execute, config)
 	if err != nil {
-		var oapiErr *oapierror.GenericOpenAPIError
-		if errors.As(err, &oapiErr) && oapiErr.StatusCode == http.StatusNotFound {
+		if oapiErr, ok := errors.AsType[*oapierror.GenericOpenAPIError](err); ok && oapiErr.StatusCode == http.StatusNotFound {
 			return
```
SerseusWasTaken, line 547: I think here this is missing as well: 
`resp.State.RemoveResource(ctx)`

SerseusWasTaken (commented):
Some questions and nitpicks, but looks really clean overall 🙂

stackit/internal/utils/retry.go
marceljk: yes `slice.Contains` would do the same. The implementation with the map is more performant, but because the list will contains just a few entires (probably never more than 3), the improvement isn't significant. I'll change it to slice.Contains to keep it simpler and better understandable

stackit/internal/services/postgresflex/database/resource.go
marceljk: Thanks

stackit/internal/services/postgresflex/instance/resource.go
marceljk: good catch. I probably had the listplanmodifier from the framework imported. Changed it to `listplanmodifier`

stackit/internal/services/postgresflex/instance/resource.go
marceljk: thanks, updated it

stackit/internal/services/postgresflex/user/resource.go
marceljk: As long the Delete method stops without any errors, the framework will automatically remove the resource from the state. But I'll add it anyways do have it also explicit in the code

stackit/internal/services/postgresflex/database/resource.go
marceljk: Based on the docs the Get endpoint doesn't send the lock status code. I asked the team if we can assume, that the GET endpoint isn't effect

stackit/internal/services/postgresflex/instance/resource.go
marceljk: Not sure if I got it correctly here or missed something. The toUpdatePayload method takes care of the  parsed flavorModel and flavor_id. 

https://github.com/stackitcloud/terraform-provider-stackit/blob/93049854f9cc0de70c959d0afa91f3a65a59c9f6/stackit/internal/services/postgresflex/instance/resource.go#L1099-L1106

The configuration of it should be exclusive or, so only flavor_id can be set or the flavor attribute. In the longterm the plan is also to deprecate the flavor attribute and remove it in the future, but for that we need first a new flavor datasource.

stackit/internal/services/postgresflex/user/resource.go
SerseusWasTaken: Ah good to know :)

stackit/internal/services/postgresflex/database/resource.go
marceljk: Just got the response, that the GET endpoint doesn't need the retry

stackit/internal/services/postgresflex/instance/resource.go
SerseusWasTaken: Ah nevermind, i didn't see that the `flavor_id` gets used in `toCreatePayload` and `toUpdatePayload` i thought it was forgotten.

stackit/internal/services/postgresflex/instance/resource.go
```diff
@@ -238,11 +406,117 @@ func (r *instanceResource) Schema(_ context.Context, req resource.SchemaRequest,
...
+				},
+				PlanModifiers: []planmodifier.Int32{
+					int32planmodifier.UseNonNullStateForUnknown(),
+				},
+			},
+			"retention_days": schema.Int32Attribute{
+				Description: descriptions["retention_days"],
+				Optional:    true,
+				Computed:    true,
+			},
```
SerseusWasTaken, line 484: We can add validation here since the api docs state that retention should be a value between 32 and 90

marceljk: done

stackit/internal/services/postgresflex/instance/resource.go
```diff
@@ -9,19 +9,26 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-log/tflog"
+	sdkUtils "github.com/stackitcloud/stackit-sdk-go/core/utils"
 
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/conversion"
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/core"
+	postgresflexUtils "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/postgresflex/utils"
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils"
-	stringplanmodifierCustom "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils/planmodifiers/stringplanmodifier"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils/planmodifiers/listplanmodifier"
+	stringplanmodifier2 "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils/planmodifiers/stringplanmodifier"
```
SerseusWasTaken, line 31:
```suggestion
	stringplanmodifier "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils/planmodifiers/stringplanmodifier"
```

stackit/internal/services/postgresflex/postgresflex_acc_test.go
```diff
@@ -1,302 +1,676 @@
-package postgresflex_test
+package postgresflex
```
SerseusWasTaken: Any reason why this package was changed in both acc tests?

stackit/internal/services/postgresflex/instance/resource.go
marceljk: There is a few lines below also an import from the stringplanmodifier from the terraform framwork. Renaming this one will cause then a conflict.

https://github.com/stackitcloud/terraform-provider-stackit/blob/34ecd69bde72e0a34e6ff782912c1782214a75a0/stackit/internal/services/postgresflex/instance/resource.go#L31-L37

stackit/internal/services/postgresflex/postgresflex_acc_test.go
marceljk: No wasn't intentional. I created complete new files for the acc tests and didn't paid much attention on the package. I'll change it

SerseusWasTaken (approved):
Looks good!

PR #1612, opened 2026-07-24
feat(albwaf): onboard WAF

SerseusWasTaken:
Relates to STACKITTPR-749

## Description
Onboards WAF

## Checklist

- [ ] Issue was linked above


stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -0,0 +1,554 @@
...
+			},
+			"labels": schema.MapAttribute{
+				Description: descriptions["labels"],
+				Optional:    true,
+				ElementType: types.StringType,
+				Validators: []validator.Map{
+					mapvalidator.KeysAre(
+						stringvalidator.RegexMatches(
+							// Regex validates that no stackit- prefix is used
+							regexp.MustCompile(`^(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9_.-]{0,6}[a-zA-Z0-9]|(?:[a-rt-zA-Z0-9][a-zA-Z0-9_.-]{7}|s[a-su-zA-Z0-9_.-][a-zA-Z0-9_.-]{6}|st[b-zA-Z0-9_.-][a-zA-Z0-9_.-]{5}|sta[a-bd-zA-Z0-9_.-][a-zA-Z0-9_.-]{4}|stac[a-jl-zA-Z0-9_.-][a-zA-Z0-9_.-]{3}|stack[a-hj-zA-Z0-9_.-][a-zA-Z0-9_.-]{2}|stacki[a-su-zA-Z0-9_.-][a-zA-Z0-9_.-]|stackit[a-zA-Z0-9_.])[a-zA-Z0-9_.-]{0,54}[a-zA-Z0-9])$`),
```
SerseusWasTaken: This regex is ridiculously complex, but only because negative lookaheads not possible are using `regexp.MustCompile`.

docs/data-sources/alb_waf_configuration.md
```diff
@@ -0,0 +1,56 @@
...
+### Optional
+
+- `custom_rule_group_name` (String) Name of the custom rule group for this WAF Configuration.
+- `labels` (Map of String) User-defined metadata as key-value pairs. Should not exceed 64 entries.
+- `managed_rule_set_name` (String) Name of the managed rule set configuration for this WAF Configuration.
+- `region` (String) The resource region (e.g. eu01). If not defined, the provider region is used.
+
+### Read-Only
+
+- `id` (String) Terraform's internal resource ID. It is structured as "`project_id`,`region`".
```
marceljk: Id should be extended with `name`. The ID should contain always all params, which are needed to call the GET endpoint

docs/resources/alb_waf_configuration.md
```diff
@@ -0,0 +1,56 @@
...
+---
+
+# stackit_alb_waf_configuration (Resource)
+
+albwaf resource schema.
+
+## Example Usage
+
+```terraform
+resource "stackit_alb_waf" "waf" {
```
marceljk: wrong resource name

docs/resources/alb_waf_configuration.md
```diff
@@ -0,0 +1,56 @@
...
+### Optional
+
+- `custom_rule_group_name` (String) Name of the custom rule group for this WAF Configuration.
+- `labels` (Map of String) User-defined metadata as key-value pairs. Should not exceed 64 entries.
+- `managed_rule_set_name` (String) Name of the managed rule set configuration for this WAF Configuration.
+- `region` (String) The resource region (e.g. eu01). If not defined, the provider region is used.
+
+### Read-Only
+
+- `id` (String) Terraform's internal resource ID. It is structured as "`project_id`,`region`".
```
marceljk: same here, it should contain all attributes for the GET endpoint. Otherwise an import is also impossible. For the terraform import, the user sets the ID and based on this information we need to get the right entity

stackit/internal/services/albwaf/waf_configuration/datasource.go
```diff
@@ -0,0 +1,182 @@
...
+	features.CheckBetaResourcesEnabled(ctx, &d.providerData, &resp.Diagnostics, "stackit_alb_waf_configuration", core.Datasource)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	apiClient := utils.ConfigureClient(ctx, &d.providerData, &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	d.client = apiClient
```
marceljk: nitpick: can be summarized to one line
```suggestion
	d.client := utils.ConfigureClient(ctx, &d.providerData, &resp.Diagnostics)
	if resp.Diagnostics.HasError() {
		return
	}
```

stackit/internal/services/albwaf/waf_configuration/datasource.go
```diff
@@ -0,0 +1,182 @@
...
+			"region": schema.StringAttribute{
+				Description: descriptions["region"],
+				Optional:    true,
+				// must be computed to allow for storing the override value from the provider
+				Computed: true,
+			},
+			"name": schema.StringAttribute{
+				Description: descriptions["name"],
+				Required:    true,
+			},
```
marceljk: nitpick: order the attributes from required to optional. So first name and then fregion

stackit/internal/services/albwaf/waf_configuration/datasource.go
```diff
@@ -0,0 +1,182 @@
...
+				ElementType: types.StringType,
+			},
+			"managed_rule_set_name": schema.StringAttribute{
+				Description: descriptions["managed_rule_set_name"],
+				Optional:    true,
+			},
+			"custom_rule_group_name": schema.StringAttribute{
+				Description: descriptions["custom_rule_group_name"],
+				Optional:    true,
+			},
```
marceljk, line 96: there attributes should be set to computed, because in the datasource they can't be set

stackit/internal/services/albwaf/waf_configuration/datasource.go
```diff
@@ -0,0 +1,182 @@
...
+	projectId := model.ProjectId.ValueString()
+	name := model.Name.ValueString()
+	region := d.providerData.GetRegionWithOverride(model.Region)
+	ctx = tflog.SetField(ctx, "project_id", projectId)
+	ctx = tflog.SetField(ctx, "region", region)
+	ctx = tflog.SetField(ctx, "name", name)
+	if name == "" {
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error updating ALB WAF Configuration", "Name must be defined when updating ALB WAF Configuration")
+		return
+	}
```
marceljk: shouldn't be needed here

stackit/internal/services/albwaf/waf_configuration/datasource.go
```diff
@@ -0,0 +1,182 @@
...
+
+	foundWAF, err := d.client.DefaultAPI.GetWAF(ctx, projectId, region, name).Execute()
+	if err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		if errors.As(err, &oapiErr) && oapiErr.StatusCode == http.StatusNotFound {
+			core.LogAndAddError(ctx, &resp.Diagnostics, fmt.Sprintf("ALB WAF Configuration with name %q not found in project %q and region %q", name, projectId, region), err.Error())
+			resp.State.RemoveResource(ctx)
+			return
+		}
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error reading ALB WAF Configuration", err.Error())
```
marceljk, line 130: Preferred error handling in datasources is utils.LogError

https://github.com/stackitcloud/terraform-provider-stackit/blob/b17abe101519e0470ed271dfdb9c9e83e91bb481/stackit/internal/services/edgecloud/plans/datasource.go#L146-L156

stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -0,0 +1,583 @@
...
```
marceljk: Is this a regex from the API docs? Because I couldn't find it there

stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -0,0 +1,583 @@
...
+							"must start and end with an alphanumeric character, may contain dashes, underscores and dots, be 1-63 characters long and NOT start with \"stackit-\""),
+					),
+					mapvalidator.ValueStringsAre(
+						stringvalidator.RegexMatches(
+							regexp.MustCompile(`^(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])$`),
```
marceljk: same here

stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -0,0 +1,583 @@
...
+									ElementType: types.StringType,
+								},
+								"load_balancer_name": schema.StringAttribute{
+									Description: descriptions["load_balancer_name"],
+									Computed:    true,
+								},
+							},
+						},
+					},
+				},
```
marceljk: I would remove this attribute. This will probably always not in sync. Because if you create a waf config and use it then in another resource, terraform would save the old value. So there would be always two terraform applies required.

We should maybe discuss this in our next Daily, how we want to handle these attributes and expose it to the users

stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -0,0 +1,583 @@
...
+	ctx = core.LogResponse(ctx)
+	tflog.Info(ctx, "ALB WAF Configuration deleted")
+}
+
+// Metadata implements [resource.Resource].
+func (r *wafResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_alb_waf_configuration"
+}
+
+// Read implements [resource.Resource].
```
marceljk: Please check the descriptions.

stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -0,0 +1,583 @@
...
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error deleting ALB WAF Configuration", fmt.Sprintf("Calling API: %v", err))
+		return
+	}
+
+	ctx = core.LogResponse(ctx)
+	tflog.Info(ctx, "ALB WAF Configuration deleted")
+}
+
+// Metadata implements [resource.Resource].
+func (r *wafResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
```
marceljk: nitpick: the order of the functions is a bit off without our usual order. Normally we have on the top Metadata, ModifyPlan, Configure, then the CRUD functions, Import and the the toPayload functions

stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -0,0 +1,583 @@
...
+	diags := req.State.Get(ctx, &model)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	ctx = core.InitProviderContext(ctx)
+
+	projectId := model.ProjectId.ValueString()
+	name := model.Name.ValueString()
```
marceljk, line 291: check if name is not empty should be added here. https://github.com/stackitcloud/terraform-provider-stackit/blob/b17abe101519e0470ed271dfdb9c9e83e91bb481/.github/docs/contribution-guide/resource.go#L282-L286

Background information for it: It's possible to wrap a terraform provider for crossplane. There it might calls the Read function, even if the create wasn't called yet.

stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -0,0 +1,583 @@
...
+
+	var labels *map[string]string
+	if !(model.Labels.IsNull() || model.Labels.IsUnknown()) {
+		diags := model.Labels.ElementsAs(ctx, &labels, false)
+		if diags.HasError() {
+			return nil, core.DiagsToError(diags)
+		}
+	}
+	return &waf.UpdateWAFPayload{
+		Name:                model.Name.ValueString(),
```
marceljk: Is the name in the update required? Based on the API docs, the update payload doesn't support a name attribute

docs/data-sources/alb_waf_configuration.md
SerseusWasTaken: oh yeah good catch.

stackit/internal/services/albwaf/waf_configuration/resource.go
SerseusWasTaken: No, not directly. In the API docs is only this description: 
> Key Formatting Rules: Length: 1-63 characters. Characters: Must begin and end with [a-zA-Z0-9]. May contain dashes (-), underscores (_), dots (.), and alphanumerics in between. Keys starting with 'stackit-' are system-reserved; users MUST NOT manage them.

I based the regex on that.

stackit/internal/services/albwaf/waf_configuration/resource.go
SerseusWasTaken: Same as the other one, it is based on the description of the field. No regex is provided in the API docs

stackit/internal/services/albwaf/waf_configuration/resource.go
SerseusWasTaken: Good point! lets talk about it

stackit/internal/services/albwaf/waf_configuration/resource.go
SerseusWasTaken: fair point, i'll reorder them

stackit/internal/services/albwaf/waf_configuration/resource.go
SerseusWasTaken: Not sure what you mean. Name is part of the path parameters and thus required

marceljk: Yes, but the path parameters are usually not set in the payload itself. For example, projectId and region are also part of the path parameters, but they aren't set here, what is also correct.

Looks like the payload is wrong generated. Here in the spec, you can see that name, region and projectId are actually readOnly attributes:
https://github.com/stackitcloud/stackit-api-specifications/blob/166aed4c740698347e3e1c71e69341f77ebd7abb/services/alb-waf/v1beta/alb-waf.json#L898-L920

This should be reported to the service team, so that they update their API spec accordingly.

stackit/internal/services/albwaf/waf_configuration/resource.go
marceljk: If the API docs doesn't provide a regex, I wouldn't add one in the provider itself. Because if we deny setting a value, which the API would support, we need to update the regex in the future, what can be annoying for our users.
In case the service team wants to have the regex here in Terraform, they should extend their api spec first and then we can adapt

stackit/internal/services/albwaf/albwaf_acc_test.go
```diff
@@ -22,44 +22,55 @@ import (
 )
 
 var (
-	//go:embed testdata/managed-rule-set.tf
-	managedRuleSetConfig string
+	//go:embed testdata/resource-max.tf
+	wafConfig string
```
marceljk: In the acc tests the resources should be in independent tests. So for the managed rule set, it should remain like before and for the new waf_configuration, there should be a new test terraform config added with a separate test function.
Otherwise it will be difficult in the future to maintain it, when the test config grows with each new resource. In the new test config for waf_config, you can also add a waf_managed_rule_set. But there you don't need to add checks for the waf_managed_ruleset. 
See for example the test configs of kms: https://github.com/stackitcloud/terraform-provider-stackit/tree/770998656083e829003cc3ed1c067739cc3033d3/stackit/internal/services/kms/testdata

stackit/internal/services/albwaf/waf_configuration/resource.go
SerseusWasTaken: Removed now

stackit/internal/services/albwaf/waf_configuration/resource.go
SerseusWasTaken: Ah yeah, i see now that the name is already declared in the first `UpdateWAF` call.

SerseusWasTaken: Im simply gonna remove that argument, as it works without it anyway.

stackit/internal/services/albwaf/albwaf_acc_test.go
SerseusWasTaken: I see. I re-added the deleted test and minimized the checks for waf managed rule set from the max test

stackit/internal/services/albwaf/waf_configuration/resource.go
SerseusWasTaken: Fair enough, will remove it

Manuelvaas (approved):

PR #1632, opened 2026-07-30
fix(postgresflex): fix state drift when migrating from v2 to v3

SerseusWasTaken:
## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/postgresflex/instance/resource.go
+++ b/stackit/internal/services/postgresflex/instance/resource.go
@@ -28,7 +28,6 @@ import (
 	postgresflexUtils "github.com/stackitcloud/terraform-provider-stackit/stackit/internal/services/postgresflex/utils"
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils"
 	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils/planmodifiers/listplanmodifier"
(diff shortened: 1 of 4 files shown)


SerseusWasTaken:
We should release https://github.com/stackitcloud/stackit-cli/pull/1505 first before releasing this fix

stackit/internal/services/postgresflex/instance/resource.go
```diff
@@ -386,22 +385,15 @@ func (r *instanceResource) Schema(_ context.Context, req resource.SchemaRequest,
 				},
 			},
 			"flavor": schema.SingleNestedAttribute{
-				Optional: true,
-				Computed: true,
+				Optional:           true,
+				Computed:           true,
+				DeprecationMessage: "flavor is deprecated and will be removed after February 2027. Use instead `flavor_id`. You can get the available flavors using the STACKIT-CLI using `stackit postgresflex flavor list`.",
```
marceljk: Would suggested to use the old command. Some users might still have an older version of the CLI installed, when they use the latest terraform version. And if they use the latest version of the CLI, they should get a warning with a hint which command they should use instead.
```suggestion
				DeprecationMessage: "flavor is deprecated and will be removed after February 2027. Use instead `flavor_id`. You can get the available flavors using the STACKIT-CLI using `stackit postgresflex options --flavors`.",
```

marceljk (approved):

PR #1636, opened 2026-07-30
fix(ske): fix planModifiers for update

Manuelvaas:
## Description

relates to STACKITTPR-775

## Checklist

- [x] Issue was linked above

### DIFF
--- a/stackit/internal/utils/planmodifiers/stringplanmodifier/use_state_for_unknown_if.go
+++ b/stackit/internal/utils/planmodifiers/stringplanmodifier/use_state_for_unknown_if.go
@@ -94,7 +94,7 @@ func StringUnchanged(attributePath path.Path) UseStateForUnknownIfFunc { // noli
 	}
 }
 
-// This function is the pathExpression compatible version of StringUnchanged and should replace StringUnchanged in the future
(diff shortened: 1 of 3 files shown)


stackit/internal/utils/planmodifiers/stringplanmodifier/use_state_for_unknown_if.go
```diff
@@ -94,6 +94,39 @@ func StringUnchanged(attributePath path.Path) UseStateForUnknownIfFunc { // noli
 	}
 }
 
+// This function is the pathExpression compatible version of StringUnchanged and should replace StringUnchanged in the future
+func StringUnchangedExpressionFunction(pathExpressions ...path.Expression) UseStateForUnknownIfFunc { // nolint:gocritic // function signature required by Terraform
```
marceljk: I think it would be nice to extend the description a bit, to avoid misuse. It's possible here to provide multiple path.Expressions and someone could assume, that only if all path.Expressions remain unchanged, it will use UseStateForUnknown. But if I got it correctly, it sets UseStateForUnknown as soon one of the path.Expression remained the same.

rubenhoenle:
fixed by #1644

Issue #1646, opened 2026-08-02
SKE cluster deletion hangs for up to 90 minutes when the cluster left a load balancer behind


FabianHardt:
## Description

Deleting an SKE cluster that ran a `Service` of type `LoadBalancer` leaves the
cluster in `STATE_DELETING` for a long time — in our case over an hour, and the
provider would have waited up to its 90 minute timeout.

The cause is the load balancer the cloud controller created *inside* the
cluster. It lives in the project, but it belongs to no Terraform state:
`terraform destroy` removes the cluster, nobody removes the load balancer, and
the shoot waits for it.

Nothing in the API says so. `STATE_DELETING` is the only signal, and the
provider's eventual error is `Cluster deletion waiting: ...`, which does not
name a cause.

## Steps to reproduce

```terraform
resource "stackit_ske_cluster" "example" {
  project_id         = var.project_id
  name               = "example"
  kubernetes_version_min = "1.35"

  node_pools = [{
    name         = "pool-01"
    machine_type = "c1a.2d"
    minimum       = 1
    maximum       = 2
  }]
}
```

1. `terraform apply`
2. In the cluster, install anything that creates a `Service` of type
   `LoadBalancer` — an ingress controller, for example. A load balancer appears
   under `stackit load-balancer list -p <project>`.
3. `terraform destroy`

## Actual behavior

`terraform destroy` triggers `DeleteCluster` and then waits. The cluster stays
in `STATE_DELETING`; `stackit load-balancer list` still shows the load balancer
in `STATUS_READY`. Nothing indicates that one is waiting for the other. After
90 minutes the provider gives up with

```
Error deleting cluster: Cluster deletion waiting: ...
```

## Expected behavior

Either of these would resolve it:

1. The shoot cleans up the load balancers it caused when it is deleted, the way
   it created them.
2. Or the cluster status names what the deletion is waiting for, so that both
   a human and the provider can report something better than "still deleting".

The second one alone would already help a lot — the state is currently
indistinguishable from "slow".

## Environment

- OS: macOS 26.6 (arm64)
- Terraform version: `v1.15.8`
- Version of the STACKIT Terraform provider: `v0.106.0`

## Additional information

I do not think the provider should delete the load balancers itself: it never
created them, and picking which ones belong to a given cluster would mean
deleting resources it does not manage. This is why I am filing it as a service
behaviour issue rather than a provider bug.

The known workaround, and what we do in our offboarding, is to delete all
[...]

PR #1647, opened 2026-08-02
docs(ske): warn that load balancers block cluster deletion

FabianHardt:
Documentation and error message only — no behaviour change. Companion to #1646.

## Problem

A `Service` of type `LoadBalancer` makes the cloud controller create a load
balancer in the project. It belongs to no Terraform state, so `terraform
destroy` removes the cluster and nobody removes the load balancer. The shoot
waits for it, and the cluster stays in `STATE_DELETING` until
`DeleteClusterWaitHandler` gives up after 90 minutes.

Nothing points at the cause. `STATE_DELETING` is indistinguishable from "slow",
and the error after the timeout is:

```
Error deleting cluster: Cluster deletion waiting: ...
```

In our case that cost about an hour before we found the load balancer still in
`STATUS_READY` next to a cluster that would not go away.

## Change

1. A note on the `stackit_ske_cluster` resource description, rendered as a
   warning callout:

   > Before destroying a cluster, remove any `Service` of type `LoadBalancer`
   > from it. Such a service makes the cloud controller create a load balancer
   > in the project, which belongs to no Terraform state. If it still exists,
   > the cluster stays in `STATE_DELETING` until this resource times out after
   > 90 minutes.

2. The timeout error now names the likely cause and the command to check:

   ```
   Cluster deletion waiting: <err>.

   A cluster can stay in STATE_DELETING because of resources it created outside
   of Terraform. The most common one is a load balancer: a Service of type
   LoadBalancer makes the cloud controller create one in the project, and it is
   not removed by deleting the cluster. Check with
   `stackit load-balancer list -p <project>` and remove any leftovers, then retry.
   ```

`docs/resources/ske_cluster.md` was regenerated with `scripts/tfplugindocs.sh`;
only that file changed.

## Deliberately not doing

Deleting the load balancers from the provider. `stackit_ske_cluster` never
created them, and working out which ones belong to a given cluster would mean
deleting resources the provider does not manage — in a project with a second
cluster or a manually created load balancer that would be actively harmful.

The real fix belongs on the service side and is filed as #1646: either the
shoot cleans up what it caused, or the cluster status says what it is waiting
for. This PR only shortens the road to that knowledge in the meantime.

## Checks

```
gofmt -l / go vet                      -> clean
go test ./stackit/internal/services/ske/...  -> ok
scripts/tfplugindocs.sh                -> only ske_cluster.md changed
```

### DIFF
--- a/docs/resources/ske_cluster.md
+++ b/docs/resources/ske_cluster.md
@@ -5,6 +5,7 @@ subcategory: ""
 description: |-
   SKE Cluster Resource schema. Must have a region specified in the provider configuration.
   -> When updating node_pools of a stackit_ske_cluster, the Terraform plan might appear incorrect as it matches the node pools by index rather than by name. However, the SKE API correctly identifies node pools by name and applies the intended changes. Please review your changes carefully to ensure the correct configuration will be applied.
+  ~> Before destroying a cluster, remove any Service of type LoadBalancer from it. Such a service makes the cloud controller create a load balancer in the project, which belongs to no Terraform state. If it still exists, the cluster stays in STATE_DELETING until this resource times out after 90 minutes.
 ---
 
 # stackit_ske_cluster (Resource)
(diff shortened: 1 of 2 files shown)


cgoetz-inovex:
closing for now as discussed in linke issue

PR #1653, opened 2026-08-03
chore: remove deprecated field region from the provider

SerseusWasTaken:
Relates to STACKITTPR-693

## Description
Removes the deprecated region field from the provider now that the deprecation period is over. 

## Checklist

- [ ] Issue was linked above


templates/guides/ske_log_alerts.md.tmpl
```diff
@@ -12,10 +12,6 @@ This guide walks you through setting up log-based alerting in STACKIT Observabil
    Begin by configuring the STACKIT and Kubernetes providers to connect to the STACKIT services.
 
    ```hcl
-   provider "stackit" {
```
rubenhoenle, line 15: I think the provider block is still needed in this example...

I would leave it there with and use the `default_region` field instead of the `region` field.

docs/guides/ske_log_alerts.md
```diff
@@ -12,10 +12,6 @@ This guide walks you through setting up log-based alerting in STACKIT Observabil
    Begin by configuring the STACKIT and Kubernetes providers to connect to the STACKIT services.
 
    ```hcl
-   provider "stackit" {
```
rubenhoenle, line 15: same here - it's the result of the template adjusted above

templates/guides/ske_log_alerts.md.tmpl
SerseusWasTaken: To me that seems that it's a no-op, which is why i removed it. I don't mind having it here though, so ill simply re-add it

rubenhoenle (approved):

PR #1663, opened 2026-08-05
chore(provider): refactor sdk client creation to allow mock injection

rubenhoenle:
### Intention

The intention of this refactoring is

- Being able to inject mocked SDK HTTP clients into resources. This makes unit-testing of resources easier.
  - Old approach ([example](https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/services/ske/ske_test.go#L20-L151)): In the past we had to start a local mock server and use a custom endpoint to run the SDK requests against it. This doesn't allow for type safety during mocked tests, etc. 
- New approach ([example](https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/services/iaas/volume/unittest/resource_test.go#L19-L89)): We can inject a mocked client directly into the resource / datasource. We don't have to spin up mock servers and use custom endpoints. Furthermore we have type safety via our [SDK mocks](https://github.com/stackitcloud/stackit-sdk-go/blob/48c975eec6181d8d76521cd5ca4fa0c2dbdb2382/services/iaas/v2api/api_default_mock.go#L21-L296).

> [!NOTE]  
> Furthermore (and probably more important) this refactoring allows us to roll out automatic retries on e.g. HTTP 429 API errors more easily across the whole Terraform provider with all it's resource and datasource implementations. See #1764 and #1771 for reference.

### What changed inside the resource and datasource implementations

Previously our resources looked like this: In each resource in the `Configure` function a new SDK client was created. The client was created using a util function (boilerplate code which had to implemented for every service, also the tests were copied in boilerplate manner):

```golang
// Configure adds the provider configured client to the resource.
func (r *volumeResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
	var ok bool
	r.providerData, ok = conversion.ParseProviderData(ctx, req.ProviderData, &resp.Diagnostics)
	if !ok {
		return
	}

	apiClient := iaasUtils.ConfigureClient(ctx, &r.providerData, &resp.Diagnostics)
		return
	}
	r.client = apiClient
	tflog.Info(ctx, "iaas client configured")
}
```

Now instead we have a pre-populated struct `core.ClientCollection` which is part of the provider data and includes an instance of each SDK HTTP client.

```golang
// Configure adds the provider configured client to the resource.
func (r *volumeResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
	var ok bool
	providerData, clients, ok := core.ParseProviderData(ctx, req.ProviderData, &resp.Diagnostics)
	if !ok {
		return
	}

	r.providerData = providerData
	r.client = clients.IaaSv2Client

	tflog.Info(ctx, "iaas client configured")
}
```

### How this works internally

First of all we have the `core.ClientCollection` struct which should be passed to every resource and datasource:

https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/core.go#L137-L178

This client collection has to be initialized somehow with all the SDK clients. Therefore we have the `initClientCollection` function and the `ClientFactory`

https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/clientutils.go#L56-L99

https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/clientutils.go#L101-L196

The `ClientFactory` interface allows us to have a `DefaultClientFactory` implementation which is used in regular production use of the STACKIT Terraform provider.

https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/clientutils.go#L198-L800

The `MockClientFactory` is another implementation of the `ClientFactory` interface and allows us to inject mocked SDK clients into the provider for API-mocked testing of resource and datasource implementations.

https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/clienttestutils.go#L46-L409

The injection of the mock client factory happens here:

https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/provider.go#L161-L178

### Splitting the provider data into two structs

In Terraform plugin framework you can have *one* struct which is passed as provider data to each resource and datasource.

In the past this struct looked like this:

https://github.com/stackitcloud/terraform-provider-stackit/blob/0c8ee622e48ea12adb99f82237d4b6fb9b6e3a4a/stackit/internal/core/core.go#L37-L85

It e.g. included a round tripper for authentication of all the SDK clients which were created using the util functions in each resource/datasource configure method.

https://github.com/stackitcloud/terraform-provider-stackit/blob/0c8ee622e48ea12adb99f82237d4b6fb9b6e3a4a/stackit/internal/core/core.go#L38

https://github.com/stackitcloud/terraform-provider-stackit/blob/0c8ee622e48ea12adb99f82237d4b6fb9b6e3a4a/stackit/internal/services/iaas/utils/util.go#L23

Since to allow for easy mocking we need to pass pre-populated SDK clients to all resources/datasources in the future we need to add this client to provider data. The provider data struct would grow therefore massively with API clients for every STACKIT service and service version.

At the same we need to keep this provider data struct within our resource/datasource implementation structs to make the `region` handling possible (using a `default_region` configured on provider level and allowing our TF users to optionally override it in each resource/datasource configuration).

https://github.com/stackitcloud/terraform-provider-stackit/blob/0c8ee622e48ea12adb99f82237d4b6fb9b6e3a4a/stackit/internal/services/iaas/volume/resource.go#L95-L99

Reminder: We can only pass *one* provider data struct to each resource/datasource. Therefore I decided to split the provider data. The provider data struct passed should consist of two parts: The formerly known provider data (including e.g. the `default_region` config value or the list of enabled provider experiments) and the client collection.

The provider data struct is meant to be kept in each regional resource/datasource implementation in the same way we did until now. The client collection should only be used in the `Configure` method of each resource/datasource implementation.

Both structs are tied together using the `core.providerDataInternal` struct.

https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/core.go#L131-L135

The `providerDataInternal` struct type isn't exposed by design from the `core` package. This way the inside the resource and datasource implementations the provider data stays the same like we've previously known it (it even shrinked since we don't need to keep all custom endpoint configs, etc. in it any more).

https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/core.go#L180-L192

### Ensuring code quality using automation

Two measures were taken to assert the code quality after this implementation:

- A [tfclientcollection](https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/tools/linters/tfclientcollection/tfclientcollection.go) linter was implemented to make sure the `core.ClientCollection` struct is only consumed from `core.ParseProviderData` but stored in the resource/datasource implementation structs or passed as a parameter to other functions called by the resource/datasource `Configure` method implementations. 
  - The `core.ClientCollection` struct can't be package-private in the `core` package because there's [one exception](https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/golang-ci.yaml#L123-L128) to that: the [generic IAM rolebinding](https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/services/iam/rolebindings/v1/generic/resource.go#L51) resource/datasource implementations.
- A unit test was added to ensure each client available in the `core.ClientCollection` struct is actually initialized using the `ClientFactory` interface in the `initClientCollection` function: [test](https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/clientutils_test.go#L84-L111)

Furthermore for testing all the client initialization implementations of the `DefaultClientFactory` some [generic test handler](https://github.com/stackitcloud/terraform-provider-stackit/blob/f7a24419da9cc14d9f8c55862cebd6ead42012f5/stackit/internal/core/clientutils_test.go#L193-L273) was implemented to get rid of the insane amount of boilerplate code we maintained beforehand.

## Checklist

- [ ] Issue was linked above


stackit/internal/core/clientutils.go
```diff
@@ -0,0 +1,800 @@
...
+
+func (t *RetryTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	base := t.Base
+	if base == nil {
+		base = http.DefaultTransport
+	}
+
+	// Preserve request body for retries if present
+	var bodyBytes []byte
+	if req.Body != nil && req.Body != http.NoBody {
```
SerseusWasTaken: This may be a potential nil pointer dereference if `req` is nil

tools/linters/tfclientcollection/tfclientcollection.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/golangci/plugin-module-register/register"
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/inspect"
+	"golang.org/x/tools/go/ast/inspector"
+)
+
+var Analyzer = &analysis.Analyzer{
+	Name:     "tfclientcollection",
+	Doc:      "Prevents the keeping the core.ClientCollection struct from being kept in a resource. Only clients should be extracted from it and kept.",
```
SerseusWasTaken:
```suggestion
	Doc:      "Prevents the core.ClientCollection struct from being kept in a resource. Only clients should be extracted from it and kept.",
```

PR #1667, opened 2026-08-06
refactor(albwaf): address open issues

SerseusWasTaken:
## Description
- Adds warning when name is planned for change for all 3 resources
- Adds missing fields of example
- Fixes missing destroy function for max and min acc tests
- Fix resource test names

relates to STACKITTPR-792

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/albwaf/albwaf_acc_test.go
+++ b/stackit/internal/services/albwaf/albwaf_acc_test.go
@@ -692,8 +692,10 @@ func testAlbWafCustomRuleGroupDestroy(s *terraform.State) error {
 			continue
 		}
 		// custom rule group transform id: "[projectId],[region],[name]"
-		name := strings.Split(rs.Primary.ID, core.Separator)[2]
-		customRuleGroupsToDestroy = append(customRuleGroupsToDestroy, name)
+		terraformId := strings.Split(rs.Primary.ID, core.Separator)
+		if len(terraformId) == 3 {
+			customRuleGroupsToDestroy = append(customRuleGroupsToDestroy, terraformId[2])
+		}
 	}
 
 	resp, err := client.DefaultAPI.ListCustomRuleGroup(ctx, testutil.ProjectId, testutil.Region).Execute()
@@ -725,8 +727,10 @@ func testAlbWafManagedRuleSetDestroy(s *terraform.State) error {
 			continue
 		}
 		// managed rule set transform id: "[projectId],[region],[name]"
-		name := strings.Split(rs.Primary.ID, core.Separator)[2]
-		managedRuleSetsToDestroy = append(managedRuleSetsToDestroy, name)
+		terraformId := strings.Split(rs.Primary.ID, core.Separator)
+		if len(terraformId) == 3 {
+			managedRuleSetsToDestroy = append(managedRuleSetsToDestroy, terraformId[2])
+		}
 	}
 
 	resp, err := client.DefaultAPI.ListManagedRuleSets(ctx, testutil.ProjectId, testutil.Region).Execute()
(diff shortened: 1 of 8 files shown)


stackit/internal/services/albwaf/custom_rule_group/resource.go
cgoetz-inovex, line 364: Not sure if it matters, just some oddity

There's no early return here if line 355 creates an error:
```go
		resp.Diagnostics.Append(req.State.Get(ctx, &stateModel)...)
```
when this happes we'd still call:
```go
	resp.Diagnostics.Append(resp.Plan.Set(ctx, planModel)...)
```

Also the last conditional is unnecessary, but it looks like we have this in multiple places.

stackit/internal/services/albwaf/managed_rule_set/resource.go
```diff
@@ -237,24 +237,11 @@ func (r *managedRuleSetResource) ModifyPlan(ctx context.Context, req resource.Mo
...
-					fmt.Sprintf(
-						"Changing the \"name\" attribute from %q to %q will destroy and recreate this resource. "+
-							"If another resource (e.g. \"stackit_alb_waf_configuration\") references this managed rule set "+
-							"by name, the replacement will fail. Remove or update that dependency before applying this change.",
-						stateModel.Name.ValueString(), planModel.Name.ValueString(),
-					),
-				)
-			}
+		if !resp.Diagnostics.HasError() {
+			utils.WarnIfNameChanges(stateModel.Name, planModel.Name, "Managed Rule Set", &resp.Diagnostics)
```
cgoetz-inovex, line 244: same conditional logic as one file above

stackit/internal/services/albwaf/waf_configuration/resource.go
```diff
@@ -87,6 +87,14 @@ func (r *wafResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanReq
...
 	}
 
+	if !req.State.Raw.IsNull() {
+		var stateModel Model
+		resp.Diagnostics.Append(req.State.Get(ctx, &stateModel)...)
+		if !resp.Diagnostics.HasError() {
+			albwafUtils.WarnIfNameChanges(stateModel.Name, planModel.Name, "WAF Configuration", &resp.Diagnostics)
+		}
+	}
+
```
cgoetz-inovex, line 97: same as above

stackit/internal/services/albwaf/albwaf_acc_test.go
```diff
@@ -743,3 +744,36 @@ func testAlbWafManagedRuleSetDestroy(s *terraform.State) error {
...
+		wafConfigurationToDestroy = append(wafConfigurationToDestroy, name)
+	}
+
+	resp, err := client.DefaultAPI.ListWAF(ctx, testutil.ProjectId, testutil.Region).Execute()
+	if err != nil {
+		return fmt.Errorf("getting resp: %w", err)
+	}
+
+	for _, item := range resp.Items {
+		if utils.Contains(wafConfigurationToDestroy, item.GetName()) {
```
cgoetz-inovex: https://docs.api.stackit.cloud/documentation/alb-waf/version/v1#tag/WAF/operation/WAFService_ListWAF

Looks like `ListWAF` uses pagination, docs don't mention a default pageSize. Is it enough to return all WAFs created in our acceptance tests?

stackit/internal/services/albwaf/albwaf_acc_test.go
```diff
@@ -743,3 +744,36 @@ func testAlbWafManagedRuleSetDestroy(s *terraform.State) error {
...
+		return err
+	}
+
+	wafConfigurationToDestroy := []string{}
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "stackit_alb_waf_configuration" {
+			continue
+		}
+		// waf configuration transform id: "[projectId],[region],[name]"
+		name := strings.Split(rs.Primary.ID, core.Separator)[2]
```
cgoetz-inovex: Sometimes I get an index out of bound panic on lines like these when running acceptance tests because `rs.Primary.ID == ""`.
Id add an if here to prevent dangling resources.

cgoetz-inovex (dismissed):
issue mentions: 
> - Place Min over Max test
probably missing from this PR

stackit/internal/services/albwaf/custom_rule_group/resource.go
SerseusWasTaken: From my understanding we're appending a list of errors here, so there is no need to necessarily return early here. I'm not sure if that's how it really works but if we skip the early return here, we are able to report the warning and any issues from parsing the `planModel` in one `terraform apply` command. If we instead return early, multiple "round trips" might be needed for that, no?

stackit/internal/services/albwaf/albwaf_acc_test.go
SerseusWasTaken: Good catch, i will just increase the page size to 100 to avoid this issue

stackit/internal/services/albwaf/managed_rule_set/resource.go
SerseusWasTaken: Yeah, but i would keep that here. Since the function `utils.WarnIfNameChanges` requires having successfully parsed the state model (since you need to pass the model's name attribute) i think this is the responsibility of the caller

cgoetz-inovex (approved):

Issue #1677, opened 2026-08-10
`stackit_redis_instance` forces replacement on parameter change


codello:
## Description

When I change the `parameters` of a `stackit_redis_instance`, the redis instance is force-replaced, because its `region` changed. However, the region is actually the same value (provider default).

## Steps to reproduce

Apply the following configuration:

```terraform
terraform {
  required_version = ">= 1.11.1"
  required_providers {
    stackit = {
      source  = "stackitcloud/stackit"
      version = "0.108.0"
    }
  }
}

provider "stackit" {
  default_region = "eu01"
}

resource "stackit_redis_instance" "test" {
  project_id = "<some-id>"
  name       = "repro-test"
  plan_name  = "stackit-redis-1.4.10-single"
  version    = "7"

  #parameters = {
  #  maxmemory_policy = "noeviction"
  #}
}
```

1. Run `terraform init`
2. Run `terraform apply`
3. Uncomment the `parameters` section (or change the `maxmemory_policy` to `volatile-lru` or something else)
4. Run `terraform apply` again (or `terraform plan`)

## Actual behavior

The first apply creates the redis instance. The second `apply` reports the following change (shortened for readability):

```
  # stackit_redis_instance.test must be replaced
-/+ resource "stackit_redis_instance" "test" {
        ...
        name                 = "repro-test"
      ~ parameters           = {
          ~ max_disk_threshold      = 80 -> (known after apply)
          + maxmemory_policy        = "noeviction"
          ...
        }
      ~ plan_id              = "3849f72f-99cc-4e2c-afda-700d66ef08f2" -> (known after apply)
      ~ region               = "eu01" # forces replacement -> (known after apply) # forces replacement
        # (3 unchanged attributes hidden)
    }
```

## Expected behavior

I would expect this change to work without replacing the redis instance.

## Known Workaround

~Setting the `region` attribute of the redis instance to `region = "eu01"` solves the problem. The redis instance is updated in-place. The error seems to only occur when using the provider-default region.~

The workaround only works during plan time. On `apply` I still get an error (see comment below).

## Environment
 - OS: macOS 26.5.2
 - Terraform version (see `terraform --version`): `OpenTofu v1.12.5`  (also tested with Terraform `v1.15.8`)
 - Version of the STACKIT Terraform provider: `v0.110.0`

**Additional information**

This feels like the kind of problem that might affect other resources as well. I haven't explicitly checked any other resources.

codello:
[...]

PR #1681, opened 2026-08-12
fix(redis, rabbitmq): pass region and instance ID in the correct order on update

johannes-engler-mw:
## Description

relates to #1677

`PartialUpdateInstance` is declared in the redis and rabbitmq SDKs as `(ctx, projectId, regionId, instanceId)`, but both resources call it as `(ctx, projectId, instanceId, region)`. Both parameters are `string`, so the transposition compiles cleanly and only shows up at runtime.

The SDK interpolates positionally into the path template:

```go
localVarPath := ".../v2/projects/{projectId}/regions/{regionId}/instances/{instanceId}"
localVarPath = strings.Replace(localVarPath, "{regionId}", ...r.regionId..., -1)
```

so the instance UUID ends up in the `{regionId}` segment and the API rejects it:

```
│ Error: Error updating instance
│ Calling API: undefined response type, status code 400, Body: Bad Request: Invalid region specified.
```

Every other call in the same two files already passes `(projectId, region, instanceId)` — including the `PartialUpdateInstanceWaitHandler` a few lines below each defect, which is what makes the two call sites look correct at a glance.

### Impact

This breaks **every** in-place update of `stackit_redis_instance` and `stackit_rabbitmq_instance`, not only changes to `parameters`. #1677 reports it via `maxmemory_policy`, but `sgw_acl` goes through the same call, so ACL changes are equally affected.

Worth noting for anyone hitting #1677: setting `region` explicitly in the configuration works around the *plan*-time symptom (the spurious force-replacement), but not this — the value is still passed into the wrong parameter slot at apply time.

### Scope

I checked every service calling `PartialUpdateInstance` against its own SDK signature, since the argument order legitimately differs between SDKs. Only these two are wrong:

| Service | Call site | SDK signature | |
| --- | --- | --- | --- |
| redis | `(projectId, instanceId, region)` | `(projectId, regionId, instanceId)` | ❌ |
| rabbitmq | `(projectId, instanceId, region)` | `(projectId, regionId, instanceId)` | ❌ |
| mongodbflex | `(projectId, instanceId, region)` | `(projectId, instanceId, region)` | ✅ |
| logme, mariadb, opensearch, postgresflex, sqlserverflex | `(projectId, region, instanceId)` | `(projectId, region/regionId, instanceId)` | ✅ |

`mongodbflex` looks like the same shape but its SDK genuinely declares `(projectId, instanceId, region)`, so it is correct as written.

## Checklist

- [x] Issue was linked above
- [ ] Examples were added / adjusted (see `examples/` directory) — not applicable, no schema or usage change
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI) — no doc-affecting change
- [ ] Unit tests got implemented or updated — see note below
- [ ] Acceptance tests got implemented or updated

**On tests:** the existing unit tests for these resources cover the helpers (`mapFields`, `toCreatePayload`, ...) and pass both before and after this change — they never reach the `Update` method, so they cannot catch this class of defect. Verifying it genuinely requires an acceptance test that performs an update against the API, which I cannot run here. Happy to add one if you would like it in this PR, though it may fit better as a follow-up covering the DSA services together.


johannes-engler-mw:
@GokceGK Any update when this will be released? Its blocking our whole infra deployments.

cgoetz-inovex (approved):

marceljk (approved):

cgoetz-inovex:
Hi @johannes-engler-mw,
thanks for your contribution, executed the acceptance tests, we'll release a new provider version later today.

PR #1689, opened 2026-08-14
Onboarding Intake(2/3): Onboard intakes resource

devanshcache:
## Description

This PR onboards the new STACKIT Intake ([ticket](https://jira.schwarz/browse/STACKITINT-1353)) service into the Terraform provider.

Intake is composed of three components:

Intake Runners: dedicated, isolated runtime data ingestion environment
Intakes: a specific data stream or topic within an Intake Runner
Intake Users: provides secure access credentials for your applications to connect to your Intake

This PR contains the Intakes part only to make a quicker and less overwhelming review.

## Checklist

- [x] Issue was linked above


cgoetz-inovex:
Hi @devanshcache ,
thanks for your contribution, I've created an internal tracking issue for the review of your changes

devanshcache:
Hi @cgoetz-inovex, Thank you looking forward for your feedback :)

examples/resources/stackit_intakes/resource.tf
```diff
@@ -0,0 +1,15 @@
...
+  catalog_auth_type            = "dremio"
+  catalog_warehouse            = "default"
+  catalog_uri                  = "https://dremio.eu01.onstackit.cloud/iceberg"
+  dremio_token_endpoint        = "https://dremio.eu01.onstackit.cloud/oauth/token"
+  dremio_personal_access_token = "my-dremio-pat-token"
+
+  labels = {
+    "env" = "development"
+  }
+}
```
Fyusel, line 15: Please also add "import-by-string-id.tf" (import statement)

stackit/internal/services/intake/intakes/datasource.go
```diff
@@ -0,0 +1,222 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	d.client = apiClient
+	tflog.Info(ctx, "Intakes client configured for data source")
+}
+
+// Schema defines the schema for the data source
+func (d *intakesDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	descriptions := map[string]string{ //nolint:gosec // descriptions
```
Fyusel, line 61: The description can be shared between datasource and resource.

stackit/internal/services/intake/intakes/datasource.go
```diff
@@ -0,0 +1,222 @@
...
+			},
+			"create_time": schema.StringAttribute{
+				Description: descriptions["create_time"],
+				Computed:    true,
+			},
+			"region": schema.StringAttribute{
+				Optional:    true,
+				Description: descriptions["region"],
+			},
+			"dremio_personal_access_token": schema.StringAttribute{
```
Fyusel, line 136: access_token should not be stored in state.
Add write only here (see for example telemetrylink resource)

stackit/internal/services/intake/intakes/datasource.go
```diff
@@ -0,0 +1,222 @@
...
+				Validators: []validator.String{
+					validate.UUID(),
+					validate.NoSeparator(),
+				},
+			},
+			"runner_id": schema.StringAttribute{
+				Description: descriptions["runner_id"],
+				Computed:    true,
+			},
+			"name": schema.StringAttribute{
```
Fyusel, line 111: This is "displayName" in the API, right?
If so this should be renamed to _displayName_ to be consistent with the API docs.
Additionally a validator should be added.

stackit/internal/services/intake/intakes/datasource.go
```diff
@@ -0,0 +1,222 @@
...
+				Description: descriptions["name"],
+				Computed:    true,
+			},
+			"description": schema.StringAttribute{
+				Description: descriptions["description"],
+				Computed:    true,
+			},
```
Fyusel, line 118: _deadLetterTopic_ and _topic_ is missing?

stackit/internal/services/intake/intakes/resource.go
```diff
@@ -0,0 +1,747 @@
...
+
+	resp.Diagnostics.Append(resp.Plan.Set(ctx, planModel)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// Schema defines the schema for the data source
+func (r *intakesResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	descriptions := map[string]string{ //nolint:gosec // descriptions
```
Fyusel, line 126: Same here: Can be shared between datasource

stackit/internal/services/intake/intakes/resource.go
```diff
@@ -0,0 +1,747 @@
...
+				},
+			},
+			"intake_id": schema.StringAttribute{
+				Description: descriptions["intake_id"],
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"name": schema.StringAttribute{
```
Fyusel, line 185: displayName?

stackit/internal/services/intake/intakes/resource.go
```diff
@@ -0,0 +1,747 @@
...
+			},
+			"region": schema.StringAttribute{
+				Optional:    true,
+				Computed:    true,
+				Description: descriptions["region"],
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"dremio_personal_access_token": schema.StringAttribute{
```
Fyusel, line 228: write only (see other comment)

stackit/internal/services/intake/intakes/resource.go
```diff
@@ -0,0 +1,747 @@
...
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"name": schema.StringAttribute{
+				Description: descriptions["name"],
+				Required:    true,
+			},
+			"description": schema.StringAttribute{
```
Fyusel, line 189: _deadLetterTopic_ and _topic_ missing

stackit/internal/services/intake/intakes/resource_test.go
```diff
@@ -0,0 +1,304 @@
...
+				return
+			}
+			if !tt.wantErr {
+				if diff := cmp.Diff(tt.expected, payload); diff != "" {
+					t.Errorf("toUpdatePayload mismatch (-want +got):\n%s", diff)
+				}
+			}
+		})
+	}
+}
```
Fyusel, line 304: _TestIntakesSavesIDsOnError_ is missing (see CONTRIBUTING.md)

stackit/internal/services/intake/intake_acc_test.go
```diff
@@ -20,70 +28,315 @@ import (
...
+	tempConfig["intake_name"] = config.StringVariable(intakeNameMinUpd)
+	return tempConfig
+}
+
+func testIntakesConfigVarsMaxUpdated(base config.Variables) config.Variables {
+	tempConfig := make(config.Variables, len(base))
+	maps.Copy(tempConfig, base)
+	tempConfig["intake_name"] = config.StringVariable(intakeNameMaxUpd)
+	tempConfig["description"] = config.StringVariable("Updated full intake description")
+	tempConfig["max_messages_per_hour"] = config.IntegerVariable(1100)
```
Fyusel, line 140: Would add also update values for the _catalog_ stuff here

stackit/internal/services/intake/intake_acc_test.go
```diff
@@ -20,70 +28,315 @@ import (
...
+		"project_id":            config.StringVariable(testutil.ProjectId),
+		"name":                  config.StringVariable(runnerNameMax),
+		"region":                config.StringVariable(testutil.Region),
+		"description":           config.StringVariable("An example runner for Intake"),
+		"max_message_size_kib":  config.IntegerVariable(1024),
+		"max_messages_per_hour": config.IntegerVariable(1100),
+	}
+}
+
+func testIntakesConfigVarsMin() config.Variables {
```
Fyusel, line 78: In the API docs only _uri_ and _warhouse_ of catalog is marked as required. So the dremio part could be removed here (auth not marked as required). API doc bug or intentionally?

stackit/internal/services/intake/intakes/resource.go
```diff
@@ -0,0 +1,747 @@
...
+				Description: descriptions["catalog_table_name"],
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"catalog_uri": schema.StringAttribute{
+				Description: descriptions["catalog_uri"],
+				Optional:    true,
```
Fyusel, line 284: marked as required in API docs

stackit/internal/services/intake/intakes/resource.go
```diff
@@ -0,0 +1,747 @@
...
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"catalog_warehouse": schema.StringAttribute{
+				Description: descriptions["catalog_warehouse"],
+				Optional:    true,
```
Fyusel, line 292: same here: marked as required in api docs (https://docs.api.stackit.cloud/documentation/intake/version/v1beta#tag/intakes/operation/create-intake)

stackit/internal/services/intake/intake_acc_test.go
```diff
@@ -20,70 +28,315 @@ import (
...
+	tempConfig := make(config.Variables, len(base))
+	maps.Copy(tempConfig, base)
+	tempConfig["intake_name"] = config.StringVariable(intakeNameMaxUpd)
+	tempConfig["description"] = config.StringVariable("Updated full intake description")
+	tempConfig["max_messages_per_hour"] = config.IntegerVariable(1100)
 	return tempConfig
 }
 
+// getDremioPAT authenticates against Dremio UI API, enables PAT support key, resolves user UUID, and issues a PAT
+func getDremioPAT(ctx context.Context, uiEndpoint, username, password string) (string, error) {
```
Fyusel, line 145: This function does not feel right.
Can you explain please why this is needed?

Fyusel (changes requested):

stackit/internal/services/intake/intake_acc_test.go
Fyusel: I just had a deeper look into this and found that there is a documentation regarding the dremio PAT (https://docs.stackit.cloud/products/data-and-ai/intake/how-tos/configure-dremio-for-stackit-intake-integration/).

How is the customer supposed to use this in terraform? Such a _getDremioPAT()_ function cannot be used by a customer. First a dremio instance needs to be created, then the customer needs to activate PAT for that instance via UI, get the dremio PAT via UI as well and is then able to create a intake instance?

devanshcache: Please note that this is an acceptance test not a Terraform resource. To run it, we need to create an Intake instance, which requires obtaining a Dremio PAT.

Fyusel: Yes, I am aware of that.
Those questions apply in general.

stackit/internal/services/intake/intake_acc_test.go
devanshcache: It is intentional in the API:
- API allows `catalog.auth` to be omitted (defaulting to `type: none` on the backend) to support unauthenticated Iceberg catalogs. In the provider schema, all Dremio fields are marked `Optional: true`.
- However in live acceptance tests (`TestAccIntakesMin`) terraform waits for the intake to reach `active` state. Without a reachable Iceberg catalogthe intake fails to connect and enters a `failed` state, breaking the test waiter.
- because Dremio is currently the only live Iceberg catalog available in our test environment, provisioning Dremio was necessary for the runner to connect successfully and reach `active`.

stackit/internal/services/intake/intake_acc_test.go
devanshcache: can we resolve this comment?

stackit/internal/services/intake/intakes/resource.go
devanshcache: good catch

stackit/internal/services/intake/intakes/resource.go
devanshcache: good catch

stackit/internal/services/intake/intakes/datasource.go
devanshcache: good catch

stackit/internal/services/intake/intakes/resource_test.go
Fyusel: still open

stackit/internal/services/intake/intake_acc_test.go
Fyusel: This doesn't actually resolve my original concern. My point was never about the test itself, but about what _getDremioPAT()_ exposes: 

There is currently no way for a customer to obtain a Dremio PAT through Terraform or any STACKIT API. The real flow today is: 

1. create the Dremio instance and user via Terraform
2. log into the Dremio UI by hand
3. manually flip on PAT support
4. manually generate the token and paste it into dremio_personal_access_token_wo. 

That's not a test artifact, that's the actual customer experience and honestly, it's a mess. Forcing a manual UI detour just to get a secret into the config defeats the point of using Terraform at all.

We already have the right pattern elsewhere in this provider: _stackit_telemetryrouter_accesstoken_, _stackit_logs_accesstoken_, _stackit_modelexperiments_token_, _stackit_modelserving_token_, and _stackit_edgecloud_token_ all mint their secret server-side via the STACKIT API at resource create time, using the STACKIT project/service-account credentials the customer already has. No login against a separate system required. That's the correct shape for this too: a STACKIT API endpoint that issues a Dremio PAT for a given Dremio user, authenticated with STACKIT project credentials, not Dremio username/password.

_getDremioPAT()_ in the acceptance test is effectively proving this gap by reverse-engineering undocumented Dremio UI endpoints (/apiv2/settings/..., /api/v3/user/...) just to get the test to run end-to-end. I don't want that pattern to exist even as a test helper. If the only way to make our own acceptance test work is to script around Dremio's internal UI API, that's a signal the product is missing a real API, not that the test needs a workaround.

devanshcache: Thanks for the feedback. We agree that a native token-minting API/resource is the ideal long-term design.

Since token issuance is not yet available via the public API, the resource was designed to take the token as a credential input.

To address the concern with `getDremioPAT()` and avoid undocumented UI endpoints in the test suite, we could remove the helper entirely and load the token via an environment variable in the acceptance test (skipping the step if not set). 

However, this requires maintaining a long-lived, pre-configured Dremio instance for CI and Dremio PATs have a maximum lifetime of 180 days, it would require periodic manual rotation to keep CI running.

Would you prefer moving to the environment variable approach despite the rotation overhead or is there another direction you would recommend?

PR #1694, opened 2026-08-17
fix(iaas): read write only field correctly from the config model

SerseusWasTaken:
## Description
Fixes the write-only field `KeyPayloadBase64WriteOnly` by reading it from the config model instead of plan model.
Also adds a regression test based on mocking that verifies that the write only field is sent to the API. 

Relates to STACKITTPR-782

## Checklist

- [x] Issue was linked above


stackit/internal/services/iaas/volume/resource_create_test.go
```diff
@@ -0,0 +1,202 @@
...
+	testProjectId        = "4e684f79-a12c-449d-aa89-bcd9d8aafaf2"
+	testRegion           = "eu01"
+	testVolumeId         = "3dee3fb9-59f0-4f97-8eeb-a4da37d05a00"
+	testKeyPayloadBase64 = "VGhlIHF1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIDEzIGxhenkgZG9ncy4="
+)
+
+// buildCreateRequest builds a resource.CreateRequest from a plan and a config model.
+// Terraform populates write-only attribute values only in the config model - never in the plan or state model.
+// That's why we need both, the plan model AND config model to build the request.
+func buildCreateRequest(ctx context.Context, t *testing.T, schemaResp *resource.SchemaResponse, planModel, configModel *Model) resource.CreateRequest {
```
rubenhoenle: I would actually prefer to have this tested via resource.UnitTest like e.g. here:

https://github.com/stackitcloud/terraform-provider-stackit/blob/de961add37876c2a347d6ec500ba61d86fdd2cf0/stackit/internal/services/sqlserverflex/sqlserverflex_test.go#L59

This is hard to mock then (I know), that's why I worked on https://github.com/stackitcloud/terraform-provider-stackit/pull/1663 . Will make your life easier in that regard 😅

SerseusWasTaken: Good suggestion. Added a commit that implements the test using `resource.UnitTest`. Mocking is not that bad tbh, at least for this simple test case.

marceljk (approved):

PR #1707, opened 2026-08-19
feat(modifiers): add UnchangedPaths

cgoetz-inovex:
- delete unused Int64Unchanged

(cherry picked from commit 4de58022b159d75159ce97316d2bd01db1e15d45)

## Description

relates to #1234

## Checklist

- [ ] Issue was linked above


stackit/internal/utils/planmodifiers/stringplanmodifier/use_state_for_unknown_if.go
```diff
@@ -94,26 +95,36 @@ func StringUnchanged(attributePath path.Path) UseStateForUnknownIfFunc { // noli
...
+
+				var stateValue attr.Value
+				resp.Diagnostics.Append(req.State.GetAttribute(ctx, match, &stateValue)...)
+				if resp.Diagnostics.HasError() {
+					return
+				}
+
+				if !stateValue.Equal(planValue) {
+					allUnchanged = false
+				}
```
SerseusWasTaken, line 125: Maybe we can do a early return here? No need to keep walking the other paths when at least one is changed

stackit/internal/utils/planmodifiers/stringplanmodifier/use_state_for_unknown_if_test.go
```diff
@@ -126,3 +130,246 @@ func TestUseStateForUnknownIf_PlanModifyString(t *testing.T) {
...
+			paths: []path.Expression{allItemValues},
+			plan:  testValues{anchor: "same", items: []string{"one", "new"}},
+			state: testValues{anchor: "same", items: []string{"one", "old"}},
+		},
+		{
+			name:  "no wildcard matches",
+			paths: []path.Expression{allItemValues},
+			plan:  testValues{anchor: "new", items: []string{}},
+			state: testValues{anchor: "old", items: []string{}},
+			want:  true,
```
SerseusWasTaken, line 318: Not sure about this behavior. I think users of the function might silently introduce bugs when this function returns true if i pass a non-existing path on accident. Shouldnt we maybe rather throw an error in that case?

cgoetz-inovex: The path exists here:
```
// state:
"items":tftypes.List[tftypes.Object["value":tftypes.String]]<>,
// plan:
 "items":tftypes.List[tftypes.Object["value":tftypes.String]]<>
```
It's an empty list in both state and plan.

My assumption would be, that an empty list in plan and state qualifies as unchanged.

adding:
```go
			if len(matched) == 0 {
				resp.Diagnostics.AddError(
					"no matches", "foo")
				return
			}
``` 
in the implementation would fail this empty list case here.

Completely non-existing paths are covered by the next case "invalid path".

SerseusWasTaken: ah my bad the next test does do what i meant :D you're right

SerseusWasTaken (approved):

PR #1720, opened 2026-08-21
feat(vpn): add BGP filter resources and gateway/connection SDK updates

steffenkoenig:
## Description

Reflects vpn changes:

- New stackit_vpn_bgp_filter and stackit_vpn_bgp_filter_rule resources and data sources, supporting the new gateway-scoped BGP route filtering API.
- stackit_vpn_gateway gains an optional network_config block (predefined_network_prefix, routing_table_id).
- stackit_vpn_connection's tunnel bgp block gains inbound_filter_id, linking a tunnel's BGP session to a stackit_vpn_bgp_filter, with tri-state set/clear/leave-untouched handling on update.
- integrity_algorithms now additionally accepts sha2_512. sha1 remains supported but is deprecated (SDK enum change, no code change needed since values are derived dynamically); a plan-time warning is now emitted when sha1 is configured, pointing users toward sha2_256/sha2_384/sha2_512 instead.

## Checklist

- [ ] Issue was linked above


cgoetz-inovex:
Hi @steffenkoenig,
thanks for your contribution, I've created an internal issue to review this PR

devpie:
It would be really great to have this functionality in one of the next releases <3

stackit/internal/services/vpn/gateway/resource.go
```diff
@@ -215,6 +225,35 @@ func (r *gatewayResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 					},
 				},
 			},
+			"network_config": schema.SingleNestedAttribute{
+				Description: schemaDescriptions["network_config"],
+				Optional:    true,
+				Attributes: map[string]schema.Attribute{
+					"predefined_network_prefix": schema.ListAttribute{
```
itrich, line 232: It seems that `predefined_network_prefix` must be a string, not an array. The VPN Gateway API blocks any payload with `predefinedNetworkPrefix` set as an array with error `500`. See https://github.com/stackitcloud/stackit-api-specifications/issues/69

We will have to wait for a new release of the Go SDK to fix this.

devpie: Thanks for catching this. One detail is worth pinning down before a new SDK release is planned around it.

This comment reports HTTP `500`, while stackitcloud/stackit-api-specifications#69 reports HTTP `400` for the array payload and includes a response body with `"status": "BAD_REQUEST"`.

The two point at different fixes:

- `400` means the server rejects the array at validation. The contract is then a single string, and only the spec needs correcting.
- `500` means the array is not handled at all. That is a server-side defect on top of the spec fix, and correcting the spec alone would not close it.

I asked the same question in the issue. For context I also added this there: `predefinedNetworkPrefix` carries no `type` at all in the spec, only `items` plus a scalar `example`. It is the only such property in the file, which is why both the Go and the Python generator produce a list.

stackit/internal/services/vpn/gateway/resource.go
```diff
@@ -54,6 +55,11 @@ type BGPGatewayConfigModel struct {
 	OverrideAdvertisedRoutes types.List  `tfsdk:"override_advertised_routes"`
 }
 
+type NetworkConfigModel struct {
```
itrich, line 58: I just tested this branch in our local setup and it throws an error:
```
│ Error: Value Conversion Error
│ 
│   with module.stackit.module.mack_one.module.connect.module.hub.module.dev.module.hub_dev_vpn.stackit_vpn_gateway.this,
│   on modules/stackit-vpn-gateway/main.tf line 15, in resource "stackit_vpn_gateway" "this":
│   15:   network_config = var.network_config
│ 
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.
│ 
│ Path: network_config
│ Target Type: *gateway.NetworkConfigModel
│ Suggested Type: basetypes.ObjectValue
```

We're creating the VPN Gateway in a reusable module with `network_config` set as a variable.

The resource in our Terraform module:
```
resource "stackit_vpn_gateway" "this" {
  project_id   = var.project_id
  display_name = "${data.stackit_resourcemanager_project.this.name}-vpn-gw"
  plan_id      = var.plan_id
  routing_type = var.routing_type
  availability_zones = {
    tunnel1 = var.availability_zones.tunnel1
    tunnel2 = var.availability_zones.tunnel2
  }

  network_config = var.network_config
}
```

and the corresponding variable:
```
variable "network_config" {
  description = "Network configuration for the VPN gateway. Leave null to have the network prefix assigned automatically."
  type = object({
    predefined_network_prefix = optional(list(string))
    routing_table_id          = optional(string)
  })
  default = null
}
```

I think using `basetypes` is safer here for optional values, see also https://github.com/Europa-Park-DD/terraform-provider-stackit/commit/572dc0ca96c9d1068ddac2f5ccaff7b879a0c657#diff-238f9f2b61941094893b6851aab64dac70d5582b03fbd2c22ba6f0d34dc6fd0fR22-R64 on how we set this up.

stackit/internal/services/vpn/gateway/resource.go
devpie: One note on how far this blocker actually reaches, since it does not affect all of `network_config`.

It only hits `predefined_network_prefix`. A block that sets just `routing_table_id` never puts an array on the wire:

- `toNetworkConfigPayload` guards the prefix with `tfutils.IsUndefined`, which is `IsUnknown() || IsNull()`. An omitted `predefined_network_prefix` is null, so `NetworkConfig.PredefinedNetworkPrefix` stays nil.
- The SDK's `NetworkConfig.ToMap` writes `predefinedNetworkPrefix` only when `IsNil` returns false, and it returns true for a nil slice. The key is therefore absent from the request body.

So this configuration is untouched by the spec bug:

```hcl
network_config = {
  routing_table_id = stackit_routing_table.example.routing_table_id
}
```

That covers serialization only. Whether the API accepts a custom routing table is untested in this PR: `TestAccVpnGatewayResourceMax` asserts `TestCheckResourceAttrSet` on `network_config.routing_table_id`, which the default table assigned by the API already satisfies, and `gateway-max.tf` sets only the prefix. A case that passes an existing routing table id would close that gap.

Worth mentioning because it means the routing table part of this PR could be usable ahead of the spec fix.

itrich:
I just executed `terraform plan` with the following `network_config` block
```
  network_config = {
    predefined_network_prefix = ["1.2.3.4/8"]
  }
```

which gives us the following plan:

```
Terraform will perform the following actions:

  # module.stackit.module.mack_one.module.connect.module.hub.module.dev.module.hub_dev_vpn2.stackit_vpn_gateway.this will be created
  + resource "stackit_vpn_gateway" "this" {
      + availability_zones = {
          + tunnel1 = "eu01-1"
          + tunnel2 = "eu01-2"
        }
      + display_name       = "hub-eu01-dev-vpn-gw"
      + gateway_id         = (known after apply)
      + id                 = (known after apply)
      + network_config     = {
          + predefined_network_prefix = [
              + "1.2.3.4/8",
            ]
          + routing_table_id          = (known after apply)
        }
      + plan_id            = "p500"
      + project_id         = "418185e2-3d5a-4e67-bf1c-07bb1717e6f3"
      + region             = "eu01"
      + routing_type       = "POLICY_BASED"
    }
```

I think `routing_table_id` should not be set at all if it's not explicitly set in the resource definition. The same should apply vice-versa for `predefined_network_prefix`.

Moreover, the data source should only report `predefined_network_prefix` or `routing_table_id` if it was explicitly set during creation.

SerseusWasTaken:
Hello @itrich and @devpie, since you are actively testing this PR here i wanted to inform you that i will integrate the `network_config` parameter that this PR implements separately, so we can move with greater pace without being restricted by multiple review rounds on a contribution PR. The rest of this PR will be integrated at a later time. 

Thank you for testing this, i will take your findings into account during the implementation.

Best regards,

itrich:
Thank you @SerseusWasTaken for the update! Implementing the network configuration parameter separately is a huge help to our team! ❤️

PR #1721, opened 2026-08-24
docs: update/fix license file date and company name

tuunit:
## Description

Update company legal name in license file


Fyusel:
Hi @tuunit,
thank you for opening this PR. In our guidance doc it is mentioned that we should use "Schwarz IT KG".

PR #1729, opened 2026-08-31
feat(automation): onboard automation templates list datasource

SerseusWasTaken:
## Description

Relates to STACKITTPR-790

## Checklist

- [ ] Issue was linked above


stackit/internal/services/automation/templates/datasource_test.go
```diff
@@ -0,0 +1,113 @@
...
+			valid: true,
+		},
+		{
+			name: "rejects nil response",
+			model: &model{
+				ProjectId: types.StringValue("project-id"),
+				Region:    types.StringValue("eu01"),
+			},
+			valid: false,
+		},
```
cgoetz-inovex, line 98: these 2 tests are a bit contradictory. From `paginate.All`:
```go
	var items []Item
	for item, err := range Items(request, opts...) {
		if err != nil {
			return items, err
		}
		items = append(items, item)
	}
	return items, nil
```
So an empty, non-nil list would never be returned.

I'd suggest to change the implementation to accept nil slices. From a glance it looks like this would be safe. `slices.SortFunc` does not panic on a nil slice.

SerseusWasTaken: Good hint. I changed the function to allow nil values but in that case to explicitly set `m.Templates` to nil to preserve the actual API response.

examples/data-sources/stackit_automation_templates/data-source.tf
```diff
@@ -0,0 +1,4 @@
+data "stackit_automation_templates" "example" {
+  project_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+  # region is taken from the provider configuration
```
rubenhoenle: Nitpick, but I would either
- add this comment to all of our examples
- leave it out here

Else it could potentially confuse users, especially newcomers to our provider. It's default behavior that the region from the provider configuration is used in the resources/datasources by default, so the examples should all look the same in that regard IMO.

stackit/internal/services/automation/templates/datasource.go
```diff
@@ -0,0 +1,198 @@
...
+
+type model struct {
+	ID        types.String `tfsdk:"id"`
+	ProjectId types.String `tfsdk:"project_id"`
+	Region    types.String `tfsdk:"region"`
+	Templates []template   `tfsdk:"templates"`
+}
+
+type template struct {
+	Id          types.String `tfsdk:"id"`
```
rubenhoenle:
```suggestion
	TemplateId          types.String `tfsdk:"template_id"`
```

Just had a quick chat with Marcel about this in the office to make sure I'm not going to nitpick here: I would not name this "id", since 

- then you don't have 2 id fields in your datasource. 
- Furthermore when you'd have a datasource which GETs a single template, the field for the template id would be called `template_id` there. So it would be consistent to also name it `template_id` here from my perspective. 😅 

Drop me a message if you didn't get my point please, my explanation might me confusing haha.

rubenhoenle (dismissed):

examples/data-sources/stackit_automation_templates/data-source.tf
SerseusWasTaken: agreed, i'll remove it.

stackit/internal/services/automation/templates/datasource.go
SerseusWasTaken: i think i get it, there already is a `ID` field top level of the resource, so this would differentiate it from that one. I'll rename the field

rubenhoenle (approved):

Issue #1737, opened 2026-09-04
sfs: resource pool create times out after a fixed 10 minutes, and the tainted retry collides with the existing pool


devpie:
### Summary

`stackit_sfs_resource_pool` cannot create a pool that STACKIT needs more than 10 minutes to
provision. The resource exposes no `timeouts` block, so the wait handler's default of 10
minutes cannot be raised from configuration. When the wait expires, Terraform marks the
object tainted and replaces it on the next run, which collides with the pool that STACKIT
already created.

### Versions

| | |
|---|---|
| Provider | `stackitcloud/stackit` v0.113.0 |
| SDK | `github.com/stackitcloud/stackit-sdk-go/services/sfs` v0.11.0 (unchanged in v0.11.1) |
| OpenTofu | 1.12.6, darwin_arm64 |
| Region | eu01, availability zone `eu01-m` |

### Where the 10 minutes come from

`CreateResourcePoolWaitHandler` sets a fixed timeout
([sfs/v1api/wait/wait.go, v0.11.0](https://github.com/stackitcloud/stackit-sdk-go/blob/services/sfs/v0.11.0/services/sfs/v1api/wait/wait.go#L39)):

```go
handler.SetTimeout(10 * time.Minute)
```

The provider calls it without overriding that value
([sfs/resourcepool/resource.go, v0.113.0](https://github.com/stackitcloud/terraform-provider-stackit/blob/v0.113.0/stackit/internal/services/sfs/resourcepool/resource.go#L298)):

```go
response, err := wait.CreateResourcePoolWaitHandler(ctx, r.client.DefaultAPI, projectId, region, *resourcePool.ResourcePool.Id).
	WaitWithContext(ctx)
```

`docs/resources/sfs_resource_pool.md` documents no `timeouts` attribute, so there is no way
to raise it from configuration. Update and delete have the same fixed 10 minutes.

### What happens when it expires

Immediately before waiting, the provider writes the identifiers to state on purpose
([same file, L288–L293](https://github.com/stackitcloud/terraform-provider-stackit/blob/v0.113.0/stackit/internal/services/sfs/resourcepool/resource.go#L288)):

```go
// Write id attributes to state before polling via the wait handler - just in case anything goes wrong during the wait handler
ctx = utils.SetAndLogStateFields(ctx, &resp.Diagnostics, &resp.State, map[string]any{
	"project_id":       projectId,
	"region":           region,
[...]

PR #1345, opened 2026-03-31
feat(dns) add timeouts to dns resources and datasources

cgoetz-inovex:
STACKITTPR-542

## Description

relates to #1234

## Checklist

- [ ] Issue was linked above
- [ ] Code format was applied: `make fmt`
- [ ] Examples were added / adjusted (see `examples/` directory)
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI)
- [ ] Unit tests got implemented or updated
- [ ] Acceptance tests got implemented or updated (see e.g. [here](https://github.com/stackitcloud/terraform-provider-stackit/blob/f5f99d170996b208672ae684b6da53420e369563/stackit/internal/services/dns/dns_acc_test.go))
- [x] Unit tests are passing: `make test` (will be checked by CI)
- [x] No linter issues: `make lint` (will be checked by CI)

PR #1740, opened 2026-09-04
feat(sfs): make resource pool wait timeouts configurable

devpie:
## Description

relates to #1737

`stackit_sfs_resource_pool` cannot create a pool that STACKIT needs more than 10 minutes to
provision. `CreateResourcePoolWaitHandler` and its update/delete counterparts set
`SetTimeout(10 * time.Minute)`, and the resource passed a context without a deadline, so that
default was the only limit and no configuration could reach it.

This adds a `timeouts` attribute (create/read/update/delete) following the pattern of
`dremio/instance`.

### Why a context deadline and not `SetTimeout`

`core/wait.WaitWithContext` applies the handler's own timeout **only when the incoming context
carries no deadline**:

```go
if _, ok := ctx.Deadline(); !ok {
	ctx, cancel = context.WithTimeout(ctx, h.timeout)
	defer cancel()
}
```
([core v0.26.0, wait.go](https://github.com/stackitcloud/stackit-sdk-go/blob/core/v0.26.0/core/wait/wait.go#L78))

Setting a context deadline per CRUD method therefore replaces the hardcoded value, and no
`SetTimeout` call is needed. Defaults stay at the wait handler value plus
`core.DefaultTimeoutMargin`, so unconfigured resources keep their current behaviour.
`TestSfsResourcePoolCreateTimeout` checks that a configured `timeouts.create` bounds the create.

### The wait error

When a wait ends because this context's deadline expired, the error names the configured timeout and
suggests raising it. The wait handler reports terminal error states and failing polls through the
same error, so the hint is added only on the deadline path, and identically in create, update and
delete. The hint is `utils.TimeoutHint` in `stackit/internal/utils`, so other resources can reuse it;
`TestTimeoutHint` pins its text.

### On point 2 of the issue

The issue also asks to keep the resource in state when the wait times out. That is already
implemented: `utils.SetAndLogStateFields` writes the IDs before the wait (CONTRIBUTING.md states
this as project doctrine, and `TestSfsResourcePoolSavesIDsOnError` covers it), the framework
initialises the create response state to a null object so no unknowns leak, and Terraform keeps the
object and marks it tainted rather than discarding it. What follows is a replace, not a lost entry.
No provider change was needed for that, only the corrected wording of the error.

`testdata/resource-pool-max.tf` sets `timeouts`, and no `ImportStateVerifyIgnore` is needed for it: terraform-plugin-testing deletes `timeouts` and `timeouts.*` from
both sides of the comparison unconditionally, after the ignore loop
([testing_new_import_state.go:398-411, v1.16.0](https://github.com/hashicorp/terraform-plugin-testing/blob/v1.16.0/helper/resource/testing_new_import_state.go#L398-L411)).
The DNS acceptance tests rely on the same behaviour.

### Not changed

`sfs/share` has the same hardcoded 10 minutes in all three of its wait handlers. Out of scope for
this issue — happy to follow up if wanted.

## Checklist

- [x] Issue was linked above
- [ ] Examples were added / adjusted (see `examples/` directory) — deliberately not: no example in `examples/` has ever carried a `timeouts` block (`git log -S timeouts -- examples/` is empty), and #1345, which added `timeouts` to the DNS resources, put the demonstrable configuration in `testdata/resource-max.tf` instead. Hardcoding durations on a registry page would also pin numbers that go stale when the SDK waiter default moves.
- [x] Acceptance tests got implemented or updated — `testdata/resource-pool-max.tf` now sets `timeouts`, mirroring `dns/testdata/resource-max.tf`, so `TestAccResourcePoolResourceMax` covers the attribute across create, import-verify and update


stackit/internal/services/sfs/resourcepool/resource.go
```diff
@@ -297,11 +311,31 @@ func (r *resourcePoolResource) Create(ctx context.Context, req resource.CreateRe
...
+		}
+		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating resource pool", fmt.Sprintf(
+			"resource pool creation waiting: %v\n"+
+				"The resource pool was created, and Terraform marks this resource as tainted, so the next apply replaces it. "+
+				"Run `terraform untaint` on it first if the next run should refresh the existing pool instead. "+
+				"If the state entry is lost, import the resource pool with the ID %q.%s",
+			err,
+			utils.BuildInternalTerraformId(projectId, region, *resourcePool.ResourcePool.Id).ValueString(),
+			timeoutHint,
+		))
```
marceljk: I would simplify the error message here a bit. I wouldn't mention that users should run `terraform untaint`, refresh or import the resource pool. The state entry should never be lost, because a few lines above, the identifier which are needed for a read are written to the state. As long this error message is printed, the resource should be saved in the state.

So the only thing I would add, would be that the timeout should be increased.

stackit/internal/services/sfs/resourcepool/resource.go
```diff
@@ -297,11 +311,31 @@ func (r *resourcePoolResource) Create(ctx context.Context, req resource.CreateRe
 	if resp.Diagnostics.HasError() {
 		return
 	}
+	// The configured timeouts belong in that partial state as well. Without them a failed wait leaves an entry whose
+	// read and delete fall back to the defaults instead of the values the operator configured.
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("timeouts"), model.Timeouts)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
marceljk: This isn't needed. The value of the timeout isn't needed in the state because for the next run (read or delete), it would check the configuration of the user. If nothing is set it fallbacks to the default value. The values from the state are not used for that.

stackit/internal/services/sfs/resourcepool/resource.go
```diff
@@ -343,6 +377,14 @@ func (r *resourcePoolResource) Read(ctx context.Context, req resource.ReadReques
 	if resp.Diagnostics.HasError() {
 		return
 	}
+	readTimeout, diags := model.Timeouts.Read(ctx, core.DefaultOperationTimeout)
```
marceljk, line 362:
```suggestion
	}
	
	readTimeout, diags := model.Timeouts.Read(ctx, core.DefaultOperationTimeout)
```
Please leave here a space between these. This makes it more readable, especially for first time contributors.

stackit/internal/services/sfs/resourcepool/resource.go
```diff
@@ -395,6 +437,15 @@ func (r *resourcePoolResource) Update(ctx context.Context, req resource.UpdateRe
 	if resp.Diagnostics.HasError() {
 		return
 	}
+	waiterTimeout := wait.UpdateResourcePoolWaitHandler(ctx, r.client.DefaultAPI, "", "", "").GetTimeout() //nolint:tfctxinit // false positive - only called to read the default wait handler timeout
```
marceljk, line 423: same here, please keep an extra line space
```suggestion
	}

	waiterTimeout := wait.UpdateResourcePoolWaitHandler(ctx, r.client.DefaultAPI, "", "", "").GetTimeout() //nolint:tfctxinit // false positive - only called to read the default wait handler timeout
```

stackit/internal/services/sfs/resourcepool/resource.go
marceljk: After reviewing the other methods, I saw that the error message remained there. Please keep the error message consistent between all methods. Either leave it like it was before or add everywhere the hint, to increase the timeout in case the context deadline was exceeded

docs/resources/sfs_resource_pool.md
```diff
@@ -68,6 +69,17 @@ Read-Only:
...
 
+
+<a id="nestedatt--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- `create` (String) A string that can be [parsed as a duration](https://pkg.go.dev/time#ParseDuration) consisting of numbers and unit suffixes, such as "30s" or "2h45m". Valid time units are "s" (seconds), "m" (minutes), "h" (hours).
+- `delete` (String) A string that can be [parsed as a duration](https://pkg.go.dev/time#ParseDuration) consisting of numbers and unit suffixes, such as "30s" or "2h45m". Valid time units are "s" (seconds), "m" (minutes), "h" (hours). Setting a timeout for a Delete operation is only applicable if changes are saved into state before the destroy operation occurs.
+- `read` (String) A string that can be [parsed as a duration](https://pkg.go.dev/time#ParseDuration) consisting of numbers and unit suffixes, such as "30s" or "2h45m". Valid time units are "s" (seconds), "m" (minutes), "h" (hours). Read operations occur during any refresh or planning operation when refresh is enabled.
```
marceljk, line 80: Please add the timeouts also for the datasource. We always add the timeouts for the resource and datasource both together

stackit/internal/services/sfs/sfs_test.go
```diff
@@ -162,3 +168,231 @@ resource "stackit_sfs_share" "example" {
 		},
 	})
 }
+
+// TestSfsResourcePoolCreateTimeoutIsConfigurable asserts that the configured `timeouts.create` value is what ends
```
marceljk: Please keep the tests easy and maintainable. You can check the test we added for the dns create timeout, something like that would be enough.
https://github.com/stackitcloud/terraform-provider-stackit/blob/5c52e10ac2ee169d84e362a5ec49128db116d2ee/stackit/internal/services/dns/dns_test.go#L16-L69

We have for example also already a MockServer implementation, which should be used here.

marceljk (commented):
Thanks for your contribution! I left some minor comments

stackit/internal/services/sfs/resourcepool/resource.go
devpie: Removed in 47412c90.

One correction for the record, because it may matter elsewhere: `Read` and `Delete` cannot consult the user's configuration. In terraform-plugin-framework v1.19.0 neither `resource.ReadRequest` nor `resource.DeleteRequest` carries a `Config` field — both only get `State` (plus `ProviderMeta`). So the timeout for those two operations does come from state, and when it is absent the default applies.

That only bites in one narrow window: after a failed create wait, the state entry holds just the identifiers, Terraform marks the resource tainted, and the destroy of that tainted object then uses the default delete timeout rather than the configured one. On every normal path the full model is in state, so it makes no difference. Given how narrow that is, I have followed your preference and dropped the line. Say the word if you would rather have that path covered and I will put it back with its test.

stackit/internal/services/sfs/resourcepool/resource.go
devpie: Done in 47412c90. The message is now just the wait error plus, when this context's deadline is what ended the wait, a sentence naming the configured timeout.

devpie: Done in 47412c90. I extracted a small `timeoutHint` helper and call it from create, update and delete, so all three read the same. It only adds the sentence when `ctx.Err()` is `context.DeadlineExceeded` — the wait handler reports terminal error states and failing polls through the same error, where the hint would name the wrong cause.

docs/resources/sfs_resource_pool.md
devpie: Added in 47412c90: the data source now carries the `timeouts` attribute with `read`, and `docs/data-sources/sfs_resource_pool.md` is regenerated.

stackit/internal/services/sfs/sfs_test.go
devpie: Rewritten in 47412c90 along the lines of the dns test, on `testutil.MockServer`, with the same note about why only create is covered. The three longer tests are gone.

stackit/internal/services/sfs/resourcepool/resource.go
devpie: Applied in 47412c90.

stackit/internal/services/sfs/resourcepool/resource.go
devpie: Applied in 47412c90. The data source got the same separation in 7f8825e3.

devpie:
Follow-up in 7f8825e3, after re-reading the result of the previous commit:

- the update wait branch still carried the summary `Error creating resource pool` while the appended hint named `timeouts.update`, so the diagnostic contradicted itself. Summary and detail now say "updating", matching what #1741 changes the same line to, so the two should not conflict.
- the hint starts on its own line again; concatenating it onto the wrapped error ran the sentences together.
- the data source read timeout got the same blank line you asked for in the resource, and the acceptance-test data source now declares a `timeouts` block, as `dns/testdata/resource-max.tf` does for its data source.

I also removed a paragraph from the PR description that still described the state write you asked me to drop.

stackit/internal/services/sfs/resourcepool/resource.go
marceljk: Good catch with the Read and Delete methods! For now it's fine to leave it like it is, because I think for the Read and Delete the timeout isn't that relevant like it's for the Create and Update. We would need it then also for our other services which supports already the timeout. I'll add a ticket for it in our backlog, to keep them all the same.

stackit/internal/services/sfs/resourcepool/resource.go
```diff
@@ -245,10 +248,21 @@ func (r *resourcePoolResource) Schema(_ context.Context, _ resource.SchemaReques
...
+// timeoutHint names the configured timeout when this context's deadline is what ended a wait. The wait handler
+// reports a timeout, a terminal error state and a failing poll through the same error, so on the other two the
+// hint would point at the wrong cause.
+func timeoutHint(ctx context.Context, operation string, timeout time.Duration) string {
+	if !errors.Is(ctx.Err(), context.DeadlineExceeded) {
+		return ""
+	}
+	return fmt.Sprintf("\nThe wait gave up after the configured `timeouts.%s` of %s; raise it if the operation regularly needs longer.", operation, timeout)
+}
+
```
marceljk: Can you move this to our utils package, so it can be also reused by other resources and datasources in the future?

devpie: Moved in 06224cef. It now lives in `stackit/internal/utils/timeouts.go` as `utils.TimeoutHint`, since nothing in it is SFS-specific. `TestTimeoutHint` next to it pins the exact text for an exceeded deadline and checks that a cancellation, a wait that failed before the deadline and a context without a deadline produce no hint.

stackit/internal/services/sfs/resourcepool/resource.go
devpie: Sounds good, thanks for tracking it in the backlog.

marceljk (approved):

SerseusWasTaken (approved):

Issue #1746, opened 2026-09-07
iaas: stackit_server panics when GetServer fails during update


devpie:
## Description

`stackit_server` crashes the provider process when the `GetServer` call at the start of `Update`
fails. The error is logged and execution continues with a `server` that is nil on that path, and a
few lines further down that value is dereferenced.

`stackit/internal/services/iaas/server/resource.go`, `main` @ a7d114b2:

```go
// :825
if server, err = r.client.DefaultAPI.GetServer(ctx, projectId, region, serverId).Execute(); err != nil {
	core.LogAndAddError(ctx, &resp.Diagnostics, "Error retrieving server state", fmt.Sprintf("Getting server state: %v", err))
}   // no return

// :840
if err := updateServerStatus(ctx, r.client.DefaultAPI, server.Status, &model, region); err != nil {
```

Any failing `GetServer` reaches this — a 500, a timeout, a transient network error. The operator
gets a plugin crash instead of the "Error retrieving server state" diagnostic that was just
recorded.

The identical missing `return` sits at `:577` in the create path. There it is caught by accident:
`mapFields` rejects a nil response at `:949` and the caller returns at `:583-585`. The visible
effect is only a misleading diagnostic — `Processing API payload: response input is nil` instead of
the call that actually failed.

## Steps to reproduce

The configuration is unremarkable; any `stackit_server` that is updated while the API is briefly
unavailable will do:

```terraform
resource "stackit_server" "example" {
  project_id   = var.project_id
  name         = "example"
  machine_type = "g1.1"
  boot_volume = {
    size        = 32
    source_type = "image"
    source_id   = var.image_id
  }
  network_interfaces = [var.nic_id]
}
```

1. `terraform apply` to create the server.
2. Change `name` (or any attribute handled by `Update`).
3. `terraform apply` again while `GET /v2/projects/{p}/regions/{r}/servers/{id}` answers `500`.

The create path can be driven deterministically with a mock server; a test for it is in the PR
below and fails without the fix.

## Actual behavior

The provider panics with a nil pointer dereference at `resource.go:840` and Terraform reports that
the plugin crashed, discarding the diagnostic the provider had already added.

## Expected behavior

`Update` returns after adding the diagnostic, so the operator sees "Error retrieving server state"
and the reason.

## Environment
 - OS: macOS 15 (darwin/arm64)
 - Terraform version: not version-specific
 - Version of the STACKIT Terraform provider: `main` @ a7d114b2

[...]

Issue #1743, opened 2026-09-04
sfs: provider panics with a nil pointer dereference when the API returns an undecodable response


devpie:
## Description

The SFS resources and data sources crash the provider process with a nil pointer dereference when an API call
returns a response that the SDK decodes to `nil`. Terraform reports it as `The plugin encountered an error, and
failed to respond to the plugin6.(*GRPCProvider).ReadResource call`, so the operator gets a plugin crash instead of
a diagnostic.

Two routes lead to a nil response, neither of which requires anything unusual from the SFS API:

**1. A 2xx answer with an empty body.** The generated SDK decoder returns early for an empty body
([`v1api/client.go`](https://github.com/stackitcloud/stackit-sdk-go/blob/services/sfs/v0.11.1/services/sfs/v1api/client.go)):

```go
func (c *APIClient) decode(v interface{}, b []byte, contentType string) (err error) {
	if len(b) == 0 {
		return nil
	}
```

`localVarReturnValue` therefore stays nil and `Execute()` returns `(nil, nil)`.

**2. A single retryable gateway error during a wait.** `WaiterHelper.Wait` returns `true` for `waitFinished` on any
fetch error, `handleError` swallows a 502 or 504 and returns a nil error, and the loop then returns the nil response
because it is already `done` — the retry never happens. Reported separately as
stackitcloud/stackit-sdk-go#11084 with a runnable reproduction.

The call sites check only the returned error. Several of the checks that are *meant* to validate a response are
themselves the dereference — they test `x.Field == nil` without first testing `x == nil`:

```go
// stackit/internal/services/sfs/resourcepool/resource.go, v0.113.0
if response.ResourcePool == nil || response.ResourcePool.Id == nil {   // panics when response is nil
```

and in both create paths the wait handler result is read by `tflog.SetField` one line *before* its guard.

## Steps to reproduce

A plain Terraform configuration cannot trigger this deterministically — it needs the API to answer a `GET` with an
empty body or a single 502, which is not something a config can provoke. The configuration involved is unremarkable:

```terraform
resource "stackit_sfs_resource_pool" "this" {
  project_id        = var.project_id
  name              = "sfs-pool"
  availability_zone = "eu01-m"
  performance_class = "Standard"
  size_gigabytes    = 512
  ip_acl            = ["192.168.2.0/24"]
}
```

The crash reproduces deterministically as a unit test against a mock server. Add this to
[...]

PR #1747, opened 2026-09-07
fix(iaas): return after a failing GetServer in the server resource

devpie:
## Description

relates to #1746

Both the create and the update path of `stackit_server` called `GetServer`, logged the error with
`core.LogAndAddError` and then carried on with a `server` that is nil on that path.

In `Update` the consequence is a crash: nothing guards the value between the failing call at `:825`
and `server.Status` at `:840`, so any failing `GetServer` — a 500, a timeout, a transient network
error — panics the provider process instead of reporting the diagnostic that was just recorded.

In `Create` the same omission is caught by accident, because `mapFields` rejects a nil response and
the caller returns on that error. The visible effect there is a misleading diagnostic
(`Processing API payload: response input is nil` instead of the call that actually failed), and the
guard is one refactor away from being gone.

The change is two `return` statements.

### Evidence

[NilAway](https://github.com/uber-go/nilaway) reports `resource.go:840` on `main` and does not
report it with this change applied. The two remaining findings in that file
(`resource.go:552`, `:775`) are a different class — the one described in #1743 — and are out of
scope here.

No unit test: the one I had written covered the create path, where the missing return only produces
a misleading diagnostic, and never reached the crash in `Update`. Removed on review. A test that
does cover the update path needs a complete server fixture to round-trip through the mock before
`Update` is reached — happy to add that if you would rather have the crash covered.

## Checklist

- [x] Issue was linked above
- [ ] Examples were added / adjusted (see `examples/` directory) — not applicable, no configuration surface changes
- [x] Docs are up-to-date: `make generate-docs` (will be checked by CI) — no schema change
- [ ] Unit tests got implemented or updated — see above, the create-path test was removed on review and does not cover the crash
- [ ] Acceptance tests got implemented or updated — a failing `GetServer` cannot be provoked against the real API


stackit/internal/services/iaas/server/resource_create_test.go
```diff
@@ -0,0 +1,84 @@
+package server_test
```
SerseusWasTaken: I don't think this test is needed here. It does not cover the second return and there is generally no need to cover every single return with a test 😄

devpie: You are right, and it is the wrong half: the test covered the create path, where the missing return only produces a misleading diagnostic. The crash is in the update path, and the test never reaches it. Removed in d52b1655.

The one thing that spoke for it was that it failed without the change, so it would catch the return being dropped again — but what it actually pinned was the diagnostic text, not the crash, which is not worth the maintenance.

A test that does cover the update path needs a complete server fixture to round-trip through the mock first (boot volume, nics) before Update is reached. Happy to add that if you would rather have the crash covered; otherwise the PR is now just the two returns.

SerseusWasTaken (approved):

marceljk (approved):

PR #1750, opened 2026-09-07
fix(iaas): remove routing table from state when it no longer exists

### LINKED ISSUE #1749: stackit_routing_table / stackit_routing_table_route: stale state after the network area is recreated (Delete only treats HTTP 404 as gone)
## Description

When a `stackit_network_area` is deleted and recreated (for example by a change that forces replacement), the platform deletes all routing tables of that area together with their routes. The Terraform state still contains the `stackit_routing_table` and `stackit_routing_table_route` resources, and every following operation on them fails.

Cause in the provider (verified, see the reproduction comment below): `Read()` of `stackit_routing_table` reports an error diagnostic on HTTP 404 instead of removing the resource from state (`stackit/internal/services/iaas/routingtable/table/resource.go`), so every refresh of the stale resource fails. `Delete()` handles 404 correctly. The first assumption that the IaaS API answers HTTP 400 for these requests did not hold for routing tables and routes; they answer 404. The API does answer 400 (`region not enabled`) for a deleted regional network area configuration while the network area still exists, which `stackit_network_area_region` does not handle (separate problem).

A second observation in `stackit_routing_table.Read()`: it calls `resp.State.RemoveResource()` after *any* API error. That call is dead code, because the error diagnostic set just before it makes Terraform keep the prior state; the fix drops it and removes the resource from state only on 404.

## Steps to reproduce

```terraform
resource "stackit_network_area" "example" {
  organization_id = var.organization_id
  name            = "example"
  # ...
}

resource "stackit_routing_table" "example" {
  organization_id = var.organization_id
  network_area_id = stackit_network_area.example.network_area_id
  name            = "example"
}

resource "stackit_routing_table_route" "example" {
  organization_id  = var.organization_id
  network_area_id  = stackit_network_area.example.network_area_id
  routing_table_id = stackit_routing_table.example.routing_table_id
  destination = {
    type  = "cidrv4"
    value = "10.0.0.0/24"
  }
  next_hop = {
    type = "blackhole"
  }
}
```

1. `terraform apply`
2. Change an attribute of `stackit_network_area.example` that forces replacement, or delete and recreate the network area outside of Terraform
3. `terraform apply` (the same happens with `terraform plan`, and with a `removed` block that uses `destroy = true`)

## Actual behavior

The apply fails with an API error for the old routing table (the routing table does not exist). The only way out is a `removed` block with `destroy = false` for the routing table and its routes, followed by a normal apply that creates them again.

## Expected behavior

- `Delete()` treats the routing table and its routes as already deleted when the network area is gone, like it does for HTTP 404.
- `Read()` removes the resources from state when they are gone and keeps the state on any other error.
- The docs of `stackit_routing_table` and `stackit_routing_table_route` mention that the platform deletes routing tables and routes together with their network area.

## Environment

Reported by the person who hit this:

 - Terraform version: `v1.15.8`
 - Version of the STACKIT Terraform provider: `v0.114.0`
 - Reproduced from Azure DevOps (Microsoft-hosted `ubuntu-latest` agents), from macOS 25.6, and from WSL with Ubuntu 24.04 LTS

My own reproduction (see the comments below):

 - OS: macOS 26.6.2 (arm64)
 - Terraform version: OpenTofu v1.12.6
 - Version of the STACKIT Terraform provider: `v0.114.0`

**Additional information**

Routing tables are still a public preview feature. Fix: #1750.

devpie:
## Description

fixes #1749

When a `stackit_network_area` no longer exists (the area or its regional configuration was deleted outside of Terraform), the IaaS API answers HTTP 404 for the routing tables and routes that belonged to it (verified against the API, see #1749). `Read()` of `stackit_routing_table` turned that 404 into an error diagnostic: it called `utils.LogError` and then `RemoveResource`, and Terraform keeps the prior state when a read reports an error. So every refresh of the stale resource failed, and a `removed` block with `lifecycle { destroy = true }` failed with it. The same code removed the resource from state after *any* API error, which was equally ineffective for the same reason.

Changes:

- `stackit_routing_table.Read()` removes the resource from state on HTTP 404 without an error diagnostic and reports every other error while keeping the state — the pattern `stackit_routing_table_route.Read()` already uses. The error summary follows the other IaaS resources (`Error reading routing table`).
- Resource-level test for `Read()` against a mocked API: 200 (state compared field by field), 404, 403, 500 and a transport error. It fails on `main` for the 404, 403 and 500 cases.
- Schema descriptions and generated docs of both resources state that the platform deletes routing tables and routes together with their `stackit_network_area_region`, and how the provider reacts.

Verification against the real API (provider v0.114.0 vs. this build, OpenTofu 1.12.6; details in #1749): with the network area deleted, `tofu plan -refresh-only` fails on v0.114.0 with `Error: Reading routing table` and succeeds with this build, reporting routing table and route as deleted; `tofu destroy -refresh=false` succeeds for the already deleted routing table and route.

Not part of this PR: `stackit_network_area_region` and `stackit_network_area_route` have related gaps, see #1751.

## Checklist

- [x] Issue was linked above
- [ ] Examples were added / adjusted (see `examples/` directory) — not needed, no attribute changed
- [ ] Acceptance tests got implemented or updated (see e.g. [here](https://github.com/stackitcloud/terraform-provider-stackit/blob/f5f99d170996b208672ae684b6da53420e369563/stackit/internal/services/dns/dns_acc_test.go)) — not changed; the existing routing table acceptance tests cover the normal lifecycle

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/resources/routing_table.md
+++ b/docs/resources/routing_table.md
@@ -5,6 +5,7 @@ subcategory: ""
 description: |-
   Routing table resource schema. Must have a region specified in the provider configuration.
   This resource is for SNA, not VPC, based networks.
+  The platform deletes routing tables and their routes together with the network area region (stackit_network_area_region) they belong to. When the network area or its region no longer exists, the provider treats the routing table as deleted: it is removed from the Terraform state on refresh, and destroying it succeeds.
   ~> This resource is part of the experimental feature routing-tables and is likely going to undergo significant changes or be removed in the future. Use it at your own discretion.
 ---
 
@@ -13,6 +14,8 @@ description: |-
 Routing table resource schema. Must have a `region` specified in the provider configuration.
 
 This resource is for SNA, not VPC, based networks.
+
+The platform deletes routing tables and their routes together with the network area region (`stackit_network_area_region`) they belong to. When the network area or its region no longer exists, the provider treats the routing table as deleted: it is removed from the Terraform state on refresh, and destroying it succeeds.
 
 ~> This resource is part of the experimental feature routing-tables and is likely going to undergo significant changes or be removed in the future. Use it at your own discretion.
 
--- a/docs/resources/routing_table_route.md
+++ b/docs/resources/routing_table_route.md
@@ -5,6 +5,7 @@ subcategory: ""
 description: |-
   Routing table route resource schema. Must have a region specified in the provider configuration.
   This resource is for SNA, not VPC, based networks.
+  The platform deletes routing tables and their routes together with the network area region (stackit_network_area_region) they belong to. When the network area or its region no longer exists, the provider treats the route as deleted: it is removed from the Terraform state on refresh, and destroying it succeeds.
   ~> This resource is part of the experimental feature routing-tables and is likely going to undergo significant changes or be removed in the future. Use it at your own discretion.
 ---
 
@@ -13,6 +14,8 @@ description: |-
 Routing table route resource schema. Must have a `region` specified in the provider configuration.
 
 This resource is for SNA, not VPC, based networks.
+
+The platform deletes routing tables and their routes together with the network area region (`stackit_network_area_region`) they belong to. When the network area or its region no longer exists, the provider treats the route as deleted: it is removed from the Terraform state on refresh, and destroying it succeeds.
 
 ~> This resource is part of the experimental feature routing-tables and is likely going to undergo significant changes or be removed in the future. Use it at your own discretion.
 
--- a/stackit/internal/services/iaas/routingtable/table/resource_test.go
+++ b/stackit/internal/services/iaas/routingtable/table/resource_test.go
@@ -3,12 +3,21 @@ package table
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/gorilla/mux"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/stackitcloud/stackit-sdk-go/core/config"
 	iaas "github.com/stackitcloud/stackit-sdk-go/services/iaas/v2api"
+
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/core"
+	"github.com/stackitcloud/terraform-provider-stackit/stackit/internal/utils"
 )
 
 func TestMapFields(t *testing.T) {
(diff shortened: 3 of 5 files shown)


SerseusWasTaken:
Hello @devpie, thank you for the contribution.

I have created a internal tracking issue and we will review your contribution soon. 

Best regards

devpie:
Hi @SerseusWasTaken, is there anything I can do to move this forward?

I rebased the branch onto current main today, so it is conflict-free, and `make test` and `make lint` pass locally. Happy to adjust anything in the change itself if that helps.

SerseusWasTaken:
Hey @devpie, this PR is planned for the next sprint, which is starting today. So we will very soon be able to take a proper look. Sorry for the delay, and thank you for your patience and keeping the PR updated!

Best regards
Jonas from the STACKIT Devloper Tools team

stackit/internal/services/iaas/routingtable/table/resource_test.go
```diff
@@ -215,3 +224,138 @@ func TestToUpdatePayload(t *testing.T) {
 		})
 	}
 }
+
+func TestRead(t *testing.T) {
```
cgoetz-inovex, line 228: I'd remove this test. We are currently working on an issue to introduce a linting rule to cover this case:
> Resource which are managed via Terraform, could be deleted either via the Portal, CLI or directly the API. For this reason, our provider needs to take care of if the API responds with a 404 and/or 403, within the Read(...) method of the resource and remove then the resource from the state.

cgoetz-inovex (changes requested):

docs/resources/routing_table.md
```diff
@@ -5,6 +5,7 @@ subcategory: ""
 description: |-
   Routing table resource schema. Must have a region specified in the provider configuration.
   This resource is for SNA, not VPC, based networks.
+  The platform deletes routing tables and their routes together with the network area region (stackit_network_area_region) they belong to. When the network area or its region no longer exists, the provider treats the routing table as deleted: it is removed from the Terraform state on refresh, and destroying it succeeds.
```
marceljk, line 8: I would remove this comment, because I would see this as expected behavior. The same applies for example, if a project was deleted, then all resources which belong to the project would be removed as well and Terraform should recognize this. So i don't see the need here to highlight it with a comment

docs/resources/routing_table_route.md
```diff
@@ -5,6 +5,7 @@ subcategory: ""
 description: |-
   Routing table route resource schema. Must have a region specified in the provider configuration.
   This resource is for SNA, not VPC, based networks.
+  The platform deletes routing tables and their routes together with the network area region (stackit_network_area_region) they belong to. When the network area or its region no longer exists, the provider treats the route as deleted: it is removed from the Terraform state on refresh, and destroying it succeeds.
```
marceljk, line 8: same here

stackit/internal/services/iaas/routingtable/table/resource_test.go
devpie: Removed in b34399e8. One note for the record: the lint rule will catch the missing 404 handling in `Read()`, but not that the resource actually leaves the state — that was what the test pinned. Happy to bring it back in whatever shape you prefer once the rule is in place.

docs/resources/routing_table.md
devpie: Removed in b34399e8, along with the same sentence on the route resource. The docs are regenerated, so both files are back to what the generator produces.

docs/resources/routing_table_route.md
devpie: Removed as well, see above.

PR #1761, opened 2026-09-10
feat: import user with new password

mhenselin:
## Description

Currently there is no possibility to import an existing PostgresFlex User with a password. This change gives the option to have the user imported with a new password. This enables terraform to reference the password in further steps.

## Checklist

- [ ] Issue was linked above


rubenhoenle:
Sorry, but I don't think this is the solution we aim for. 

The actual root cause here is that you currently can't rotate the user password via Terraform. We should allow this somehow.

For the import you would import the user first and then rigger a rotation of the password via Terraform.

Thanks for the PR, but I consider this technical debt and nothing we should even consider :sweat_smile:

PR #1762, opened 2026-09-11
feat(alb): add alt_port

SerseusWasTaken:
## Description
Adds attribute `altPort` to configure an alternative port for the health checks.

## Checklist

- [ ] Issue was linked above

### DIFF
--- a/stackit/internal/services/alb/testfiles/resource-max.tf
+++ b/stackit/internal/services/alb/testfiles/resource-max.tf
@@ -87,7 +87,7 @@ resource "stackit_network" "listener_network" {
   project_id       = var.project_id
   name             = var.network_name_listener
   ipv4_nameservers = ["1.1.1.1"]
-  ipv4_prefix      = "10.11.10.0/24"
+  ipv4_prefix      = "10.2.2.0/24"
   routed           = "true"
 }
 
@@ -95,7 +95,7 @@ resource "stackit_network" "target_network" {
   project_id       = var.project_id
   name             = var.network_name_targets
   ipv4_nameservers = ["1.1.1.1"]
-  ipv4_prefix      = "10.11.11.0/24"
+  ipv4_prefix      = "10.2.3.0/24"
   routed           = "true"
 }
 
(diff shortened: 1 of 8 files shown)


stackit/internal/services/alb/testfiles/resource-max.tf
```diff
@@ -86,15 +87,15 @@ resource "stackit_network" "listener_network" {
...
+  ipv4_prefix      = "10.2.2.0/24"
   routed           = "true"
 }
 
 resource "stackit_network" "target_network" {
   project_id       = var.project_id
   name             = var.network_name_targets
   ipv4_nameservers = ["1.1.1.1"]
-  ipv4_prefix      = "10.11.11.0/24"
+  ipv4_prefix      = "10.2.3.0/24"
```
marceljk, line 98: Did you had some conflicts with the prefix range with the old values? If yes, I would suggest to either rewrite the Acc tests in the future to create a clean project for every run, to prevent potential conflicts. Or you can set ipv4_prefix_length instead of the ipv4_prefix, then the IaaS api assigns an available ipv4_prefix.

But this is nothing which must be done in this PR. Just a hint for future improvements :)

stackit/internal/services/alb/testfiles/resource-min.tf
```diff
@@ -40,7 +40,7 @@ resource "stackit_network" "network" {
   project_id       = var.project_id
   name             = var.network_name
   ipv4_nameservers = ["1.1.1.1"]
-  ipv4_prefix      = "192.168.3.0/25"
+  ipv4_prefix      = "10.0.0.0/25"
```
marceljk, line 43: same here

stackit/internal/services/alb/testfiles/resource-max.tf
SerseusWasTaken: I just encountered the error message that the given prefix is disallowed and took a look at the network-ranges of the acc test project, which is how i ended up with the new values. 

I agree that creating a clean project for each test is probably the proper solution though, other tests do that as well. I will do that in the future should i encounter a similar problem again

marceljk (approved):

marceljk (approved):

PR #1764, opened 2026-09-11
chore(objectstorage): add retry on 429 for objectstorage control plane

h3adex:
## Description

Adds retry for object storage control plane API

## Checklist

- [x] Issue was linked above


h3adex:
Test Code:

```terraform
// Single bucket
resource "stackit_objectstorage_bucket" "test" {
  project_id = var.stackit_project_id
  name       = "rate-limit-test"
}

# Read the same bucket 200 times via separate datasource instances.
# for_each over a set of indices gives 200 independent Read calls,
# each hitting the Object Storage API enough to trigger the 60 req/min limit
# and exercise the retry + jitter logic without creating 100 real buckets.
data "stackit_objectstorage_bucket" "test" {
  for_each = toset([for i in range(200) : tostring(i)])

  project_id = stackit_objectstorage_bucket.test.project_id
  name       = stackit_objectstorage_bucket.test.name
  region     = stackit_objectstorage_bucket.test.region
}
```

Current upstream build (Error is wrong displayed it is 429 and not 404). Also fixed that:
 
Screenshot 2026-09-14 at 11 59 45


With new build trying to read 200 buckets:

Screenshot 2026-09-14 at 11 59 31

stackit/internal/services/objectstorage/compliance-lock/datasource.go
```diff
@@ -119,7 +119,8 @@ func (d *compliancelockDataSource) Read(ctx context.Context, req datasource.Read
 			"Reading compliance lock",
 			fmt.Sprintf("Compliance lock does not exist in project %q.", projectId),
 			map[int]string{
-				http.StatusForbidden: fmt.Sprintf("Project with ID %q not found or forbidden access", projectId),
+				http.StatusForbidden:       fmt.Sprintf("Project with ID %q not found or forbidden access", projectId),
+				http.StatusTooManyRequests: objectstorageUtils.RateLimitErrMsg,
```
marceljk, line 127: retry is missing in this datasource

marceljk (dismissed):

stackit/internal/services/objectstorage/compliance-lock/resource.go
```diff
@@ -165,20 +165,32 @@ func (r *compliancelockResource) Create(ctx context.Context, req resource.Create
...
 		var oapiErr *oapierror.GenericOpenAPIError
 		ok := errors.As(err, &oapiErr)
 
 		if !(ok && oapiErr.StatusCode == http.StatusConflict) {
-			core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating compliance lock", fmt.Sprintf("Calling API: %v", err))
+			utils.LogError(ctx, &resp.Diagnostics, err, "Error creating compliance lock", fmt.Sprintf("Calling API: %v", err),
+				map[int]string{http.StatusTooManyRequests: objectstorageUtils.RateLimitErrMsg},
+			)
 			return
 		}
```
SerseusWasTaken: This if check only guarantees that the encountered error is not a `StatusConflict`, it does not ensure we encountered a `StatusTooManyRequests`, unless i missed something. We should not assume that it is a rate limit in this case. I would prefer to have an explicit if check for rate limit that prints the `objectstorageUtils.RateLimitErrMsg` and a generic error branch that does not.

stackit/internal/services/objectstorage/utils/util.go
```diff
@@ -18,8 +19,37 @@ import (
...
+//     are typically consumed within a few seconds; without jitter all goroutines wake at the
+//     exact same moment ~50s later and collide again.
+//   - 60s cap: covers a full rate-limit window reset.
+//   - 15 attempts: total budget ≈790s, covering up to ~790 concurrent goroutines and
+//     well above the ~1000s run time for 1000 buckets at 60 req/min.
+var RateLimitRetryConfig = utils.RetryConfig{
+	Attempts: 15,
+	Backoff: func(attempt int) time.Duration {
+		base := min(10*time.Second*(1<<uint(attempt-1)), 60*time.Second)
+		// Add ±25% jitter: spread = base/4, centered on base -Y [0.75×base, 1.25×base]
```
SerseusWasTaken: What does the -Y refer to? 😄

h3adex: Changed to something easier

rubenhoenle (changes requested):
Wrapping every API call with a util function will be a mess to roll out across the whole provider and to maintain. I suggest we go with some retry handler in the SDK clients instead. See #1771 for reference. This PR also makes use of the `Retry-After` header which we should definitely take into account.

